PR #666, opened 2014-05-04
Game Information / GiantBomb - feature improvements and rename

TomBebbington:
- Rename to VideoGames
- Add several sorting methods and relevancy info
- Add release date under the heading
- Use product template

It might take a while to load because you have to get the user reviews for each release of the game. I capped the releases at 5 and made it start from the latest release as that should be the most up-to-date. Also, make sure you have an API key when you're testing it.

### DIFF
--- a/share/spice/video_games/video_games.js
+++ b/share/spice/video_games/video_games.js
@@ -31,7 +31,6 @@
             },
             normalize: function(item) {
                 return {
-                    title: item.name,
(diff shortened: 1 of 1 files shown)


MrTimscampi:
So, what's the status on this ?

I was planning on making a Spice for this, before seeing it already existed, but nothing triggers. Was this never finished ?

I could take over and update it if @TomBebbington doesn't plan to update it anymore ?

PR #850, opened 2014-06-02
WIP: forvo audio with single item

jdorweiler:
The audio view currently only shows tiles.  We should have a single item audio view. selection_047
cc @chrismorast @bsstoner @russellholt

### DIFF
--- a/share/spice/forvo/forvo.js
+++ b/share/spice/forvo/forvo.js
@@ -1,151 +1,51 @@
-nrj("soundmanager2/script/soundmanager2-nodebug-jsmin.js", true);
-
 (function(env) {
-        "use strict"
 
-    env.ddg_spice_forvo = function (api_result) {
-        "use strict";
+    var script = $('[src*="/js/spice/forvo/"]')[0];
+    var source = $(script).attr("src");
+    var query = source.match(/forvo\/([^\/]+)\/\w+/)[1];
 
+    env.ddg_spice_forvo = function(api_result) {
+        
         if (api_result.attributes.total < 1) return;
-
-        var script = $('[src*="/js/spice/forvo/"]')[0];
-        var source = $(script).attr("src");
-        var query = source.match(/forvo\/([^\/]+)\/\w+/)[1];
-
-        console.log(api_result.items);
 
         // Display the Spice plug-in.
         Spice.add({
-            id: 'forvo',
+            id: "forvo",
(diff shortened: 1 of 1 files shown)


javathunderman:
@moollaza @jagtalon @jdorweiler I'm going to pick this up and see what happens, but I'm not really sure if I'm the best person to tackle this. Anyone else want to do this?

moollaza:
@javathunderman give it a go! We're all here to help :+1:

javathunderman:
@moollaza Is there any old code to work off of?

moollaza:
@javathunderman there was a branch that I deleted because it was really outdated. I would model this after the SoundCloud IA in terms of the JS and templates:

https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/sound_cloud/sound_cloud.js#L38

The most important part is specifying the `model` and `view` are Audio. This will allow the tiles to automatically have an audio player so that clicking the tile will play the audio and show the progress.

moollaza:
I restored the branch so you can see the old code, but like I said a lot of it is outdated.

jdorweiler:
Check out dictionary too.  It also uses audio. https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/dictionary/definition/dictionary_definition.js

javathunderman:
@moollaza @jdorweiler No, can't do it. The JS is just a bit too complex for me.

moollaza:
@javathunderman I suggest you at least give it a go! That way we can help you in the PR and you'll learn some stuff along the way.

The dictionary IA is indeed very advanced, but the Forvo IA, AFAIK shouldn't be much more different than a standard Spice that uses templates to produce  a tile view.

As mentioned the only different/important piece is that `Spice.add()` needs to specify the `view` and `model` are `'Audio'` and then in your `normalize` function you'll need to specify a few special properties that the `Audio` model expects:

```js
usingWaveformImage: usingWaveformImage,
duration: o.duration,
streamURL: streamURL
```

The code should look very similar to SoundCloud. All the crazy JS in the Forvo branch is now replaced with `view: 'Audio'` 

You can do it :+1:

javathunderman:
Alright, you've convinced me. :+1: 

I'm currently working on the ISS location (#1559) and Melting/Boiling temperatures (submitting a PR shortly)

PR #1208, opened 2014-10-29
Add querystring param to nypass yummly login wall

bsstoner:
@jagtalon 

cc @chrismorast @pswam


jedwards1211:
:+1: for showing me this trick, though I was disappointed that of all the recipe websites you could have picked, you picked one that requires sign up (unless you know this trick I guess?) and probably tracks users, for DuckDuckGo.  I'm on DDG because I don't want to deal with that kind of BS.

PR #1298, opened 2014-11-30
Lastfm (Album): Port to new spice api

killerfish:
@jagtalon strip_html, is it supposed to work for html entities like `&quot;` as well? Also advise what we do in case description is missing. SS below
lastfmalbum


jagtalon:
This looks good! Will just test it

jagtalon:
@killerfish Yeah `strip_html` doesn't convert things like &quot;, but maybe it should. If it doesn't have a description, it's probably not better than the links so maybe it should not be displayed. What do you think?

jagtalon:
:+1:

PR #1300, opened 2014-11-30
Lastfm (ArtistAlbum): Port to new spice api

killerfish:
@jagtalon what do you recommend we do with this one, have limited data. I wanted to include the playcount and also raise the limit on albums displayed, what are your thoughts on this?
lastfmaa2


jagtalon:
Putting on hold due to https://github.com/duckduckgo/zeroclickinfo-spice/pull/1301#issuecomment-69954686

jagtalon:
@killerfish so sorry, but closing this for now. We're looking for other providers like https://bop.fm/ or maybe we'll think of an alternative design for this.

PR #1301, opened 2014-12-01
Lastfm (ArtistTracks): Port to new spice api 

killerfish:
@jagtalon i think we need another template design, for item (image + two column split).
lastfmat


jagtalon:
@killerfish @chrismorast @abeyang I really like the design, although I'm sorry we'd have to contact them first because, if I remember correctly, we're not allowed to use the images. I'll get back to you guys as soon as we hear from them.

jagtalon:
Closing for now for reasons described in https://github.com/duckduckgo/zeroclickinfo-spice/pull/1300#issuecomment-71098300

PR #1302, opened 2014-12-01
Lastfm (TopTracks): Port to new spice api

killerfish:
@jagtalon I didnt include the css changes as with #1301, so lmk know your thoughts, comparing this. Also we need a better sign for ~ (it represents listeners, the play button was for playcount). Also should we limit result to 20? or what value?
toptracks

### DIFF
--- a/share/spice/lastfm/top_tracks/lastfm_top_tracks.handlebars
+++ b/share/spice/lastfm/top_tracks/lastfm_top_tracks.handlebars
@@ -1,1 +1,0 @@
-<a href="{{url}}">{{name}}</a> by <a href="{{artist.url}}">{{artist.name}}</a>
--- a/share/spice/lastfm/top_tracks/lastfm_top_tracks.js
+++ b/share/spice/lastfm/top_tracks/lastfm_top_tracks.js
@@ -1,48 +1,52 @@
-ddg_spice_lastfm_top_tracks = function (api_result) {
+(function (env) {
+    "use strict";
+    env.ddg_spice_lastfm_top_tracks = function (api_result) {
 
-    if(api_result.error || !api_result.toptracks || !api_result.toptracks.track || api_result.toptracks.track.length === 0) {
-        Spice.failed('lastfm_top_tracks');
-    }
- 
-    Spice.add({
-        id: 'lastfm_top_tracks',
-        name: 'Top Songs (' + api_result.toptracks['@attr'].country + ')',
-        data: api_result.toptracks.track,
-        meta: {
-            itemType: 'Tracks',
-            sourceName: 'Last.fm',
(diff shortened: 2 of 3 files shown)


jagtalon:
@killerfish Can "United States" be in the "Showing 49 Tracks" line?

jagtalon:
@moollaza @killerfish I just remembered that we weren't allowed to use the images! Dang, I really like this design. Maybe we should contact them again.

jagtalon:
I think we can do a list view here and would still work without images. Will get back to you soon!

jagtalon:
Closing for now for reasons described in https://github.com/duckduckgo/zeroclickinfo-spice/pull/1300#issuecomment-71098300

jagtalon:
@killerfish since we can't do images, would you be interested in making this something like whois?

https://duckduckgo.com/?q=whois+duckduckgo.com&ia=whois

screen shot 2015-02-19 at 12 56 32 pm

killerfish:
Its just we cant use images yes? ill check out this one plus something i had in mind

jagtalon:
@killerfish yeah unfortunately we can't use images. :/

jagtalon:
@killerfish so would you be interested in making it look like WhoIs? :)

killerfish:
Sure!

killerfish:
@jagtalon isnt there are a way we could represent all tracks without using the list view? It just doesnt seem fit right with me (there is a long list of tracks, i'd want to atleast include half, which just look odd with list view).

moollaza:
@killerfish we actually just added in a new Template group called `list` which is designed to be used with the record_template. It will allow the initial display height to be controlled and adds a "show more" button so it will expand. How long is the list of songs?

killerfish:
@moollaza 49 (in this example) :D

moollaza:
@killerfish oh wow. How about we limit it to show the first 10/15/20 (whatever looks best). I think being able to see the top 10 tracks is most important. If you really need to see the 48th top track, you can click through to the Last.FM ;)

killerfish:
@moollaza do we have list view documented?

moollaza:
@killerfish unfortunately, not yet but here's an example: https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1088#issuecomment-89651349

Your `data` obj needs `list` property (like `record_data` for an infobox), which should be an array of objects. You then need to provide a template to `list_content` which will take each object in the `list` array as its context and render an `<li>` filled with the output from your `list_content` template.

jagtalon:
Hey, @killerfish! Still interested in doing the list view? :neckbeard:

moollaza:
Closing. This is stale. Happy to reopen

PR #1313, opened 2014-12-02
TV show IA

tvdavid:
Hey all,

This is not a definite version yet intended for complete scrutiny, but I've been working on this for a couple of days and would love to get some early feedback.

This IA uses the www.tvmaze.com API to display all sorts of information about TV shows. Currently it answers to the generic query "$query tv show" with general information about that show, and to "$query next episode" with the name & airdate for that show's upcoming episode. If we want, it could be expanded to additional queries such as "popular tv shows", "$query previous episode", or "$query episode list".

Screenshots below, or you can play with it live on: http://duckdev.tvmaze.org:5000/?q=the+walking+dead+tv+show

tvshow
nextepisode

I'd love some feedback about this IA in general, the layout / look & feel, and the templates that I've used. Specifically I wasn't very sure about the auxiliary template used in Show: I borrowed the HTML from the wikipedia IA, but that one doesn't seem to be included for this repository nor could I find the template that it's using? At first I tried to put the _record_ template there, but that looked pretty bad.

Also let me know if there's any additional information you'd like to see on these pages, or if you have ideas for additional answer types than I mentioned above :)


jagtalon:
Hey, @tvdavid! We'll be checking this soon! Sorry for the late reply!

jagtalon:
@tvdavid 

- Would have time and date as something like "Dec. 2, 2014" instead of "10-2-2014". Also what's the timezone?

screen shot 2015-01-14 at 12 47 55 pm

jagtalon:
Will put this up on a server so that our designers can play with it! :)

share/spice/tvmaze/nextepisode/tvmaze_nextepisode.js
```diff
@@ -0,0 +1,32 @@
+(function (env) {
+    "use strict";
+
+    env.ddg_spice_tvmaze_nextepisode = function(api_result){
+        if (!api_result._embedded) {
```
jagtalon: Should check if `api_result` exists in the first place

share/spice/tvmaze/show/tvmaze_show.js
```diff
@@ -0,0 +1,80 @@
+(function (env) {
+    "use strict";
+
+    env.ddg_spice_tvmaze_show = function(api_result){
+        if (!api_result.id) {
```
jagtalon: Check if `api_result` exists before accessing ID

tvdavid:
Cool!

The *airdate* property is always in the show's "native timezone": the timezone of the country where new episodes premiere. A property with the ISO-8601 formatted date is available too though, if we want to show something ("in XX hours") compared to the user's timezone.

tvdavid:
> Would have time and date as something like "Dec. 2, 2014" instead of "10-2-2014".

@jagtalon Looks like I need to create a new handlebar helper to format it like that, correct? Is there a way to place that in a "common" location where it can be shared between multiple IA's, or should it be copied into each IA's javascript file?

jagtalon:
No common location at the moment, but yeah handlebars. :)

tvdavid:
> Would have time and date as something like "Dec. 2, 2014" instead of "10-2-2014".

Done now as well.

jagtalon:
@tvdavid thanks for that--testing it out!

jagtalon:
@tvdavid is it possible for the result to come out in http://api.tvmaze.com/instantsearch/shows?q=the+walking+dead instead of redirecting to http://api.tvmaze.com/shows/73? Our current setup doesn't handle redirects at the moment unfortunately.

jagtalon:
In the mean time, @abeyang what do you think of the layout? http://duckdev.tvmaze.org:5000/?q=the+walking+dead+tv+show

tvdavid:
@jagtalon Ouch, everything seems to work fine here, do you mean something that only affects the production environment?

It's not super trivial to change, and the redirect approach is much preferred for performance reasons. Could it be fixed on your end?

jagtalon:
@tvdavid unfortunately we get a lot of issues with open redirects--is it possible to maybe call two endpoints as an alternative?

abeyang:
This looks good, except can we remove "show information" from the right? I think it's apparent enough without it. Thanks! 

pasted_image_1_26_15__10_09_am

tvdavid:
@abeyang Done!

@jagtalon I've been thinking about a smarter solution, but I can't really understand how the problem of open redirects applies to this situation. If you care to brainstorm about it, I'm all ears :)

If we can't use redirects, using two separate requests would be a good alternative, but I didn't see a way of doing so in the backend. How would this be possible?

jagtalon:
@tvdavid Ah! I didn't see this before--what if we just use the http://api.tvmaze.com/search/shows?q=the%20walking%20dead endpoint and always get the first result? Since that endpoint doesn't do any redirecting, it should work well!

tvdavid:
@jagtalon That would work for the generic ("the walking dead tv show") query, but not for things like "previous episode" and "next episode", because for performance reasons the regular search endpoint doesn't allow embedding.

We've thought about this for a while, and concluded that since handling redirects apparently isn't universally supported, we should support an alternative as well for such an important API endpoint.

Just pushed the update. Let me know if everything works properly now!

moollaza:
@jagtalon @tvdavid everything seems to work well now on http://duckdev.tvmaze.org:5000/?q=the+walking+dead+tv+show

Nicely done guys, let's get this live!

lib/DDG/Spice/Tvmaze/Nextepisode.pm
```diff
@@ -0,0 +1,36 @@
+package DDG::Spice::Tvmaze::Nextepisode;
+
+use DDG::Spice;
+
+# the next episode will change the moment an episode has aired, so cache it for just one hour
+spice is_cached => 1;
+spice proxy_cache_valid => "200 301 302 404 1h";
```
moollaza: @tvdavid do we need to cache 301, 202 and esp. 404 responses?

share/spice/tvmaze/nextepisode/tvmaze_nextepisode.js
```diff
@@ -0,0 +1,39 @@
+(function (env) {
+    "use strict";
+
+    env.ddg_spice_tvmaze_nextepisode = function(api_result){
+        if (!api_result || !api_result._embedded) {
+            return Spice.failed('tvmaze');
```
moollaza: these need to match the Perl package/callback name: `Spice.failed('tvmaze_nextepisode');`

share/spice/tvmaze/previousepisode/tvmaze_previousepisode.js
```diff
@@ -0,0 +1,39 @@
+(function (env) {
+    "use strict";
+
+    env.ddg_spice_tvmaze_previousepisode = function(api_result){
+        if (!api_result || !api_result._embedded) {
+            return Spice.failed('tvmaze');
```
moollaza: same here, `Spice.failed('tvmaze_previousepisode');`

share/spice/tvmaze/nextepisode/tvmaze_nextepisode.js
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        Spice.add({
+            id: "tvmaze_nextepisode",
+            name: "TV Shows",
```
moollaza: I think the `name` for these should just be be `Television`

moollaza:
Everything here LGTM :+1:

moollaza:
Continued in #1483

lib/DDG/Spice/Tvmaze/Nextepisode.pm
tvdavid: Ah, the 301 and 302 could be removed now that we're no longer using an endpoint that redirects. The 404 was intentional though: if no match was found for your query, the API will return a 404. Such a result should be cached just the same, I think.

tvdavid:
@moollaza Great to see, and you're very welcome! :)

moollaza:
apologies for the delayed notice @tvdavid but this is live now! https://duckduckgo.com/?q=game+of+thrones+show&ia=entertainment

tvdavid:
I see it's working perfectly, awesome! If there are any issues or ideas for other TV-related IA's, be sure to let me know.

moollaza:
@tvdavid will do!

PR #1340, opened 2014-12-11
Movie: Check movie year against searched year

### LINKED ISSUE #1339: Movies disambiguation with year.
If someone types in the year of the movie release, just show that single result when multiple results are available.

img_5872

https://duck.co/ia/view/movie

MrChrisW:
Fixes #1339

Ping pong @jagtalon 

**Example:**
image
image

__________________________________________
~~Note: There may be a relevancy issue, example search for "Babe 2006" https://duckduckgo.com/?q=Babe+2006+rotten+tomatoes~~ Fixed

### DIFF
--- a/share/spice/movie/movie.js
+++ b/share/spice/movie/movie.js
@@ -26,20 +26,25 @@
         // Get original query.
         var script = $('[src*="/js/spice/movie/"]')[0],
             source = $(script).attr("src"),
-            query = source.match(/movie\/([^\/]+)/)[1];
+            query = decodeURIComponent(source.match(/movie\/([^\/]+)/)[1]);
         
         //Check if the query contains a year
-        var year = decodeURIComponent(query).match(/\d{4}/gm), 
+        var year = query.match(/\d{4}/),
             singleResult = [];
-        
+
+        //Ensure year is vaild between 1900 - current year + 5
+        year = (year >= 1900 && year <= new Date().getFullYear()+5) ? year : null;
+
         if(year) {
+        ///Check movie year and title against query
         $.each(api_result.movies, function(key, result) {
-        //Check movie year and title against query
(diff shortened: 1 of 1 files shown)


jagtalon:
@MrChrisW Nice--what about queries like "RoboCop (2014 film) rotten tomatoes"?

It looks like `trim` doesn't work on IE 8 either: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

jagtalon:
Also would like some indentation in there. :)

MrChrisW:
@jagtalon The main duckduckhack JS contains the `String.prototype.trim` polyfill so using trim shouldn't be an issue.

MrChrisW:
@jagtalon That's a wonderful query *"RoboCop (2014 film) rotten tomatoes"* :laughing: Thanks for the curve ball :soccer: 

Few options when comparing the query to the movie title.

+ Strip everything within round brackets `/ *\([^)]*\) */g`
  + Probably not the best option as someone could search a movie that contains brackets in the title. Example: [(500) Days of Summer](http://www.imdb.com/title/tt1022603/)

+ Strip keywords `film`, `movie` and round brackets.
 + Movies can contain these keywords and this isn't very maintainable :frowning:

jagtalon:
@MrChrisW Maybe check if there's `\d{4}` inside the brackets? Would that work?

MrChrisW:
@jagtalon Best I can think of is below. This will allow the query `RoboCop (2014) rotten tomatoes.

````javascript
 if (query.match(/\(.+?\)/g)) { #match parenthesis
     query = query.replace(/\(|\)/g, ''); #remove parenthesis
 }
`

However `RoboCop (2014 film) rotten tomatoes` is a little more difficult as you're dealing with arbatery keywords `film` `movie` etc.

MrChrisW:
Ping @jagtalon

jagtalon:
@MrChrisW got it! :+1:

PR #1352, opened 2014-12-17
Refactoring forecast spice

kethinov:
**Changes:**
- Namespaced CSS.
- Removed ID selectors in CSS and replaced them with classes. Updated JS references accordingly.
- Precached array lengths in JS for loops (along with other frequently referenced object member lookups) for a [performance improvement](http://stackoverflow.com/questions/6973942/is-optimizing-javascript-for-loops-really-necessary).
- More efficient temperature converter.
- Lots of other similar small JS optimizations.

**Other changes similar to my [last PR](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1341):**
- Made code formatting consistent (now only follows one style for things like whitespace, quotes, etc).
- Refactored JS to use function statements rather than function expressions where possible. This has the following benefits:
  - This is the most readable way to obey the [onevar rule](http://wonko.com/post/try-to-use-one-var-statement-per-scope-in-javascript).
  - This will give a named function in a stack trace in the debugger rather than an anonymous function, which makes debugging easier.
  - There is a [small performance advantage](http://jsperf.com/fn-expression-vs-statement).

**Questions:**
- While everything seems to work in my testing, these changes are more extensive than the last PR I opened. Can anyone share a good list of test cases aside from things like "forecast tomorrow" and "forecast next week" etc?
- `forecast_icons.handlebars` currently prefetches two different graphics (for both the light and dark themes) for all users regardless of which theme is selected. I think it would be better to engineer a way of fetching only the graphics that are relevant to whichever theme the user has selected, but I don't think that's possible purely from within the spice code. Can anyone speak to that?


kethinov:
@moollaza and @sdougbrown, any thoughts on my last post?

sdougbrown:
The theme switching thing is a problem that needs to be solved holistically, and right now this is the only edge case where we're doing this so it's not yet worth the effort to re-address.   It doesn't make sense to regress on purpose, either.

kethinov:
Yeah, I don't think it can or should be done from within the spice, at least right now anyway. Support for fetching only the relevant images would have to be baked into the downstream JS that controls themes. After that we could refactor the template to not fetch two of each image at page load time, but support for this in the theme JS is certainly a dependency.

jagtalon:
@kethinov changing the temperature doesn't seem to be working on mobile. It works on desktop though:

screen shot 2015-01-19 at 4 49 12 pm

jagtalon:
@kethinov it looks like it doesn't work well in the dark theme:

screen shot 2015-01-19 at 4 51 43 pm

Here's the one that's live for comparison:

screen shot 2015-01-19 at 4 51 49 pm

jagtalon:
@MariagraziaAlastra @moollaza @mattr555 @MrChrisW would like to request some help reviewing / testing this if you're free since a lot of code has changed and this IA triggers a lot. Thanks!

kethinov:
Thanks for pointing that out! I'm gonna run a full battery of browser / OS / environment tests on it, iron out the issues, and update this PR ASAP.

jagtalon:
@kethinov sweet that sounds good. This IA gets used a lot so it's important that it works well. :)

jagtalon:
@kethinov just tell us if you're ready for a review!

jagtalon:
@kethinov a bunch of changes were made to Forecast IA--make sure you incorporate those in!

moollaza:
@kethinov apologies for the delay -- are you still interested in finishing this? It looks like the Forecast Spice has changed a bit and so the PR is unmergable. Would you be interested in re-basing this and addressing the conflcits?

kethinov:
At this point I think the fastest way would be to close this PR, refork it, and redo the changes. Trying to rebase and sort out the merges would be tricky.

moollaza:
>At this point I think the fastest way would be to close this PR, refork it, and redo the changes. Trying to rebase and sort out the merges would be tricky.

I definitely agree. I'm very sorry I didn't around to merging this much sooner. When the new PR comes in, I promise I'll get around to it as soon as possible!

Thanks again for taking the time to contribute, we really appreciate it!

Issue #1373, opened 2014-12-23
CodeSearch example query doesn't work


jdorweiler:
javascript console.log example

MrChrisW:
Working. Fixed in #1376 (Internal Issue)

PR #1387, opened 2015-01-01
CodeSearch: Change endpoint from http to https.

jagtalon:
@MrChrisW @jdorweiler It looks like it wasn't working because the HTTP endpoint was redirecting to the HTTPS endpoint, and we don't have anything configured to handle that. Switching to HTTPS fixes the problem.

screen shot 2015-01-01 at 1 53 55 pm


jagtalon:
This fixes https://github.com/duckduckgo/zeroclickinfo-spice/issues/1373

jagtalon:
CC @MrChrisW

MrChrisW:
@jagtalon We'll need to update the endpoint to  https://searchcode.com/ not https://searchco.de as the SSL cert is for the .com domain. After that it's good to go!

jagtalon:
@MrChrisW updated!

MrChrisW:
LGTM :+1:

MrChrisW:
@jagtalon Not sure why the `Remove CouponMountain IA` commit 78dae69 is showing up in this... Anyway!

Issue #1389, opened 2015-01-02
Timer: doesn't always appear in production, but ok in DuckPAN


mattr555:
It probably has something to do with the `hidden` attribute setting [here](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/timer/timer.js#L16), as the html is loaded, but the hidden attribute is set:

![screenshot - 01012015 - 10 28 44 pm](https://cloud.githubusercontent.com/assets/4411471/5594063/95417748-9205-11e4-8344-8b9f723138ca.png)

This appears to be a race condition, as it works sometimes (notably on refresh). I'm not convinced the flash of unstyled content is really a bug...
cc @jagtalon @kethinov

kethinov:
Looks like the JS (https://duckduckgo.com/share/spice/timer/1078/timer.spice.js) is loading prior to the CSS (https://duckduckgo.com/share/spice/timer/1078/timer.css), which isn't the case in DuckPAN.

Due to that, [the line you noted](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/timer/timer.js#L16) never executes because the element the `load` event is supposed to attach to isn't yet present in the DOM.

I just submitted a fix here: https://github.com/duckduckgo/zeroclickinfo-spice/pull/1390

PR #1390, opened 2015-01-02
Fix for issue #1389

kethinov:
See #1389 for background.

I tried using the jquery `on('load', ...)` approach to detect the injection of the `link` element, but it wasn't working, thus why I used the `setInterval` approach instead.

### DIFF
--- a/share/spice/timer/timer.js
+++ b/share/spice/timer/timer.js
@@ -13,7 +13,25 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
     env.ddg_spice_timer = function(api_result) {
         
         //prevent flash of unstyled content if spice shows before css loads
-        $('link[href$="timer.css"]').load(removeHiddenAttr);
+        function waitForCss() {
+            var interval = setInterval(findTimerCss, 10),
+                tries = 0;
+
+            function findTimerCss() {
+                tries++;
+                if ($('link[href$="timer.css"]')[0]) {
+                    clearInterval(interval);
+                    removeHiddenAttr();
+                }
+              
+                //give up after 30 seconds
+                else if (tries > 3000) {
+                    clearInterval(interval);
+                }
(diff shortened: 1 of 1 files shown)


share/spice/timer/timer.js
```diff
@@ -13,7 +13,25 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
...
         
         //prevent flash of unstyled content if spice shows before css loads
-        $('link[href$="timer.css"]').load(removeHiddenAttr);
+        function waitForCss() {
+            var interval = setInterval(findTimerCss, 10),
+                tries = 0;
+
+            function findTimerCss() {
+                tries++;
+                if ($('link[href$="timer.css"]')[0]) {
```
mattr555: @kethinov This didn't seem to fix the flash, perhaps when this condition is met, you should attach the `removeHiddenAttr` function rather than call it?

out

bsstoner:
This is an interesting problem... Probably something we should eventually look at solving at a lower level for all spices? Maybe we should consider blocking JS execution entirely until all the CSS for the spice loads?

share/spice/timer/timer.js
kethinov: Thanks, doh, sorry I missed that. PR updated.

kethinov:
I agree @bsstoner, this is a lot of fuss for individual spices to have to worry about. This probably is not the only spice that has the flash of unstyled content issue either.

In addition to exploring a global block of the spice JS from executing before the CSS loads, another option might be to simply provide some kind of interface that allows spices to opt-in to this kind of delay like what we've built here, except at the API level by abstracting the "is the CSS loaded" JS detection logic and providing a user-definable method fired when the detection registers.

mattr555:
@bsstoner Yeah, this is something we should look at.

The current fix works for me, thanks @kethinov! :+1:

PR #1393, opened 2015-01-04
Today in History (New design)

killerfish:
@jagtalon @abeyang Apologies, for the delay (my machine's charger got fried). So here is the new implementation, without the filters, let me know if there are any issues.

//cc @chrismorast @bsstoner 

todayinh

------
> IA Page: https://duck.co/ia/view/today_in_history


abeyang:
I agree with @jagtalon . Also, @killerfish , is the new width of the tiles being automatically calculated by the content, or is the expanded size the same for all?

jagtalon:
@abeyang do you think it should even expand or just be like https://duckduckgo.com/?q=nginx+vs+apache&ia=qa? It's a bit non-standard.

abeyang:
@jagtalon I agree it's non-standard, but we can't just use ellipses because clicking on the tile won't take you anywhere (from what I remember -- please correct me if I'm wrong). Thus, we won't ever see the full content for those that are truncated.

@killerfish can we also add underlines to the links **only** on mouseover? And make default text (non-links) #aaa instead of #888?

jagtalon:
@abeyang what if it was at the bottom like the others?

screen shot 2015-01-07 at 2 10 44 pm

jagtalon:
@killerfish 
today_in_history_at_duckduckgo (thanks to @jkanarek)

share/spice/today_in_history/content.handlebars
```diff
@@ -0,0 +1,2 @@
+<div class="year"><div class="title_year h-m text--primary">{{wiki_text_to_html year 0}}</div></div>
+<div id="event" class="string t-m">{{wiki_text_to_html str 1}}</div>
```
jagtalon: "event" isn't very descriptive and could collide with some other thing. `id`s should be unique too

abeyang:
Sorry for all the back and forth, @killerfish -- I'm gonna try to help you out here by consolidating some feedback. 

**First**, can we make these changes inside the css:

```css
.zci--today_in_history .string {
    color: #aaa;
    line-height: 1.3;
    height: 9em;
}
.zci--today_in_history .year {
    margin-top: -0.4em;
    margin-bottom: 1.5em;
}
.zci--today_in_history .colchange:hover {
    text-decoration:underline;
}
.t-m { font-size: 1.1em; }
```

**Second**, can we change the tab and header to look more like this (note that date is in the header, not the tab):
screen shot 2015-01-07 at 12 55 46 pm

**Third**, can we try to have the tiles show full content and dynamically adjust to fit? So, I think the default tile size is fine, but anything bigger will auto-expand, rather than having to use the +/- icons? We're not sure if this is the right approach, but we just want to see how this will look & feel.

Thanks! LMK if you have any questions!

killerfish:
guys, thanks for the feedback, ill update accordingly, @jagtalon if you catch any other parsing errors like that one, please let me know.

killerfish:
@abeyang regarding the third point you made, i think the +/- do a better job, dynamic fits would cause inconsistent expansions across the different tiles, which imo wont fit right. I'll put up an ss, for discussion.

jagtalon:
@killerfish no problem. Just ask if you have any questions!

abeyang:
Hi @killerfish, yeah, you might be right, but we'd like to just see how it'll look. Let us know when you've updated the code so we can try it out on our end!

killerfish:
Updated code, lmk your thoughts ^^

jagtalon:
@killerfish will put your code up!

jagtalon:
@killerfish @abeyang it's up on https://ddh1.duckduckgo.com/?q=today+in+history

- Content getting cut off and parsing error:

screen shot 2015-01-15 at 6 59 11 am

screen shot 2015-01-15 at 7 00 17 am

- Date probably shouldn't be in the tab name. It's taking up space. Should be in the "Showing 47 Historical Events" line.

screen shot 2015-01-15 at 7 00 44 am

jagtalon:
@killerfish things disappear when changing the theme:

screen shot 2015-01-15 at 7 02 58 am

jagtalon:
@killerfish instead of adding your own colors, you can use the class `tx-clr--dk` like in https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/currency/detail_mobile.handlebars#L11 (@moollaza should be in the docs, I think)

killerfish:
Yea height should be a multiple of line-height, so 9.1 would work. I'll make rest of the changes (will go through wikitext notation, so we dont get parsing issues).

abeyang:
@killerfish @jagtalon actually, the css changes I've suggested should have taken the text into account. Need to make a few more in order to properly see the changes:

- Please **remove** the float:left in .zci--today_in_history .title_year
- Move the +/- div so that it's in front of the year title, like so:

screen shot 2015-01-15 at 11 32 06 am

Tiles should look like this:

screen shot 2015-01-15 at 11 34 38 am

**Also, please make the tab and header to look like this:**

Tab and Title

Finally, @killerfish can we try auto-expanding the tiles dynamically, w/o the need of hitting the +/- icons? :smile:

killerfish:
@abeyang the last commit has the tiles expanded dynamically, w/o the +/-  and its up on https://ddh1.duckduckgo.com/?q=today+in+history. @jagtalon can you confirm if its still cutting off? Im updating the rest, though i wont be able to cover all the parsing cases immediately, only the cases which would be relevant to us. (there are too many cases, and the docs are scattered around)

abeyang:
@killerfish oh hm, my bad. I think I was looking at a cached version then -- you're right. Let me poke around at that, and will get back to you.

For now, can we bold the date in the metabar, and use month day (you have day month right now). Thanks!

abeyang:
--along with **removing** the float:left in .zci--today_in_history .title_year

abeyang:
--sorry, can we also have these changes too.. change the **height** and **color** to match here:

http://cl.ly/image/3b452z1E3p3k

Overall, we're really digging this -- including the auto-expand. It's really, really a solid work, so totally appreciate it. Just as a head's up, we *may* need to change the link colors due to some pushback internally (though the Design Team really digs black), but anyways, just so that you know.

Good job! :+1:

abeyang:
--ack! One more (sorry, scatter-brained today). Can we not have the title be a link, but instead, make the whole card clickable (basically, the card will go to where the title is currently going to). Thanks!

killerfish:
@abeyang are we sticking with excluding the +/- or should i put that back in?

jagtalon:
@abeyang @killerfish cool--will set it up on a server and check if everything is good!

jagtalon:
@killerfish yeah it looks like I don't see one getting cut off today

abeyang:
@killerfish let's leave the +/- out -- it's looking pretty good w/o them. Let us know when you've updated your code!

jagtalon:
@abeyang @killerfish Man I completely missed this, but there's one case where having variable tile sizes won't work:

screen shot 2015-01-29 at 1 33 49 pm

Not sure what a good solution is--will think about it as well.

@killerfish to make it work on the dark theme, add the `tx-clr--dk` class:

screen shot 2015-01-29 at 1 35 25 pm

abeyang:
@killerfish and @jagtalon . Is it possible to **disable** the grid view for this IA? I know it's not conventional for DDG, but might be permissible in this case because it is a timeline of sorts.

bsstoner:
@abeyang @jagtalon @killerfish We don't have a way to explicitly disable it. We do have a parameter for changing the threshold of when it's enabled. You could just pass a really large value for `meta.minItemsForModeSwitch` variable. As long as the # of items is less than that minimum it will never enable the grid button.

For reference, Airlines uses this option to enable grid mode after fewer items: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/airlines/airlines.js#L98

abeyang:
@bsstoner Awesome!

@killerfish can we just do that in this case? Let's put a really large number (9999?) for `meta.minItemsForModeSwitch`

killerfish:
updated all changes, please review, should be good

jagtalon:
@killerfish woot! Checking it out

jagtalon:
@killerfish Nice! I just have three comments:

- There's a space in the `<a>` tag making it underline the space.

screen shot 2015-02-19 at 2 39 44 pm

- Looks like this one just highlighted "Serf" instead of "Serfdom"

screen shot 2015-02-19 at 2 40 21 pm

- Tiles should probably have the same sizes on mobile :/

screen shot 2015-02-19 at 2 48 34 pm


@abeyang it's up in https://ddh9.duckduckgo.com/?q=today+in+history&ia=todayinhistory.

jagtalon:
@killerfish sorry about that--thought that the variable tile size would be easy :/

moollaza:
Looks good to me. Nicely done team! Would like a design sign-off and then we can get this one live!

moollaza:
This is a problem we run into often, so perhaps not a fix for this IA in particular, but the MetaBar is getting cut short on mobile view so it looks a little odd:

2015-02-20_00h18_13

Not much we can do to fix that other than changing "Historical Events" to "Events"

killerfish:
@jagtalon thats okay, im just pondering over how we can fit this for mobile. Thanks for pointing out the cases, ill update.

killerfish:
@jagtalon we could always switch back to the +/- button to expand/detract, which might be better for mobile view?

abeyang:
@killerfish you're right -- let's add the `+/-` buttons back in. Can we also make the border of the circles a tad darker -- maybe try `#ccc`?

abeyang:
^ Actually, we have to use rgba to accommodate for darker tiles. Let's try: `rgba(50,50,50,0.2)`. Thanks!

killerfish:
@abeyang done :bangbang: this should do for mobile view also, though any other layout changes, lmk

moollaza:
@killerfish sorry about the delay! I just noticed you were waiting for us on this. I'll take a look..

lib/DDG/Spice/TodayInHistory.pm
```diff
@@ -1,33 +1,85 @@
 package DDG::Spice::TodayInHistory;
+# ABSTRACT: List events which occured in history given current day
 
 use DDG::Spice;
+use DateTime;
+use DateTime::TimeZone;
+use Time::Piece;
```
moollaza: Do any of these need to be added to the dist.ini?

killerfish: Dont think so, DateTime is already included, and Time::Piece is in core.

moollaza:
@killerfish @jagtalon @abeyang this LGTM :+1: 

I noticed a few small bugs we can deal with after merge:

1. The +/- button isn't a perfect circle, looks like a width of `1.4em` fixes that (from `1.5em`)
2. If you expand several tiles, we aren't updating the width of the carousel so the last tile gets cut off.
3. If the last tile is an expandable one (it is today), it gets cut off after expanding (same problem)
4. In mobile view, expanding a tile forces it onto a new line. Shrinking it doesn't put it back, making the IA container taller. Perhaps on mobile we should increase the height of the tiles, not the width?

I really want to merge this because those are fairly small issues. Just need a final OK from @abeyang on the design.

http://pizza-emerald.codio.io:5000/?q=today+in+history&iax=1&ia=todayinhistory

moollaza:
Actually, I just found another issue which should be addressed -- if you switch to another tab, then come back to TodayInHistory - we duplicate the +/- icons!

2015-03-19_09h56_24

moollaza:
Also, looks like the content is getting clipped on iPhone:

2015-03-19_10h05_08

I guess we can increase the tile height?

moollaza:
aaand one more thing. I expected the tile to be a clickable link -- is it supposed to be? If so we need to set a "url" property for each item.

killerfish:
@moollaza tile wont be a clickable link, ive fixed the duplication of buttons, and their width. Would need @abeyang to advise on tile height change.

moollaza:
Up on DDH2! `https://ddh2.duckduckgo.com/?q=today+in+history&ia=todayinhistory

abeyang:
Hi @killerfish , I still see clipping on the tiles in desktop view. Can we modify the css a bit:

``` css
.zci--today_in_history .year {
  margin-bottom: 3.2em;
}

.zci--today_in_history .string {
  height: 9.1em;
}
```

This also fixed the mobile view -- at least it did for me. LMK how that goes for you...

moollaza:
@killerfish hold off on that -- looks like the DDH2 machine is outdated. I'll redeploy and ping you when it's ready

moollaza:
Alright @killerfish @abeyang, update DDH2 and redeployed: https://ddh2.duckduckgo.com/?q=today+in+history&ia=todayinhistory

jagtalon:
@killerfish Nice, I like it! Please merge with master to fix the conflicts.

@moollaza looks like ddh2 is outdated--please update. :)

moollaza:
Everything should be back now on DDH 2

abeyang:
Thanks @moollaza .

@killerfish , in addition to what [I said earlier](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1393#issuecomment-90988851), can we make these css changes:

``` css
.zci--today_in_history .circle_p {
  border: none; /* delete the border completely */
  text-align: right;
  margin-top: 0.2em;
}
```

Moreover, for that div element, can we add this class to it: `tx-clr--grey-light`. Thanks!

killerfish:
@abeyang the circle plus year changes look good, but using height as 9.1em for string chops off the string.

abeyang:
@killerfish can you check in your changes so we can take a look?

killerfish:
cutofftext

moollaza:
@killerfish terribly sorry about how long this IA has gone on for...

In the interest of finally getting this live, let's skip the tile expansion for now. The frontend team will be bringing this functionality to tiles soon, but I don't have an ETA. We can simply trim the text for now as other IAs do.

Let's get this live!

share/spice/today_in_history/today_in_history.js
```diff
@@ -1,33 +1,158 @@
...
+        
+        Spice.add({
+            id: "today_in_history",
+            name: "Today in History",
+            data: array,
+            meta: {
+                itemType: "historical events for " + ourquery,
+                sourceUrl: 'http://en.wikipedia.org/wiki/'+ ourquery,
+                sourceName: 'Wikipedia',
+                minItemsForModeSwitch: '9999',
```
bsstoner: There's a nicer way to do this now. You can either disable it with this:
```
disableModeSwitch: true
```
or you can hide it entirely with this:
```
hideModeSwitch: true
```

moollaza:
Continued in #1988

PR #1394, opened 2015-01-05
Adding GithubLanguages search

joerussbowman:
This adds an instant answer functionality where queries that start or end with a known computer language will automatically do a search for repositories matching the rest of that query, with the language paramater, on GitHub. So, 'javascript redis' would find any repositories for projects written in javascript that relate to redis.

I was asked to submit an Instant Answer by Caine Tighe. Approval and suggestions for approach were provided by @yegg and mr.Shu.

Thank you

### DIFF
--- a/lib/DDG/Spice/GithubLanguages.pm
+++ b/lib/DDG/Spice/GithubLanguages.pm
@@ -1,0 +1,40 @@
+package DDG::Spice::GithubLanguages;
+# ABSTRACT: Write an abstract here
+# Start at https://duck.co/duckduckhack/spice_overview if you are new
+# to instant answer development
+
+use DDG::Spice;
+
+spice is_cached => 1;
+
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
+name "GithubLanguages";
+source "";
+icon_url "https://github.com/favicon.ico";
+description "Succinct explanation of what this instant answer does";
+primary_example_queries "first example query", "second example query";
+secondary_example_queries "optional -- demonstrate any additional triggers";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#category
+# category "";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
+# topics "";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GithubLanguages.pm";
+attribution github => ["GitHubAccount", "Friendly Name"],
+            twitter => "twitterhandle";
+
+# Triggers
+my @triggers = share("triggers.txt")->slurp;
+triggers startend => @triggers;
+
+# Handle statement
+handle remainder => sub {
+
+    # optional - regex guard
+    # return unless qr/^\w+/;
+
+    return unless $_;    # Guard against "no answer"
+
+    return $_;
+};
+
+1;
--- a/lib/DDG/Spice/GithubLanguages.pm
+++ b/lib/DDG/Spice/GithubLanguages.pm
@@ -31,7 +31,8 @@ handle query_lc => sub {
     # structuring the search this way gives the first langauge specified
     # precedence. So have 'python javascript' passes the query
     # 'javascript language:python' to github.
-    if ($query =~ /^["']*($langs)\b/ || $query =~ /\b($langs)["']*$/) {
+    #if ($query =~ /^["']*($langs)\b/ || $query =~ /\b($langs)["']*$/) {
+    if (/^["' ]*($langs)\b/ || /\b($langs)["' ]*$/) {
         $l = $1;
     } 
     
--- a/lib/DDG/Spice/GithubLanguages.pm
+++ b/lib/DDG/Spice/GithubLanguages.pm
@@ -28,6 +28,8 @@ triggers startend => @triggers;
 
 my $langs = join("|", @triggers);
 
+spice to => 'https://api.github.com/search/repositories?q=$1&callback={{callback}}';
+
 # Handle statement
 handle query_lc => sub {
 
@@ -35,7 +37,7 @@ handle query_lc => sub {
 
     # optional - regex guard
     # return unless qr/^\w+/;
-    my $lang = m/^\b$langs\b|\b$langs\b$/;
+    my $lang = $query =~ /^$langs\b/;
     $query =~ s/^\b$langs\b|\b$langs\b$//;
     #return unless $_;    # Guard against "no answer"
 
--- a/lib/DDG/Spice/GithubLanguages.pm
+++ b/lib/DDG/Spice/GithubLanguages.pm
@@ -9,39 +9,39 @@ spice is_cached => 1;
 
 # Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
 name "GithubLanguages";
-source "";
+source "GitHub";
 icon_url "https://github.com/favicon.ico";
-description "Succinct explanation of what this instant answer does";
-primary_example_queries "first example query", "second example query";
-secondary_example_queries "optional -- demonstrate any additional triggers";
+description "Github search by language";
+primary_example_queries "javascript redis", "mongodb javascript";
 # Uncomment and complete: https://duck.co/duckduckhack/metadata#category
-# category "";
+category "programming";
 # Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
-# topics "";
+topics "programming";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GithubLanguages.pm";
-attribution github => ["GitHubAccount", "Friendly Name"],
-            twitter => "twitterhandle";
+attribution github => ["https://github.com/joerussbowman", "Joe Bowman"],
+            twitter => "joerussbowman";
 
 # Triggers
 my @triggers = share("triggers.txt")->slurp;
 triggers startend => @triggers;
 
+chomp(@triggers);
 my $langs = join("|", @triggers);
 
-spice to => 'https://api.github.com/search/repositories?q=$1&callback={{callback}}';
+spice to => 'https://api.github.com/search/repositories?q=$1&sort=stars&callback={{callback}}';
 
 # Handle statement
 handle query_lc => sub {
 
     my $query = $_;
+    my $l = ""; 
+    if ($query =~ /^($langs)\b/ || $query =~ /\b($langs)$/) {
+        $l = $1;
+    } 
+    
+    $query =~ s/^($langs)\b|(\b$langs)$//;
 
-    # optional - regex guard
-    # return unless qr/^\w+/;
-    my $lang = $query =~ /^$langs\b/;
-    $query =~ s/^\b$langs\b|\b$langs\b$//;
-    #return unless $_;    # Guard against "no answer"
-
-    return "$query language:$lang";
+    return "$query language:$l";
 };
 
 1;
--- a/lib/DDG/Spice/GithubLanguages.pm
+++ b/lib/DDG/Spice/GithubLanguages.pm
@@ -20,7 +20,7 @@ my @triggers = share("triggers.txt")->slurp;
 triggers startend => @triggers;
 
 chomp(@triggers);
-my $langs = join("|", @triggers);
+my $langs = join("|", map(quotemeta, @triggers));
 
 spice to => 'https://api.github.com/search/repositories?q=$1&sort=stars&callback={{callback}}';
 
@@ -28,11 +28,11 @@ handle query_lc => sub {
 
     my $query = $_;
     my $l; 
-    if ($query =~ /^($langs)\b/ || $query =~ /\b($langs)$/) {
+    if ($query =~ /^["']*($langs)\b/ || $query =~ /\b($langs)["']*$/) {
         $l = $1;
     } 
     
-    $query =~ s/^($langs)\b|\b($langs)$//;
+    $query =~ s/^["']*($langs)\b["']*|["']*\b($langs)["']*$//;
     
     # make sure there is an actual query, and not just a language term search
     for ($query) {
@@ -43,7 +43,7 @@ handle query_lc => sub {
         }
     }
     
-    return "$query language:\"$l\"";
+    return "${query} language:\"${l}\"";
 };
 
 1;
--- a/lib/DDG/Spice/GithubLanguages.pm
+++ b/lib/DDG/Spice/GithubLanguages.pm
@@ -26,15 +26,20 @@ attribution github => ["GitHubAccount", "Friendly Name"],
 my @triggers = share("triggers.txt")->slurp;
 triggers startend => @triggers;
 
+my $langs = join("|", @triggers);
+
 # Handle statement
-handle remainder => sub {
+handle query_lc => sub {
+
+    my $query = $_;
 
     # optional - regex guard
     # return unless qr/^\w+/;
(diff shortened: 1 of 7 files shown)


lib/DDG/Spice/GithubLanguages.pm
```diff
@@ -0,0 +1,47 @@
...
+
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
+name "GithubLanguages";
+source "GitHub";
+icon_url "https://github.com/favicon.ico";
+description "Github search by language";
+primary_example_queries "javascript redis", "mongodb javascript";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#category
+category "programming";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
```
MrChrisW: Please remove any unnecessary comments.

share/spice/github_languages/github_languages.js
```diff
@@ -0,0 +1,69 @@
...
+    env.ddg_spice_github_languages = function(api_result){
+
+        if (!api_result || !api_result.meta.status === 200) {
+          return Spice.failed('github');
+        }
+
+        // There's a next linkpage link provided by the api, this converts it
+        // to a link to the next page on the actual github search, so by clicking
+        // More on Github you automatically get taken to the next page of results
+        var sourceUrl = api_result.meta.Link[0][0].
```
MrChrisW: Some API results. Example query [`game maker language`](https://api.github.com/search/repositories?q=%20language%3Agame%20maker%20language&sort=stars&callback=ddg_spice_github_languages) do not return any link results meaning this will fail. You'll need to ensure the `api_result.meta.Link` property exists before using it.

lib/DDG/Spice/GithubLanguages.pm
```diff
@@ -0,0 +1,47 @@
...
+chomp(@triggers);
+my $langs = join("|", @triggers);
+
+spice to => 'https://api.github.com/search/repositories?q=$1&sort=stars&callback={{callback}}';
+
+# Handle statement
+handle query_lc => sub {
+
+    my $query = $_;
+    my $l = "";
```
MrChrisW: No need to use = "" to initialize this var. `my $l;` will work fine

MrChrisW:
@joerussbowman Thanks for your contribution :+1: I've added some feedback to your code.

MrChrisW:
@joerussbowman 
I don't think this IA should trigger if the search contained no additional words beyond the trigger example: `javascript` should not trigger but `javascript node` should. You'll need to remove your word triggers and check if there's anything left to search for. 

This may also overlap with the **search code** IA.

joerussbowman:
Thanks Chris, I'll hit these sometime this week, when I can get some time to work on it. Appreciate the fast review.

MrChrisW:
No problem take your time.

Feel free to let me know if you have any questions :+1:

share/spice/github_languages/github_languages.js
```diff
@@ -0,0 +1,69 @@
...
+        // Render the response
+        Spice.add({
+            id: "github_languages",
+
+            // Customize these properties
+            name: "GitHub Language Search",
+            data: api_result.data.items,
+            meta: {
+                sourceName: "GitHub",
+                sourceUrl: sourceUrl,
```
jagtalon: Remove trailing commas because it won't work on IE 8

lib/DDG/Spice/GithubLanguages.pm
```diff
@@ -0,0 +1,40 @@
...
+chomp(@triggers);
+my $langs = join("|", @triggers);
+
+spice to => 'https://api.github.com/search/repositories?q=$1&sort=stars&callback={{callback}}';
+
+handle query_lc => sub {
+
+    my $query = $_;
+    my $l; 
+    if ($query =~ /^($langs)\b/ || $query =~ /\b($langs)$/) {
```
jagtalon: It looks like this triggers with just the language in the query such as "javascript". Is that intended?

jagtalon: Maybe it should trigger with "github javascript", too?

jagtalon:
@joerussbowman The description text could overflow:

screen shot 2015-01-08 at 1 18 19 pm

joerussbowman:
Thanks guys. 

I'm delayed on working on this, busy week for me. 

I intend to not have the instant answer run on a language only search term. The last thing I was working on earlier this week was I caught a big where the whole thing blew up if the query included quotes. 

I did think about changing the term to require github, ie : 'github javascript redis' but then was concerned about conflict with github instant answer.  I was also considering using '$language on github'  which probably makes the most sense. The question then being will it get triggered much? 

I pulled the css from the github spice, it might be worthwhile to flag the overflow issue there as well. 

Appreciate the front end review jagtalon, thank you.

jagtalon:
@joerussbowman alright cool. For 'github javascript redis', you could probably modify the github instant answer to not trigger on that, but you could probably do that in the next iteration instead.

joerussbowman:
@jagtalon can I ask how you are doing the css testing? I see you're using something that looks like it's emulating an iphone? The reason I ask is in both chrome and firefox (on linux, my only test platform) the box size is large enough that there's no chance of overflow. I'm not sure I have anything available to me to do that testing? Sorry, I'm an ops guy, I don't do front end html work at that level very often.

joerussbowman:
btw, I see tests are failing, I actually need to do a huge update to the tests because I've been tracking down lots of issues popping for terms like c++ and long terms like game maker language. I'll be overhauling the tests before I resubmit the pull request.

MrChrisW:
Hey @joerussbowman I tend to use this [online tool](http://quirktools.com/screenfly/#u=http%3A//duckduckgo.com&w=320&h=568&a=37) to test different screen sizes. It uses an iframe so it will work with localhost.

jagtalon:
@joerussbowman I personally use https://developer.chrome.com/devtools/docs/device-mode and browserstack.com.

Looking forward to your changes! Just ping us when you're ready.

jagtalon:
@joerussbowman need any help from us?

joerussbowman:
@jagtalon I'm not sure yet :)

I'm working two issues on this, primarily the first right now.

1. Testing all the languages I found I have a few incoming queries that don't get handled correctly. The problem strings are c++, f# and one other one with ( characters. I think it has to do with the prefix regex match. I chatted with Caine Tighe about this and he suggested another way to approach parsing the request string so I'm going to look at that.

2. I validated that the request being generated by the duckpan server to the github api generates results, but they're not displaying. My experience with the query parsing means I'm going to need to stop and test all the languages at least once before I continue, so I'll have to write a script to handle to handle that. 

If you guys have any ideas that strike you because of the problem description I'd be glad to hear them. Otherwise I'll keep plugging away. Tomorrow I start about 4 hours earlier than normal for our maintenance window, I probably won't touch this until tomorrow afternoon if I'm not too tired. 

Thanks for checking in and offering help, I appreciate it.

jagtalon:
Cool! Looking forward to it!

> I validated that the request being generated by the duckpan server to the github api generates results, but they're not displaying

What do you mean by generating results but not displaying? What query is it?

joerussbowman:
The problem I'm seeing is when I run duckpan server on my codio and do a search for "game maker language" I'm not seeing any instant answer results displayed.

However, when I go to the terminal and copy the url the server made a request to and try it in my browser I see github did provide results. The url is - https://api.github.com/search/repositories?q=test%20language%3A%22game%20maker%20language%22&sort=stars&callback=ddg_spice_github_languages

However, I have not put a ton of time into this. I'm still working on the query handling as I'm still seeing some issues with parsing the request. I'm focusing on that first.

joerussbowman:
Ok, I think I am a little stuck here, there's obviously a problem with my regex when searching for certain langugaes as a prefix, works great when they're at the end though. As far as queries go if I switched to just having the language match at the end it would pass. If someone better at regex can identify why the prefix search is failing though, I'd appreciate it. I might have to hit perlmonks or something. 

The issue is 'f# test' and 'graphviz (dot) test ' both fail to match.I believe 'c++ test' is having a similar problem and falls back to matching 'c' in the languages list and thus fails the test on the output which becomes a search for '++ test language:"c"'

I have a simple test script I'm using to run the regex against all the possible languages, has to be run in the same directory as the triggers.txt file.

```perl
#!/usr/bin/env perl

my @triggers;

open (TRIGGERS, "triggers.txt") or die "Couldn't open triggers.txt";

while (<TRIGGERS>) {
    chomp $_;
    push(@triggers, $_);
}

close TRIGGERS;

my $regex = join("|", map(quotemeta, @triggers));

for (@triggers) {
    my @queries = ("$_ test", "test $_");
    for (@queries) {
        if (/^["' ]*($regex)\b/ || /\b($regex)["' ]*$/) {
            print "MATCH ${_}\n";
        } else {
            print "FAIL ${_}\n";
        }
    }
}
```perl

(hopefully my markdown there works)

**Comment by joerussbowman** [comment]
btw, I did try a suggestion from Caine where I tried to grab the languages from the query with the regex then test. I could get the failed queries to then fail and just do a return so at least I'd catch most queries at start/end. However, it would still provide inconsistent results for c++ which I did not believe was acceptable, something like this in the @queries loop

```perl
        my @langs = /^["' ]*($regex)\b/ || /\b($regex)["' ]*$/;
        if (scalar(@langs) < 1) {
            print "NO LANGUAGE MATCH ${_}\n";
        }

```

Whereas for the actual test I'd just do a return rather than printing the fail message.

jagtalon:
@joerussbowman hey! Sorry for the late reply--it looks like things are in order now? Do you still need help with the regexp?

screen shot 2015-01-29 at 1 30 07 pm
screen shot 2015-01-29 at 1 30 26 pm

joerussbowman:
No, I figured out the regex. I see the build is reporting as failed but when I read the output from the failed builds it looks like there's a general problem and not something specific to my contribution?

moollaza:
@joerussbowman yup there was a problem before, but I'm running the build again and I expect it should pass.

jagtalon:
@joerussbowman sweet! Checking your IA then. :)

jagtalon:
@moollaza I have a few concerns here and I'm still thinking about it, but I'd love to hear your input as well:

- I think the tab name should just be "Software" and have "Showing 30 Repositories Using C#" in the metabar. I wonder if people will get the functionality of the IA with this change.
- It comes up a lot:

screen shot 2015-01-30 at 9 07 01 am

screen shot 2015-01-30 at 8 51 41 am

screen shot 2015-01-30 at 8 58 50 am

screen shot 2015-01-30 at 8 59 20 am

joerussbowman:
Couple things 

1. Maybe instead of software, it should be "Repositories"? 

2. Should the triggers list be tweaked? I just took what github had in their search form select. Maybe some could be removed. As a *nix ops guy I generally would search for bash or zsh, not shell for example.

share/spice/github_languages/github_languages.js
```diff
@@ -0,0 +1,75 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_github_languages = function(api_result){
+
+        if (!api_result || !api_result.meta.status === 200) {
+          return Spice.failed('github');
```
moollaza: should be: `return Spice.failed('github_languages');`

share/spice/github_languages/github_languages.handlebars
```diff
@@ -0,0 +1,4 @@
+<span>{{name}}</span>
```
moollaza: Looks like this isn't being used?

share/spice/github_languages/footer.handlebars
```diff
@@ -0,0 +1,7 @@
+<div class="tile__update__info one-line">
+  <i class="star starIcon"></i>
+  {{watchers}} &middot; Updated {{GitHub_last_pushed pushed_at}}
+</div>
+
+
+
```
moollaza: Let's remove the extra whitespace here

share/spice/github_languages/github_languages.js
jagtalon: Oh please check the existence of `meta` as well before accessing `status`

joerussbowman:
Guys,

I just want to say how awesome of an experience it is to have someone doing reviews and catching these things vs the last 5+ years I've spent just hacking at code alone. I apologize if my attention to detail isn't quite up to snuff, this is the first code I've worked on in a few months as my job has had me focused on Windows engineering and running through a lot of procedural stuff. Plus, I've always been more ops than programmer. 

I'll work on these over the weekend as time permits and try to have a new commit by Monday morning. If I could get some final direction on what you guys would like to see for the title/metabar that would be helpful. 

Thanks!

moollaza:
@joerussbowman I agree on changing the tab to "Repositories" or even "Software"

Regarding the triggers, I'm also a concerned about how many of the trigger languages are dictionary words like "python", "assembly", "gas", "shell", "crystal", "dylan", "dart", "cucumber", etc.

I see two ways we can ensure this doesn't pop a bad result:
1. Force a smaller query space by requiring "github" or "code" or some other word that indicates this is a "code" search
2. If the matched language is a dictionary word, give the IA a `signal: 'low'` (unless #1^ is true), so that it doesn't display until clicked.

Also, did we consider adding this to the existing GitHub Spice by allowing for a `lang` keyword in the query? e.g "github calculator lang:javascript"

joerussbowman:
iirc the only discussion we had about the GitHub spice was to reconfigure the github instant answer to not trigger if a language was defined.

Now though, I'm wondering if this doesn't actually make more sense to actually not be an IA of it's own and instead have the functionality moved to the existing GitHub Spice.

It could probably be a lot more flexible that way. The query could be inspected to see if a language is mentioned, then have it set the language parameter.

ie: github javascript calculator, github calculator javascript, calculator javascript github... -> 'calculator lang:javascript' search via GitHub api.

I think this approach would probably address all issues and make a lot more sense.

moollaza:
@joerussbowman I totally agree, I was on the verge of suggesting that, but didn't want to make you throw away the code you've written and the time you've spent on this. Overall though, I think that approach would definitely be better. Also, we're going to soon have a filtering dropdown on the UI of Spice IA's, which at the very least would enable you to filter the intial results based on some property (like the language). Potentially though, it could be used to do a full requery of the API which would be amazing.

Then you could search "github calculator", change the dropdown to "JavaScript" and get results that match your keyword and language.

I'm not sure how far away that functionality is, but regardless we can still use most of this code if it were integrated into the existing GitHub Spice.

joerussbowman:
Well my code/time isn't as important as getting users the best results for their query, so if that's the solution that makes the most sense I'll work on seeing what I can do to get this functionality in the GitHub Spice. 

I'll start on that this weekend.

jagtalon:
@joerussbowman yeah that sounds like a good solution too!

joerussbowman:
fyi: I didn't forget about this, got stuck working over the weekend. I'll try to poke at it this coming weekend if I can block some time to focus on it.

MrChrisW:
@joerussbowman No problem! Take your time, just ping the team when ready for review!

joerussbowman:
I'll generate a new pull request with the changes to Github spice. If you want to go ahead and delete this request please do. I still have the fork with the code I need to move over.

MrChrisW:
@joerussbowman Sure, I take it your migrating these changes directly into the GitHub IA?

joerussbowman:
Yea, that's the plan. Going to have that one detect if a language is speficied then add the lang: parameter to the request.

MrChrisW:
@joerussbowman Sounds good to me! Thanks again for taking the time to contribute. As usual ping any of the team when you're ready with the new PR :+1: Closing for now.

PR #1396, opened 2015-01-05
Flights: Fix #1377

### LINKED ISSUE #1377: Flights - RouterHelper: Not working in Production
https://duckduckgo.com/?q=delta+Toronto+to+Amsterdam

Looking in the dev panel, the request to `/js/spice/route_helper` returns no results in production, but works correctly in DuckPAN. 

https://duckduckgo.com/js/spice/flights/route_helper/YYZ/AMS/arr/2014/12/26?hourOfDay=10

@tommytommytommy I noticed that RouteHelper.pm is only expecting 1 param to be passed along because there is no `spice from` and only `$1` is being used, but the `router_helper.js` builds a much larger query with several params: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/flights/route/flights_route.js#L59-63

Can you double check if everything is correct?

jagtalon:
Problem was that $1 would be encoded so we need a `spice from` before appending it to the URL.

screen shot 2015-01-05 at 2 18 33 pm

### DIFF
--- a/lib/DDG/Spice/Flights/RouteHelper.pm
+++ b/lib/DDG/Spice/Flights/RouteHelper.pm
@@ -7,9 +7,11 @@ attribution github => ["https://github.com/tommytommytommy", 'tommytommytommy'];
 # cache responses for 5 minutes
 spice proxy_cache_valid => "200 304 5m";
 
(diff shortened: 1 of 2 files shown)


tommytommytommy:
Just tested this out on my instance; looks good to me. Thanks for the debugging help!

jagtalon:
@tommytommytommy Nice!

PR #1398, opened 2015-01-06
news: oops

hunterlang:
i am an idiot!

@bsstoner @jagtalon 

this fixes the barack obama news issue. once we added a story with the entity relaxation, we were bailing on the rest of the stories!


bsstoner:
:+1:

PR #1400, opened 2015-01-07
fix attribution

jdorweiler:
@russellholt @jagtalon  This should fix all the messed up attributions


jagtalon:
@jdorweiler looks good!

Issue #1395, opened 2015-01-05
DNS: Tab name should probably be DNS instead of the domain name


jagtalon:
If you look at the tab, it says "Asana.Com". It should probably be just "DNS" or maybe "Domain Information" to be more descriptive

![screen shot 2015-01-05 at 10 27 21 am](https://cloud.githubusercontent.com/assets/81969/5615064/8fcf7356-94c5-11e4-9934-74bef27e073b.png)

MrChrisW:
This change was introduced in #1128 the previous value was "Answer"

jagtalon:
@MrChrisW Oh I see--I think it might be better for it to be "Answers" mostly because the way the letters are cased in "Asana.Com" and "Duckduckgo.Com" are bothering me haha

MrChrisW:
Fixed in #1401

PR #1401, opened 2015-01-08
DNS: Change name value to "Answer". Add error checking of api_result.

MrChrisW:
+ Fixing #1395 - Changing `name` value to "Answer"
+ Adding additional error checking of api_result. - `if (!api_result)`


jagtalon:
:+1:

PR #1404, opened 2015-01-09
people in space: fix url link

jdorweiler:
@jagtalon The people in space api changed and the url link is broken.  This fixes that.


jagtalon:
:+1:

PR #1412, opened 2015-01-13
Packagist.js - Normalizing downloads with DDG function.

GregorioMartinez:
Changing normalization of download # from regex to DDG.commifyNumber()


jagtalon:
:+1:

PR #1420, opened 2015-01-16
Instagram images by ID

jkv:

Allows a user to search for an instagram short code to view the image and details on it.

A way to search by short code in Instagram without using their API directly.

http://instagram.com/developer/embedding/

Only option.

No.

instagram uZAo9qkgTU
instagram C

Instragrammers

Yes, https://duck.co/ideas/idea/147/instagram-images-by-id

None that I could find.

I don't think so.

Nothing special required. http://instagram.com/developer/embedding/

Searching for new instant answers.

selection_067

## Checklist

```

    - Windows 8
        [x] Google Chrome
        [] Firefox
        [] Opera
        [x] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 7
        [] IE 8

    - Mac OSX
        [x] Google Chrome
        [x] Firefox
        [] Opera
        [x] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [x] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [] Google Chrome
        [] Opera
```


MrChrisW:
@jkv Thanks for your contribution :+1: 

A few notes. 

+ We cannot use the "embedding" feature as this would expose users directly to the Instagram domain resources which we like to avoid due to privacy concerns. Usually any external images are loaded through a DuckDuckGo proxy using the [imageproxy](https://github.com/duckduckgo/duckduckgo-documentation/blob/master/duckduckhack/spice/spice_handlebars_helpers.md#imageproxy) helper.

+  The API doesn't seem to respond with a correctly formated JSON response when it's unable to locate an image using the shortcode provided, instead it simply displays "No Media Match" or "Page Not Found" with a 404 error code. This is a problem as we don't want to be loading a resource that responses with that code for queries like `Instagram site` or `Instagram app`

+ I'd like to better match shortcodes to avoid triggering this every time the keyword is searched however it doesn't look like shortcodes follow any rules in terms of length or types of characters to avoid triggering on searches like `Instagram D`. We may want to only trigger if a shortcode is a certain length.

moollaza:
@jkv Thanks a lot for making this pull request - @MrChrisW pointed out, the proper way to implement this Instant Answer would be to use the Instagram API, but unfortunately it requires OAuth for all API requests, which we don't currently support.

For now i'm going to close this PR because we can't do much more with it. Sorry that we have to throw away your hard work. Are there any other ideas you'd be interested in working on? Please be sure to send us an email before you work on anything so we can make sure your time and effort won't be wasted!

PR #1422, opened 2015-01-16
Whois: Update Style #1414

### LINKED ISSUE #1414: Whois Spice - Additional Updates Needed
We need to make updates to https://github.com/duckduckgo/zeroclickinfo-spice/pull/1331. In the end, it should look like this:

screen shot 2015-01-14 at 2 14 36 pm

Changes:
- Remove byline completely ("whois lookup")
- CSS changes:

``` css
.zci--whois .zci__more-at { padding-top: 0; }
.zci--whois .record td { padding-left: 0; }
.zci--whois .record .record__cell__key { text-transform: capitalize; }
```

Finally, we need to make these changes (not shown):
- Remove colons at the end of each key
- "Expires" should be "Expires On"

And that's it!

MrChrisW:
Fixes #1414 @abeyang  @jagtalon 

selection_070

### DIFF
--- a/share/spice/whois/whois.css
+++ b/share/spice/whois/whois.css
@@ -16,15 +16,17 @@
 }
 
 .zci--whois .zci__more-at {
-    padding-top: 0.75em;
+    padding-top: 0;
 }
 
 .zci--whois .record td {
     padding: 0.4em;
+    padding-left: 0; 
     border-top: solid 1px #DDD;
     border-bottom: solid 1px #DDD;
 }
 
 .zci--whois .record .record__cell__key {
     width: 150px;
+    text-transform: capitalize;
 }
--- a/share/spice/whois/whois.js
+++ b/share/spice/whois/whois.js
@@ -84,7 +84,7 @@
             'Registered to': get_first_by_key(contacts, 'name'),
             'Email': get_first_by_key(contacts, 'email'),           
             'Last updated': prettifyTimestamp(api_result.updatedDate),
-            'Expires': prettifyTimestamp(api_result.expiresDate),
+            'Expires On': prettifyTimestamp(api_result.expiresDate),
             'Registrar': api_result.registrarName,
             'Name Servers': nameServers
         };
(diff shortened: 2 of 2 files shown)


share/spice/whois/whois.css
```diff
@@ -16,15 +16,17 @@
...
 
 .zci--whois .record td {
     padding: 0.4em;
+    padding-left: 0; 
     border-top: solid 1px #DDD;
     border-bottom: solid 1px #DDD;
 }
 
 .zci--whois .record .record__cell__key {
     width: 150px;
```
MrChrisW: Is this `width` fine? It adds spacing between the keys and results.

MrChrisW:
@abeyang I don't believe we can remove colons at the end of each key as this is built into the `records` template.

abeyang:
Thanks @MrChrisW ! We'll take a look at it and get back to you.. will also look into the colons. /cc @jagtalon

jagtalon:
@MrChrisW nice! @abeyang yup, coming right up!

jagtalon:
@abeyang https://ddh1.duckduckgo.com/?q=whois+facebook.com&ia=whois. Looks good to me.

moollaza:
@abeyang this looks great -- should we perhaps consider updating this styling at a template level? So other IA's using the record template will be styled like this?

abeyang:
@jagtalon The keys should be in title case ("Last Updated"), and can we see if we can remove the colons? It seems like the dev is able to do so for the [BPM Goodie](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/803#issuecomment-70484379).

MrChrisW:
@abeyang That's because the Goodie is manually generating the html content and not using the template system like the Spice. :smile:

jagtalon:
@abeyang I updated the "Last Updated"--the colons, however, are at the template level (internally) and can't be changed from here. I can add an Asana task to control that.

https://ddh1.duckduckgo.com/?q=whois+duckduckgo.com&ia=whois

abeyang:
I see. Yes, let's put that as a task (removing colons), but otherwise, looks good! :+1:

jagtalon:
@abeyang nice!

@MrChrisW 

oisodse

PR #1423, opened 2015-01-16
Launchbug: fixed domain name

puskin94:
@jagtalon I fixed the wrong domain name, can you merge it?

### DIFF
--- a/share/spice/launchbug/launchbug.js
+++ b/share/spice/launchbug/launchbug.js
@@ -39,7 +39,7 @@
             name: "Launchpad",
             data: api_result,
             meta: {
-                sourceName: "Launchpad.com",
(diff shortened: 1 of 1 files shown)


mattr555:
@puskin94 Thanks! I kind of expected the bold title of the bug to be a link; could we make it one using [`title_content`](https://duck.co/duckduckhack/spice_templates_overview#textdetail-template) and another handlebars template? Unless @jagtalon has a better way to do this? I couldn't find a url property in the docs or in use currently.

puskin94:
@mattr555 No problem!! Waiting for a community feedback! (?)

jagtalon:
:+1: 

@mattr555 oh I see what you mean. It should probably change internally.

PR #1424, opened 2015-01-16
Added a LICENSE file

th4t:
I took the liberty of copying and adjusting the LICENSE file from the [ios](https://github.com/duckduckgo/ios/blob/master/LICENSE) repo.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,13 @@
+Copyright 2015 DuckDuckGo
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


moollaza:
@th4t thanks a lot! I'm not entirely sure that's the correct way to add the Apache License though (meaning we may have made a mistake with those repos!)

It seems that you need to have a LICENSE containing the actual text of the license and they even suggest you prepend a notice to each source file (the notice matches the format of the text you've added as the license)

I'll have a talk with the team about doing the 2nd part but the first part should be easy to do.

Details: http://choosealicense.com/licenses/apache-2.0/

th4t:
No problem at all. Very interesting, thanks for the explanation! The 2nd part seems cumbersome indeed. I changed the file, in case it will turn out to be a good first approximation. [Looking at other repos](https://github.com/duckduckgo/opera-zeroclickinfo/blob/master/LICENSE)  they seem to go into the same direction, but need a bit more love.

jagtalon:
@th4t you might be interested in https://github.com/duckduckgo/zeroclickinfo-goodies/issues/931 as well. :)

th4t:
@jagtalon thanks for the hint! A few minutes well spent :)

PR #1429, opened 2015-01-19
SeatGeek - looking up gigs based on location

andrey-p:
This is building on an IA that already exists, which allows you to retrieve upcoming concerts for an artist using the SongKick API. I've just added some extra features to it.

It allows you to retrieve upcoming concerts based on a location search (city, venue or geolocation).

Makes what's usually a rather tedious search a lot easier.

[SeatGeek](https://seatgeek.com)

There was an already functional AI with this data source.

There's [Last.fm](http://www.last.fm), but its API doesn't seem as powerful. [SongKick](http://www.songkick.com/) is another one, and contains as much information, but can't be used alongside the SeatGeek API due to its terms of use.

"upcoming concerts in London", "live shows at cafe oto", "gigs in my area"

Music lovers. I've already used it a couple of times, and learned about new gigs at my favourite small venue.

[This](https://duck.co/ideas/idea/351/show-concerts-music-shows-near-me), [this](https://duck.co/ideas/idea/11/songkick-ia) and [this](https://duck.co/ideas/idea/623/songkick).

It changes the existing SeatGeek IA. There's a conflict with an existing SongKick IA (which isn't live as far as I can tell).

I wanted to use the template that the IA already had, but I couldn't figure out a good way of sharing a single template between multiple endpoints. I ended up going with one of the default Spice templates to avoid having copies of the same handlebars template in several places, but I'm not sure the default template is ideal.

No API key needed, though they require you not to display ticket listings on behalf of other vendors. Terms [here](https://seatgeek.com/api-terms).

screenshot

That disappearing date subtitle is the reason why I think this template isn't ideal.

(Not massively sure about that missing search button icon, I'm assuming DuckPAN disagreeing with Arch Linux.)

Haven't run browser tests at this point but the current front-end code is minimal - I'd also like to look into a possibility of a better template before diving into that.


moollaza:
@andrey-p thanks a lot for this! We'll be sure to review it shortly.

>I wanted to use the template that the IA already had, but I couldn't figure out a good way of sharing a single template between multiple endpoints. I ended up going with one of the default Spice templates to avoid having copies of the same handlebars template in several places, but I'm not sure the default template is ideal.

Unfortunately you can't share templates between IA's yet, but sticking with a built-in template is actually preferred so that's alright :+1:

jagtalon:
@andrey-p Yeah, I think we should go with the old look here instead of the built-in one. Copying it in different folders should be okay although not ideal.

andrey-p:
Thanks @moollaza and @jagtalon! 

For what it's worth, I did manage to kinda get some template sharing to work - I've kept that branch here: https://github.com/andrey-p/zeroclickinfo-spice/tree/seatgeek-additions/lib/DDG/Spice/SeatGeek . `Events.pm` had `call_type => 'self'` and acted as an entry point. Its front-end contained all the shared elements and did a `$.getScript` on the appropriate endpoint for artists, venues or cities.

I eventually abandoned that line of thought as it wasn't easy to run automated tests against it and I kept getting an error from DuckPAN which led me to believe I was doing something really non-standard. Plus, I couldn't use the `handle` function so I wasn't able to access the geolocation API.

MariagraziaAlastra:
@andrey-p Thanks for opening the PR and sorry for the confusion about sharing the same layout!

So, in theory it should be possible to share the same share dir files across submodules, but it's quite an edge case, so for now let's stick with the standards and go with the ugly duplicated code solution:
since we want to display the same info for concerts, could you please replace the default templates with the custom one currently used for SeatGeek? It's mostly a copy-paste work, I fear :)

andrey-p:
Done. Left a little note at the root of that folder reminding people to add changes across all four instances of the duplicated code. That'll hopefully avoid some confusion when someone wants to update this IA a year from now.

share/spice/seat_geek/events_by_city/seat_geek_events_by_city.js
```diff
@@ -1,17 +1,14 @@
...
-
-        var months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'June', 'July', 'Aug', 'Sept', 'Oct', 'Nov', 'Dec'];
-
-        var days = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
+        var query = DDG.get_query(),
+            months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'June', 'July', 'Aug', 'Sept', 'Oct', 'Nov', 'Dec'],
+            days = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
 
         Spice.add({
             id: "seat_geek",
```
MariagraziaAlastra: @andrey-p Please change the ids, each one should be unique, so for example this could be "seat_geek_events_by_city", etc.

MariagraziaAlastra:
@andrey-p Good idea, writing a note! Thanks :)

MariagraziaAlastra:
@andrey-p Uh, just seen the update! Unfortunately GH only sends a notification mail whenever there's a new comment.

I'll test this and get back to you tomorrow with the results! In the meantime it's not necessary to include my name in the attribution for the submodules containing the new API requests: those are your idea, you created them, so you should get all the glory! :crown:

andrey-p:
@MariagraziaAlastra I *did* still use your templates though - removing attribution for that would seem like a pretty uncool thing to do.

share/spice/seat_geek/events_by_city/seat_geek_events_by_city.css
```diff
@@ -0,0 +1,77 @@
+.zci--seat_geek .tile--seat_geek {
```
MariagraziaAlastra: @andrey-p All these references must be changed as well, to reflect the new ids, or the rules won't be applied:
screen shot 2015-01-22 at 20 41 29

andrey-p: Oops, didn't realize the scoping class was based off that ID. Makes sense in retrospect. Fixed in latest commit.

lib/DDG/Spice/SeatGeek/EventsNearMe.pm
```diff
@@ -0,0 +1,40 @@
...
+    'concerts',
+    'live',
+    'live shows',
+    'shows',
+    'gigs';
+
+spice proxy_cache_valid => "418 1d";
+spice is_cached => 0;
+
+spice to => 'http://api.seatgeek.com/2/events?taxonomies.name=concert&lat=$1&lon=$2&range=10mi&&callback={{callback}}';
```
MariagraziaAlastra: I'd prefer this to have either the default range (30mi) or a little wider. In some places there's a quite a lot of distance between cities and people might be interested in concerts in those near cities or regions as well, especially if they live in some little city which usually doesn't have lots of music events :)

andrey-p: @MariagraziaAlastra Good catch, I didn't think of that - I guess I'm pretty lucky to live in a city with loads of venues :) Done.

jagtalon:
@MariagraziaAlastra it's up on https://ddh1.duckduckgo.com/?q=live+shows+in+london&ia=concerts if you'd like to check it out

MariagraziaAlastra:
Thanks @jagtalon!

MariagraziaAlastra:
Hey @andrey-p, apparently specifying "concerts" as taxonomy doesn't always get music events back. I tried this query: https://ddh1.duckduckgo.com/?q=live+shows+in+italy&ia=concerts 
(Which shouldn't even trigger, because "Italy" isn't a city, but it's listed as if it was in the API response: http://api.seatgeek.com/2/events?taxonomies.name=concert&venue.city=italy) and it shows a sports event right now: 
screen shot 2015-01-23 at 15 55 35

Turns out that the API allows returning both concerts and sports event with that taxonomy, please see [the documentation](http://platform.seatgeek.com/#events) (scroll down to "taxonomies Argument and look on the right column)

So, after a bit of investigation, it seems like the best thing to do to distinguish between sports and concerts is to check for the existence of the "genres" key for each performer, inside the JSON response, here's an example: http://api.seatgeek.com/2/events?taxonomies.name=concert&performers.slug=katy-perry

Please let me know if you have any issues!

lib/DDG/Spice/SeatGeek/EventsByArtist.pm
```diff
@@ -0,0 +1,43 @@
...
+
+spice proxy_cache_valid => "200 304 12h";
+
+spice to => 'http://api.seatgeek.com/2/events?taxonomies.name=concert&performers.slug=$1&callback={{callback}}';
+
+handle remainder_lc => sub {  
+    # If query starts with any of these assume it's one of the other queries
+    return if ($_ =~ /^(in |at |near me)/);
+
+    # Removes spaces from the start and beginning of the query
```
MariagraziaAlastra: Minor wording change: "from the beginning and end"

andrey-p:
@MariagraziaAlastra Wow, this is a pretty weird one. It kinda seems like a data entry problem on their side. The query does list those events as being in "Italy, UK" and the `taxonomies.name` option has worked pretty well otherwise. I've made a post on their [dev forum](https://zohodiscussions.com/sgplatform#Topic/70211000000261003) to let them know.

It might still be a good idea to put some genre checks in place, but I'm wondering whether that might prevent some lesser-known artists from being displayed, in case SG doesn't have genre info on them.

I'll run a few quick tests against their API to make sure.

MariagraziaAlastra:
@andrey-p Yeah I know, totally unexpected! Thanks for taking the time to write that post on their dev forum, and please let me know if I can help with tests or anything :)

andrey-p:
Hmm... Some digging around some tiny venues came up with the following performer:

`curl 'http://api.seatgeek.com/2/events?taxonomies.name=concert&performers.slug=campfires'`

Excerpt from the result:

```json
         "performers": [
                {
                    "has_upcoming_events": true,
                    "home_venue_id": null,
                    "id": 122267,
                    "image": null,
                    "images": {},
                    "name": "Campfires",
                    "primary": true,
                    "score": 0,
                    "short_name": "Campfires",
                    "slug": "campfires",
                    "stats": {
                        "event_count": 1
                    },
                    "taxonomies": [
                        {
                            "id": 2000000,
                            "name": "concert",
                            "parent_id": null
                        }
                    ],
                    "type": "band",
                    "url": "https://seatgeek.com/campfires-tickets"
                }
            ],
```

I was hoping the `genres` array will be there but empty, which would've still allowed us to differentiate between music and non-music performers, but that doesn't quite work.

I reckon if this is a one-off mistake on SG's part, it might be better to not put any genre checks in and avoid skipping smaller gigs. What do you think?

MariagraziaAlastra:
@andrey-p That's too bad. Anyway, skipping smaller gigs isn't ideal, but it's still better than having sports events trigger when a user is searching for concerts, so please go ahead implementing this genres check.

In the meantime we'll send feedback directly to SeatGeek and see if they can improve their API, so we'll be able to get better results.

How does that sound?

andrey-p:
@MariagraziaAlastra If you think that's the best approach, I'll go with that :) I'll probably get back to you on Monday with a fix.

andrey-p:
@MariagraziaAlastra If you think that's the best approach, I'll go with that :) I'll get back to you on Monday with a fix.

MariagraziaAlastra:
@andrey-p Great, thanks!

share/spice/seat_geek/events_by_artist/seat_geek_events_by_artist.js
```diff
@@ -0,0 +1,168 @@
...
+        var months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'June', 'July', 'Aug', 'Sept', 'Oct', 'Nov', 'Dec'],
+            days = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
+            events;
+            
+        // sometimes we get false positives from the SeatGeek API
+        // where for example sports games get displayed instead of concerts
+        //
+        // filtering out events whose performers don't have genres
+        // seems to be a way to guard against that
+        events = $.grep(api_result.events, function (item) {
```
MariagraziaAlastra: @andrey-p Could you please move the check inside the normalize function? That goes already through all the items one at a time, so you can just check whether at least one of the performers for that event has the "genre" key, and if that's not the case, return `null`, so that item will be skipped.

andrey-p: @MariagraziaAlastra Thanks, that's a good thing to know - makes everything cleaner too.

One thing I noticed though - if all the events are filtered out in the `normalize` function (like in that `gigs in italy` case), `Spice.failed()` doesn't get called. Is that a problem? Or something that needs to be reported as an issue with the DDG platform?

MariagraziaAlastra: @andrey-p It shouldn't be much of a problem, we usually perform checks like this inside the normalize function (see an example [here](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/book/book.js#L34-L46), or we use the relevancy block (example [here](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/quixey/quixey.js#L136), see also the [documentation](https://duck.co/duckduckhack/spice_displaying#relevancy)), without calling `Spice.failed()`, because that's more about errors in the request, or not getting any result at all. 
The relevancy is a different problem, but yeah, it would be probably a good idea to try to handle more gracefully the lack of relevant results.

andrey-p: @MariagraziaAlastra That makes sense. I imagine using the relevancy block would've been a bit better for this specific purpose but from what I can tell it doesn't allow more complex logic.

MariagraziaAlastra: @andrey-p Yeah exactly, that's why in this case it's better to just check inside the normalize function.

MariagraziaAlastra:
@andrey-p Nice work!
Updated ddh1 with the latest changes: https://ddh1.duckduckgo.com/?q=concerts+in+london&ia=concerts

Some screenshots:
screen shot 2015-02-03 at 15 59 16
screen shot 2015-02-03 at 15 59 30


@jagtalon Would you like to test this and sign off before merging?

moollaza:
@andrey-p @MariagraziaAlastra 

A few things:

- The "More at" link points to a search for the full query -- can we parse out just the venue/artist/location?
- Any way we can get more results? Most of my searches return less than a page of results
- Is the outline around the date supposed to be circular? It's a little bit oval-shaped for me
- How come there's no answer for https://ddh1.duckduckgo.com/?q=hozier+concert -- the API has lots of results

andrey-p:
> - The "More at" link points to a search for the full query -- can we parse out just the venue/artist/location?

It works like that when searhing for specific artists (ie "concerts bjork" will search for just "bjork" on the More At link). The SeatGeek search engine seemed to handle city/venue queries like "concerts in london" or "concerts at cafe oto" reasonably fine so I thought it might be useful to disambiguate.

> - How come there's no answer for https://ddh1.duckduckgo.com/?q=hozier+concert -- the API has lots of results

Hmm, that would be the result filtering. It had to be added to protect against incorrect results in the SeatGeek API. [See above for discussion.](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1429#issuecomment-72102213)

> - Any way we can get more results? Most of my searches return less than a page of results
- Is the outline around the date supposed to be circular? It's a little bit oval-shaped for me
- "upcoming artist_name concerts/gigs/etc" would be a good trigger, it was one of the first queries I thought of

These should be fairly straightforward to add/fix. I'll do that later today.

jagtalon:
@andrey-p sweet, just ping us if you're ready to go! :)

andrey-p:
@moollaza Just fixed everything from the list except for the `hozier concert` issue - do you think anything needs to be changed about the result filtering?

@jagtalon I guess just waiting to see if the result filtering thing is ok, otherwise it seems good to go!

jagtalon:
@andrey-p @moollaza yeah I think we shouldn't touch the filtering for now because of the reasons stated in https://github.com/duckduckgo/zeroclickinfo-spice/pull/1429#issuecomment-72227064. 

Looks good, and looks like @moollaza's suggestions have been made. :+1: 

screen shot 2015-02-06 at 3 39 49 am
screen shot 2015-02-06 at 3 41 30 am
screen shot 2015-02-06 at 3 43 18 am

andrey-p:
@jagtalon That's excellent, thank you very much. Working on this was really good fun!

MariagraziaAlastra:
@andrey-p It's live! https://duckduckgo.com/?q=concerts+in+leeds&ia=concerts

:fireworks:

andrey-p:
@MariagraziaAlastra Sweet! :)

PR #1430, opened 2015-01-19
Update LICENSE, restore boilerplate text

moollaza:
Copied the license from http://www.apache.org/licenses/LICENSE-2.0.txt -- not sure why the brackets are different there, but its the most official source, so I'm going with it.


moollaza:
//cc @jagtalon @th4t

jagtalon:
:+1:

PR #1445, opened 2015-01-24
XKCD: Simple, static prev/next/explain links

moollaza:
//cc @chrismorast @jagtalon @gamehelp16

Updated to match @chrismorast's design:

2015-01-23_21h39_47
2015-01-23_21h40_06

### DIFF
--- a/share/spice/xkcd/display/content.handlebars
+++ b/share/spice/xkcd/display/content.handlebars
@@ -6,13 +6,13 @@
       {{#xkcd_previousNum num}}
       <a href="/?q=xkcd+{{num}}">Prev</a>
       {{/xkcd_previousNum}}
-      <span>|</span>
+      <span class="sep"></span>
 
       {{#if has_next}}
         {{#xkcd_nextNum num}}
           <a href="/?q=xkcd+{{num}}">Next</a>
         {{/xkcd_nextNum}}
-        <span>|</span>
+        <span class="sep"></span>
       {{/if}}
       <a href="http://www.explainxkcd.com/wiki/index.php/{{num}}">Explain</a>
     </div>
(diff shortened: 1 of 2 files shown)


moollaza:
@chrismorast I just noticed we're using the `zci__caption` class for the header -- should that be `zci__header`?

moollaza:
Here's it with `zci__header`:

2015-01-23_21h48_50

MrChrisW:
@moollaza I'm just nitpicking! This is an edge case (and therefore probably not worth fixing). On screens with a width smaller then 270px the buttons are a little to the right. This is probably tolerated.. however we could throw a media query in there to fix it.

This is due to the following selectors:
*padding* in `zci__subheader.xkcd--date`
*margin-left* in `.zci--xkcd .xkcd--buttons`

selection_120

chrismorast:
Nice catch @MrChrisW ! @moollaza , Can you add a little more margin to the top and bottom of the image to give it a little more breathing room with the surrounding elements? Maybe an addition .25em or .5em. Other than that, this looks good to me!

chrismorast:
@moollaza - in addition, can we turn off the visited link color here?

jagtalon:
@moollaza Nice--I kinda like the old one, but this looks good too

jagtalon:
@moollaza ping! :D

moollaza:
Alright @MrChrisW @chrismorast implemented your suggestions.

I've increased the margins around the image (`1em` above, `0.5em` below), and removed the weird padding I had for the subtitle, so now it's a little close to the title (as per our own CSS)

2015-02-19_22h48_11

**Mobile**:
2015-02-19_22h51_24

**Mobile, < 250px** (not sure anything is this narrow?):
2015-02-19_22h52_13

moollaza:
...and removed the visited link colour

jagtalon:
@moollaza looking good! :+1:

PR #1446, opened 2015-01-24
Timer: add new trigger (time)

brownl:
It makes sense for a search like `time 3 minutes` to bring up the countdown timer.

### DIFF
--- a/lib/DDG/Spice/Timer.pm
+++ b/lib/DDG/Spice/Timer.pm
@@ -12,6 +12,7 @@ attribution twitter => 'mattr555',
             github => ['https://github.com/mattr555/', 'Matt Ramina'];
 
 triggers startend => ['timer', 'countdown'];
+triggers start => ['time'];
 
 spice call_type => 'self';
 


jagtalon:
@moollaza do you know why this would prevent Time.pm from triggering?

@brownl add to `t/Timer.t`?

moollaza:
>do you know why this would prevent Time.pm from triggering?

@jagtalon not entirely sure what you're talking about? I tried that query and it triggers as expected with the PR. What happened when you tried?

jagtalon:
@moollaza Oh I mean why did "time in india" (which should trigger Time.pm) stop working? Or is it only me?

brownl:
It seems like we can't have the same trigger for multiple instant answers. Is there a solution to this?

moollaza:
@brownl @jagtalon this may have been related to a bug that was recently fixed in the duckduckgo repo.

I haven't released the fix, but after installing the latest version from master and then testing your query, `time in india`, everything works as expected. Time.pm returns a result:

2015-01-31_12h02_12

FYI I also tested this with the duckpan server and the correct Time IA produced a result.

Once I've released the duckduckgo repo, you'll be able to see the same results.

moollaza:
Proof the Timer Spice also works as expected:

2015-01-31_12h06_07

brownl:
Thanks

jagtalon:
@moollaza nice thanks!

jagtalon:
:+1:

Issue #1436, opened 2015-01-21
Flights: Needs some CSS love


jagtalon:
This has regressed from it's original state. Mainly the title size and padding but also need to make sure the larger text (title and times) should have the class `tx-clr--dk` (so that it has the color `#333`). 

Should be:
![vx 143 at duckduckgo](https://cloud.githubusercontent.com/assets/81969/5839802/b9a7347a-a15e-11e4-9638-fea0ed8351f3.jpg)

Instead we have:
![screen shot 2015-01-21 at 11 14 38 am](https://cloud.githubusercontent.com/assets/81969/5839805/bfb4337c-a15e-11e4-9df9-1e05d84e2980.png)

CC @abeyang @chrismorast

jagtalon:
Actually, if it had Route's CSS it would be fine:

![screen shot 2015-01-23 at 8 30 02 am](https://cloud.githubusercontent.com/assets/81969/5875352/1de62fd2-a2da-11e4-962b-3307c9927cd2.png)

https://duckduckgo.com/?q=Jetblue+Boston+to+Los+Angeles&ia=route

MrChrisW:
Fixed by @tommytommytommy in #1449

PR #1449, opened 2015-01-25
Update airlines IA display to mirror flights_route

tommytommytommy:
Addresses #1436 

screen shot 2015-01-25 at 1 34 36 am

Currently, the airlines IA (which searches by airline code and flight numbers) and the flights_route IA (which searches by airline and source/destination cities) do not share any code. This PR unifies the appearance of the two IAs. Eventually, we should try to unify the two IAs to avoid having to maintain two different sets of code. (This is on my list of weekend projects to attempt :)

### DIFF
--- a/share/spice/airlines/airlines.js
+++ b/share/spice/airlines/airlines.js
@@ -23,7 +23,7 @@
         if (!api_result || !api_result.flight) {
             return Spice.failed('airlines');
         }
- 
+        
         // Check if flight is an array or not.
         var flight = [];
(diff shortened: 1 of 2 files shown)


moollaza:
@tommytommytommy nice, looking good to me. 
Would like @abeyang to sign off

abeyang:
Yes, this is good to go! :+1: 

FYI, Thom is working on updating the tiles, but I think that's independent of this.

PR #1452, opened 2015-01-27
Hearthstone Spice debut

Akryum:
# Spice Pull Request

It shows infos about any Hearthstone card with a thumbnail.

It allows to get instantly main informations like what a card does, without searching in a database or a wiki.

Gamepedia Hearthstone Wiki
http://hearthstone.gamepedia.com/

It seems to be the most accurate, and has image for every card.

Not better alternatives:
http://hearthstonejson.com/
http://hearthstoneapi.herokuapp.com/

hearthstone leeroy
hearthstone tidecaller
hearthstone murloc warleader

Gamers!

No.

None I can think of.

No thanks. :)

The API is a Mediawiki API. A Warranty disclaimer applies.

On the duckduckgo timeline (https://duckduckgo.com/about). :)

duckduckgo-hearthstone3

## Checklist

```

    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [x] Google Chrome
        [x] Firefox
        [x] Opera
        [] IE 8
        [] IE 9
        [] IE 10
(Tested on IE11)

    - Windows XP
        [] IE 7
        [] IE 8

    - Mac OSX
        [] Google Chrome
        [] Firefox
        [] Opera
        [] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [x] Firefox
        [] Native Browser
        [x] Google Chrome
        [] Opera
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 ---
--sudo: false
+sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 ---
+-sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 ---
-sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 ---
+sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
--- a/lib/DDG/Spice/Hearthstone.pm
+++ b/lib/DDG/Spice/Hearthstone.pm
@@ -7,7 +7,7 @@ use DDG::Spice;
 spice is_cached => 1;
 spice proxy_cache_valid => "200 4d";
 
-#Metadata
+# Metadata
 name "Hearthstone Card Search";
 source "Bytevortex (Gamepedia Proxy)";
 icon_url "http://eu.battle.net/hearthstone/static/images/icons/favicon.ico";
--- a/lib/DDG/Spice/Hearthstone.pm
+++ b/lib/DDG/Spice/Hearthstone.pm
@@ -9,7 +9,7 @@ name "Hearthstone Card Search";
 source "Bytevortex (Gamepedia Proxy)";
 icon_url "http://eu.battle.net/hearthstone/static/images/icons/favicon.ico";
 description "Get a preview of any Hearthstone card.";
-primary_example_queries "hearthstone leeroy jenkins", "hearthstone leeroy";
+primary_example_queries "hearthstone leeroy jenkins", "hearthstone leeroy", "leeroy hearthstone", "hearthstone mirror";
 category "entertainment";
 topics "gaming";
 code_url "https://github.com/Akryum/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Hearthstone.pm";
@@ -17,7 +17,7 @@ attribution github => ["https://github.com/Akryum", "Akryum"],
             twitter => "Akryum";
 
 # Triggers
-triggers any => "hearthstone";
+triggers startend => "hearthstone";
 
 spice to => 'http://bytevortex.net/hearthstone.php?search=$1';
 
@@ -27,6 +27,11 @@ spice wrap_jsonp_callback => 1;
 handle remainder => sub {
 
     return unless $_;    # Guard against "no answer"
+    
+    # Keyword Blacklist
+    if(/(game|instruction|stoves|warcraft|deck|forum|wiki)/) {
+        return;
+    }
     
     return $_;
 };
--- a/lib/DDG/Spice/Hearthstone.pm
+++ b/lib/DDG/Spice/Hearthstone.pm
@@ -16,6 +16,7 @@ code_url "https://github.com/Akryum/zeroclickinfo-spice/blob/master/lib/DDG/Spic
 attribution github => ["https://github.com/Akryum", "Akryum"],
             twitter => "Akryum";
 
+
 # Triggers
 triggers startend => "hearthstone";
 
@@ -23,17 +24,22 @@ spice to => 'http://bytevortex.net/hearthstone.php?search=$1';
 
 spice wrap_jsonp_callback => 1;
 
+# Keyword blacklist
+my $keyword_guard = qr/game|instruction|stoves|warcraft|deck|forum|wiki|reddit/;
+
 # Handle statement
 handle remainder => sub {
-
-    return unless $_;    # Guard against "no answer"
+    
+    # Guard against "no answer"
+    return unless $_;
     
     # Keyword Blacklist
-    if(/(game|instruction|stoves|warcraft|deck|forum|wiki)/) {
-        return;
-    }
+    return if (/$keyword_guard/);
     
-    return $_;
+    # Regex guard
+    return $_ if (/^([a-z0-9':!\/,.-]+\s*)+$/i);
+    
+    return;
 };
 
 1;
--- a/lib/DDG/Spice/Hearthstone.pm
+++ b/lib/DDG/Spice/Hearthstone.pm
@@ -3,8 +3,11 @@ package DDG::Spice::Hearthstone;
 
 use DDG::Spice;
 
+# Caching
 spice is_cached => 1;
+spice proxy_cache_valid => "200 4d";
 
+#Metadata
 name "Hearthstone Card Search";
 source "Bytevortex (Gamepedia Proxy)";
 icon_url "http://eu.battle.net/hearthstone/static/images/icons/favicon.ico";
@@ -16,12 +19,13 @@ code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/
 attribution github => ["https://github.com/Akryum", "Akryum"],
             twitter => "Akryum";
 
-
 # Triggers
 triggers startend => "hearthstone";
 
+# Target URL
 spice to => 'http://bytevortex.net/hearthstone.php?search=$1';
 
+# JS Callback
 spice wrap_jsonp_callback => 1;
 
 # Keyword blacklist
--- a/lib/DDG/Spice/Hearthstone.pm
+++ b/lib/DDG/Spice/Hearthstone.pm
@@ -1,0 +1,34 @@
+package DDG::Spice::Hearthstone;
+# ABSTRACT: Retrieve the card infos from Hearthstone game.
+
+use DDG::Spice;
+
+spice is_cached => 1;
+
+name "Hearthstone Card Search";
+source "Bytevortex (Gamepedia Proxy)";
+icon_url "http://eu.battle.net/hearthstone/static/images/icons/favicon.ico";
+description "Get a preview of any Hearthstone card.";
+primary_example_queries "hearthstone leeroy jenkins", "hearthstone leeroy";
+category "entertainment";
+topics "gaming";
+code_url "https://github.com/Akryum/zeroclickinfo-spice/lib/DDG/Spice/Hearthstone.pm";
+attribution github => ["https://github.com/Akryum", "Akryum"],
+            twitter => "Akryum";
+
+# Triggers
+triggers any => "hearthstone";
+
+spice to => 'http://bytevortex.net/hearthstone.php?search=$1';
+
+spice wrap_jsonp_callback => 1;
+
+# Handle statement
+handle remainder => sub {
+
+    return unless $_;    # Guard against "no answer"
+    
+    return $_;
+};
+
+1;
--- a/lib/DDG/Spice/Hearthstone.pm
+++ b/lib/DDG/Spice/Hearthstone.pm
@@ -12,7 +12,7 @@ description "Get a preview of any Hearthstone card.";
 primary_example_queries "hearthstone leeroy jenkins", "hearthstone leeroy", "leeroy hearthstone", "hearthstone mirror";
 category "entertainment";
 topics "gaming";
-code_url "https://github.com/Akryum/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Hearthstone.pm";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Hearthstone.pm";
 attribution github => ["https://github.com/Akryum", "Akryum"],
             twitter => "Akryum";
 
@@ -37,6 +37,8 @@ handle remainder => sub {
     return if (/$keyword_guard/);
     
     # Regex guard
+    # Allows string similar to any Hearthstone card name: multiple words separated by spaces, commas or dashes.
+    # The card name may also have dots, double dots, slashes or an exclamation mark (Those Blizzard devs).
     return $_ if (/^([a-z0-9':!\/,.-]+\s*)+$/i);
     
     return;
--- a/lib/DDG/Spice/Hearthstone.pm
+++ b/lib/DDG/Spice/Hearthstone.pm
@@ -12,7 +12,7 @@ description "Get a preview of any Hearthstone card.";
 primary_example_queries "hearthstone leeroy jenkins", "hearthstone leeroy";
 category "entertainment";
 topics "gaming";
-code_url "https://github.com/Akryum/zeroclickinfo-spice/lib/DDG/Spice/Hearthstone.pm";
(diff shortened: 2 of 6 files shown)


share/spice/hearthstone/hearthstone.handlebars
```diff
@@ -0,0 +1,31 @@
...
+        <p>
+            <span>Cost: <b>{{cost}}</b> <image class="zci--hearthstone--icon" src="http://hydra-media.cursecdn.com/hearthstone.gamepedia.com/8/8b/Mana_icon.png"/></span>
+            
+            {{#if attack}}
+            <span> - Attack: <b>{{attack}}</b> <image class="zci--hearthstone--icon" src="http://hydra-media.cursecdn.com/hearthstone.gamepedia.com/f/fe/Attack_icon.png"/></span>
+            {{/if}}
+            
+            {{#if health}}
+            <span> - Health: <b>{{health}}</b> <image class="zci--hearthstone--icon" src="http://hydra-media.cursecdn.com/hearthstone.gamepedia.com/1/17/Health_icon.png"/></span>
+            {{/if}}
```
MrChrisW: You'll need to use the `imageProxy` helper for all images, see https://duck.co/duckduckhack/spice_handlebars_helpers#imageproxy

share/spice/hearthstone/hearthstone.js
```diff
@@ -0,0 +1,49 @@
...
+        // Card Hero
+        if(card.hero === "Any") {
+            card.hero = "Any Hero";
+        }
+        
+        
+        // Render the response
+        Spice.add({
+            id: "hearthstone",
+            name: "Card",
```
MrChrisW: Thinking this should be update this to "Hearthstone", "Hearthstone Card" or the standard "Answer".

.travis.yml
```diff
@@ -1,5 +1,4 @@
 ---
-sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml
```
MrChrisW: This file was changed recently in the master branch. Looks like your master branch hasn't been updated with those changes. Probably the easiest option is to manually add the `sudo: false` line to your local file and commit. Another option is to perform a rebase.

share/spice/hearthstone/hearthstone.js
```diff
@@ -0,0 +1,49 @@
...
+            name: api_result.Has_name,
+            type: api_result.Has_card_type,
+            rarity: api_result.Has_rarity,
+            hero: api_result.Has_class,
+            cost: api_result.Has_mana_cost,
+            attack: api_result.Has_attack,
+            health: api_result.Has_health,
+            description: api_result.Has_description,
+            flavor: api_result.Has_flavor_text
+        };
```
MrChrisW: Its it possible the API won't return one of any of the values above?

MrChrisW:
@Akryum This is a great first pass! Thanks for contributing :cake:  I've added some feedback and I'm sure other community leaders will check this out shortly. Feel free to ask any questions. Do you own the API endpoint? bytevortex.net

Akryum:
@MrChrisW Thanks for the feedback! And yes I own bytevortex.net ;)

MrChrisW:
@Akryum No problem! Happy to help. Great!

Akryum:
Feedback applied! :)

MrChrisW:
Thanks @Akryum, We'll just need to await a staff review or other feedback from the community!

Akryum:
Ironed out a few quirks.

MrChrisW:
@Akryum :+1:

Akryum:
Got a question: How are IA supposed to be translated/regionalized?

MrChrisW:
@Akryum Good question! 

Right now all Instant Answers provide English results. Translation will come later, so initially we're moving forward with just English answers.

moollaza:
@Akryum @MrChrisW 

This is looking pretty good so far, I have 2 concerns:

1. We're using the base template and a lot of custom HTML. Let's use a pre-defined template group (I think the `info` group should work here) and reduce the amount of custom code for this. I think  the look/feel of the new [MTG Spice](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/mtg/mtg.js) https://duckduckgo.com/?q=mtg+nullify&ia=magicthegathering would be ideal. It also makes use of the InfoBox so we have room to show lots of data with minimal effort and templating.

2. We're not doing any relevancy checking against the query. Right now this will trigger whenever the query has "hearthstone" **anywhere** in the query. I'm sure a large portion of those potential queries aren't a search for a hearthstone card (e.g "hearthstone game", "hearthstone instructions", "hearthstone stoves")

    Let's improve our triggers and add some checking so we're confident the query that triggered this is likely to be a search for a hearthstone card and then we don't have to worry about showing lots of potentially irrelevant results.

share/spice/hearthstone/hearthstone.js
```diff
@@ -0,0 +1,53 @@
...
+
+        // Validate the response
+        if (!api_result || api_result.error || !api_result.Has_name || !api_result.page) {
+            return Spice.failed('hearthstone');
+        }
+        
+        // Card Data
+        // May (but very unlikely) return <undefined> values for : image_url, Has_card_type, Has_rarity,
+        // Has_class, Has_mana_cost
+        var card = {
```
moollaza: This should be handled with a `normalize` function: https://duck.co/duckduckhack/spice_displaying#data-normalization

share/spice/hearthstone/hearthstone.js
```diff
@@ -0,0 +1,53 @@
...
+            page: api_result.page,
+            image: api_result.image_url,
+            name: api_result.Has_name,
+            type: api_result.Has_card_type,
+            rarity: api_result.Has_rarity,
+            hero: api_result.Has_class,
+            cost: api_result.Has_mana_cost,
+            attack: api_result.Has_attack, // <undefined> for cards other than minions
```
moollaza: If there's a chance of these being undefined, I would first check to see if it's defined, and if so, in your normalize function, create the property and set the value accordingly

Akryum:
@moollaza I don't really think of a good way of restricting the trigger... A keyword blacklist maybe?

Akryum:
Using info template:
duckduckgo-hearthstone4

Akryum:
I think that duckpan test modify my .travis.yml everytime. :/

lib/DDG/Spice/Hearthstone.pm
```diff
@@ -0,0 +1,39 @@
...
+
+# Handle statement
+handle remainder => sub {
+
+    return unless $_;    # Guard against "no answer"
+    
+    # Keyword Blacklist
+    if(/(game|instruction|stoves|warcraft|deck|forum|wiki)/) {
+        return;
+    }
```
MrChrisW: Can you place the guard keywords outside of the `handle` function like:

`my $keyword_guard = qr/game|instruction|stoves|warcraft|deck|forum|wiki/;`

Then inside:
`return if (/$keyword_guard/);`

Just for maintainability, easier to add new keywords.

MrChrisW:
@moollaza Thanks for the comments. :+1: 
I did have a think about relevancy and my thought process was: 

- Check the query matches a list of cards (slurped in perl) - *unmaintainable (too many cards + adding new cards)*

- Modify the trigger to include explicitly the word "card" - *not user friendly*

A regex guard looks like the best way to go.

Just a note, the [mtg spice](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Mtg.pm) doesn't include a regex guard and isn't currently live.

Akryum:
Sometimes, the description of the card is cut too short:
duckduckgo-hearthstone5
It may have something to do with the info templates and handlebars.

Akryum:
Resizing the window bring the full description back.
Also, the flavor text appears randomly for some cards and not for others.

moollaza:
> I think that duckpan test modifies my .travis.yml all the time. :/

@Akryum Which DuckPAN test? You shouldn't need to commit a change to the .travis.yml

Akryum:
Running the duckpan test command modifies the .travis.yml file automatically and deletes sudo:false each time.

jagtalon:
@moollaza @Akryum yeah this happens because .travis.yml is autogenerated. Will make a PR for it!

jagtalon:
Can be replicated with `dzil build` as well

jagtalon:
@moollaza @Akryum https://github.com/duckduckgo/zeroclickinfo-spice/pull/1456

jagtalon:
@Akryum we can't have bolded text in the IA (usually) since it's primary use is in pointing out the query in the search results.

Akryum:
So do we keep it thus way?

jagtalon:
@abeyang It's up on https://ddh1.duckduckgo.com/?q=hearthstone+leeroy+jenkins&ia=hearthstone

jagtalon:
@Akryum maybe changing the color is better. @abeyang what do you think?

lib/DDG/Spice/Hearthstone.pm
```diff
@@ -0,0 +1,45 @@
...
+handle remainder => sub {
+    
+    # Guard against "no answer"
+    return unless $_;
+    
+    # Keyword Blacklist
+    return if (/$keyword_guard/);
+    
+    # Regex guard
+    return $_ if (/^([a-z0-9':!\/,.-]+\s*)+$/i);
```
jagtalon: What is it guarding against? It's a bit hard to tell from the regexp--could you add a comment that explains it. :)

lib/DDG/Spice/Hearthstone.pm
```diff
@@ -0,0 +1,45 @@
...
+spice is_cached => 1;
+
+name "Hearthstone Card Search";
+source "Bytevortex (Gamepedia Proxy)";
+icon_url "http://eu.battle.net/hearthstone/static/images/icons/favicon.ico";
+description "Get a preview of any Hearthstone card.";
+primary_example_queries "hearthstone leeroy jenkins", "hearthstone leeroy", "leeroy hearthstone", "hearthstone mirror";
+category "entertainment";
+topics "gaming";
+code_url "https://github.com/Akryum/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Hearthstone.pm";
```
jagtalon: Please change to `https://github.com/duckduckgo/...`

share/spice/hearthstone/hearthstone.js
```diff
@@ -0,0 +1,94 @@
...
+                            value: api_result[key]
+                        });
+                    }
+                });
+                
+                // Normalized data
+                var card = {
+                    title: item.Has_name,
+                    url: "http://hearthstone.gamepedia.com/" + item.page,
+                    image: item.image_url,
```
jagtalon: I wonder if we should maybe check the image_url? Is it always available?

Akryum:
Imho, the color change may be a bit too subtle compared to bold.

Akryum:
I added abbreviation for card keywords like "Battlecry" api-side, so that user can hover them and get a hint at what it does. No moficiation needed on the spice side.

jkanarek:
@abeyang any ideas for a rollover style that looks a little less..."buttony"?

moollaza:
@Akryum let's avoid any type of "rollover" effects. We don't typically use that for Instant Answers because it doesn't work on mobile devices.

Akryum:
This is just a <abbr> HTML tag using the system tooltip, nothing special.

jagtalon:
@Akryum Nice thanks. Yeah it might be too subtle--we'll get back to you on the style. In the mean time, here's a relevancy issue:

screen shot 2015-01-30 at 2 39 27 am

Maybe card should be stripped out in the `.pm` file.

Akryum:
It should be fixed api-side (may need to empty cache).

Akryum:
How is the spice cache emptied? Every day?

moollaza:
@Akryum regarding the rollover, thanks for explaining, a tooltip should be okay, but if its important enough to show, I'd prefer we found a way that works for all devices.

The caching of results can be controlled by your IA based on the `spice proxy_cache_valid` setting. It defaults to:

`spice proxy_cache_valid => 200 1d`, which means we'll cache successful (200) API responses for one day. You can turn that higher or lower if you prefer, but we obviously lean towards more caching so we can return results faster.

Akryum:
Well, for this kind of info we can go with a strong caching imho.

Akryum:
@moollaza Is this correct? `spice proxy_cache_valid => "400 2d";`

Akryum:
The cache was emptied so it's now okay on ddh1 : https://ddh1.duckduckgo.com/?q=hearthstone+leeroy+jenkins+card&ia=hearthstone

abeyang:
Hi @Akryum, it's shaping up nicely. Is it possible to move the cost / attack / health to the card details (including the icons)?

And let's also unbold the description and un-italicize the footnote (and also use a smaller font and lighter color)...

pasted_image_2_2_15__12_46_pm

Akryum:
Hi @abeyang, what's the correct way to add the icons in the infobox? (It escapes html)

Akryum:
duckduckgo-hearthstone7

Akryum:
This is the result for now:
duckduckgo-hearthstone7

jagtalon:
@Akryum I actually like that. Looks super clean CC @abeyang

abeyang:
@Akryum can you throw this onto DDH1? Would like to see this design with other cards (esp. those with a fuller description). But I think we're almost there!

Akryum:
@abeyang Well, you may ask @jagtalon for that I think. :3

jagtalon:
@Akryum @abeyang sorry about that. Updated: https://ddh1.duckduckgo.com/?q=hearthstone+leeroy+jenkins+card&ia=hearthstone

jagtalon:
@abeyang I don't think we have `#999` as a standard color in https://duckduckgo.com/styleguide. What about using the `tx-clr--lt` class which adds `#888`?

screen shot 2015-02-06 at 3 49 01 am

Akryum:
Flavor text nos has `tx-clr--lt` css class.

moollaza:
@Akryum thanks!

@jagtalon can you update the code on DDH1 please? and then @abeyang can take another look and sign-off.

jagtalon:
@abeyang @moollaza here you go! https://ddh9.duckduckgo.com/?q=hearthstone+tidecaller&ia=hearthstone looking good!

moollaza:
Looks great! Would like a design sign-off and then we can merge it in :+1:

abeyang:
Hi @Akryum , after looking at a few examples, I think it's best to just fit the little "gag" into the subtitle. For example, the [Crackle](https://ddh9.duckduckgo.com/?q=hearthstone+crackle&ia=hearthstone) card would look more like this:

> **Crackle**
> Snap! This card! Pop!
> 
> Deal 3-6 damage. Overload: (1)

Hope that makes sense. Let us know if you need help with anything!

moollaza:
@abeyang unfortunately the Info template doesn't allow for a `subtitle` property.

We could add the markup to the description template, but unless we add additional CSS to shift it on the page it's going to look a little odd because the subtitle div will be contained in the `zci__content` div that wraps the description.

I think this has been fixed in the updated templates, so perhaps we can leave it as-is and change it later?

chrismorast:
He means subheader @moollaza 
screen shot 2015-02-20 at 3 51 59 pm

Akryum:
@chrismorast Do I put the flavor text subheader div in the content template or is there any property for that?

chrismorast:
Hi @Akryum , put that content in the subtitle property. We will be deploying an update to the template to allow for this within the next day or two.

moollaza:
@Akryum sorry! I didn't realize you had updated this -- we don't get notified of commits, only comments!

moollaza:
LGTM :+1:

moollaza:
@Akryum I just noticed we're not using the CSS file -- can you make a PR to remove it?

moollaza:
Nevermind I see what's happening. The text comes as HTML from the API. Probably best to use `DDG.stripHTML()` but this is fine.

moollaza:
@Akryum congrats! It's live: https://duckduckgo.com/?q=hearthstone+murloc+warleader&ia=hearthstone

Akryum:
Hi! It seems that the card IA is no longer working. Is there any problem I can help with?

moollaza:
@Akryum do you have any examples? I just tried and it worked:

hearthstone_murloc_warleader_at_duckduckgo_and_duckduckgo_-_hipchat

Akryum:
@moollaza It seems that some cards no longer display the IA like these:
https://duckduckgo.com/?q=hearthstone+blizzard
https://duckduckgo.com/?q=hearthstone+oasis
https://duckduckgo.com/?q=hearthstone+boom+bot

I just tested my api endpoint and it got quite slow now. Has the IA a timeout? (I will see what I can do to speed up the api.)

Akryum:
I have added a cache system to greatly speed up calls to my api so the IA works again after a few try (when the card cache is ready).

moollaza:
@Akryum yes there's a default timeout of 5000ms. 

Your 3 examples all worked for me, so I guess everything is good now?

Akryum:
Yes, evrything is working again :)

Issue #1310, opened 2014-12-02
InTheaters: Show more movies


jagtalon:
Would be great if we could show more than 12 movies.

![screen shot 2014-12-02 at 4 11 00 pm](https://cloud.githubusercontent.com/assets/81969/5270961/d6fcff84-7a3d-11e4-90f6-2d7f7f9a7036.png)

https://duck.co/ia/view/in_theaters

NeoSilky:
Just fixing this now, watch out for the pull request!

jagtalon:
Fixed by @NeoSilky in #1453

PR #1453, opened 2015-01-28
InTheatres - Changed number of movies shown to 30

NeoSilky:
What did I fix?

* Basically this issue - #1310. A simple change to 30 movies shown. I can further change it to whatever value you need but I think 30 should be enough.


moollaza:
LGTM

Would like @jagtalon to signoff.

jagtalon:
@NeoSilky :+1: 

g47wl7n

PR #1456, opened 2015-01-29
.travis.yml: Remove sudo: false

jagtalon:
Since .travis.yml is autogenerated by https://github.com/SineSwiper/Dist-Zilla-TravisCI, we can't just modify that file ourselves. We'd have to wait for https://github.com/SineSwiper/Dist-Zilla-TravisCI/issues/25 to get fixed first.

@moollaza

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 ---
-sudo: false
 before_install:
   - export HARNESS_OPTIONS=j10:c HARNESS_TIMER=1
   - rm .travis.yml


moollaza:
Understood. Thanks @jagtalon

jagtalon:
@moollaza Sorry I completely linked to the wrong repo! We still can't do it, but just wanna point out that we're using https://github.com/Getty/p5-dist-zilla-plugin-travisci

moollaza:
@jagtalon wow, good catch I totally thought we were using the other one. We should definitely switch to that one (no one is mainting ours, the other looks like it's well supported/reviewed) and then I don't think we need to maintain an actual copy of the file in the repo (which is very confusing)

kentfredric:
>  Hopefully then we won't need to have the actual file in the repo, which is very confusing!

There's no way to have the file missing without having a build branch.

Issue #1462, opened 2015-01-30
Transit/SEPTA: On time should be color #5B9E4D (Green)


jagtalon:
![screen shot 2015-01-30 at 3 17 32 pm](https://cloud.githubusercontent.com/assets/81969/5982652/3053c53a-a893-11e4-9478-94a7a71a5d2a.png)

jagtalon:
Similar to https://duckduckgo.com/?q=next+train+from+Metropark+to+New+York+Penn&ia=njtransit

javathunderman:
@jagtalon I'll give it my best shot.

javathunderman:
@jagtalon Not completely sure that this is a universal fix (it seems a bit too easy to change a CSS color value and be done with it), so let me know if I missed something!

jagtalon:
@javathunderman it looks good!

jagtalon:
Fixed by @javathunderman in #1464

PR #1464, opened 2015-01-31
Fix an issue with on time trains and design.

javathunderman:
This should fix issue #1462. I'm not completely sure that this works with trains that are to arrive in 1 minute or something like that, but feel free to take a look. 

cc @jagtalon

### DIFF
--- a/share/spice/transit/septa/transit_septa.css
+++ b/share/spice/transit/septa/transit_septa.css
@@ -36,7 +36,7 @@
 }
 
 .tile--septa .septa__badge {
-    background-color: #e0e0e0;
+    background-color: #5B9E4D;
     height: 12px;
     border-radius: 6px;
     padding: 0 6px;
@@ -59,5 +59,5 @@
 .tile--septa .septa__cancelled-tile .septa__time,
 .tile--septa .septa__cancelled-tile .septa__details,
 .tile--septa .septa__cancelled-tile .septa__status {
-    color: #bbb;
+    color: #5B9E4D;
 }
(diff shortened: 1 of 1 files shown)


share/spice/transit/septa/transit_septa.css
```diff
@@ -59,5 +59,5 @@
 .tile--septa .septa__cancelled-tile .septa__time,
 .tile--septa .septa__cancelled-tile .septa__details,
 .tile--septa .septa__cancelled-tile .septa__status {
-    color: #bbb;
-}
\ No newline at end of file
+    color: #5B9E4D;
```
mattr555: Hey @javathunderman, this is actually the color of the text when the train is cancelled. We gray out the tile when that happens, and so the color of the text is changed. It's probably left better as gray instead of green :smile:

javathunderman: @mattr555 Fairly sure that something was wrong there. Will fix! :+1:

javathunderman:
@mattr555 Thanks for picking that up! I couldn't see any canceled trains (probably because no one listens to weather forecasters after "Snowmageddon"). :smile:

moollaza:
@javathunderman mind posting a pic? Generally we ask the devs post a screenshot when they've modified the look of an IA so we can get a quick idea of what's changed

javathunderman:
fixedstatus
Right, here it is!

javathunderman:
@moollaza @jagtalon Anything I need to fix?

moollaza:
@javathunderman looking at NJT I'm seeing that we're using is `#60AE50` (for the Green), can you please switch to that?

javathunderman:
Didn’t the issue by @jagtalon say to use a different color code? Anyway, yeah, I’ll fix it.

javathunderman:
Should be ok. @moollaza

mattr555:
@javathunderman Looks good, thanks!

Issue #1458, opened 2015-01-29
GitHub: Showing results when the query is just "GitHub"


moollaza:
A search for "GitHub" shows a list of GitHub repos. We shouldn't be popping this IA unless there's more to the query than just the trigger.

https://duckduckgo.com/?q=GitHub&ia=software

puskin94:
Hei @moollaza , I will fix this soon ;)

MrChrisW:
Fixed by @puskin94 in #1467

PR #1467, opened 2015-01-31
Fix Github Spice: trigger

puskin94:
This fix the issue reported by @moollaza https://github.com/duckduckgo/zeroclickinfo-spice/issues/1458.
Now the trigger "github" without any parameters will not works. Modified the test file.
The fix is a bit "rude", tell me if this is ok ;)

### DIFF
--- a/lib/DDG/Spice/Github.pm
+++ b/lib/DDG/Spice/Github.pm
@@ -20,12 +20,8 @@ handle remainder_lc => sub {
 
 	s/^github\s+|\s+github$//;
 	return unless $_;
-    return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
-   	return;
-	
-	
+	return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
 	return;
-    
 };
 
 1;
--- a/lib/DDG/Spice/Github.pm
+++ b/lib/DDG/Spice/Github.pm
@@ -17,9 +17,14 @@ spice to => 'https://api.github.com/legacy/repos/search/$1?callback={{callback}}
 spice proxy_cache_valid => '200 30d';
 
 handle query_lc => sub {
-    s/^github\s+|\s+github$//;
-    return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
-    return;
+	my $param = shift;
+	if ($param =~ /^github (.+)|(.+) github$/i) {
+		s/^github\s+|\s+github$//;
+    	return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
+    	return;
+	}
+	return;
+    
 };
 
 1;
--- a/lib/DDG/Spice/Github.pm
+++ b/lib/DDG/Spice/Github.pm
@@ -16,13 +16,13 @@ triggers startend => "github";
 spice to => 'https://api.github.com/legacy/repos/search/$1?callback={{callback}}';
 spice proxy_cache_valid => '200 30d';
 
-handle query_lc => sub {
-	my $param = shift;
-	if ($param =~ /^github (.+)|(.+) github$/i) {
-		s/^github\s+|\s+github$//;
-    	return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
-    	return;
-	}
+handle remainder_lc => sub {
+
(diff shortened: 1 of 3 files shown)


MrChrisW:
@puskin94 This works. However I think a more standard way would be like this: 

1. Change `query_lc` to `remainder_lc` [line 19](https://github.com/puskin94/zeroclickinfo-spice/blob/git/lib/DDG/Spice/Github.pm#L19-Lundefined)

2. Use `return unless $_;` to against "no answer" [example](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/template/lib/DDG/Spice/Example.pm#L34) - this will prevent this triggering on just 'github'

and finally *(optional fix - if you feel like it :wink:)*: 

3. Update the javascript to grab the query remainder from the `$(script).attr("src")` instead of using `DDG.get_query()` [line 10](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/github/github.js#L10-L11) - [More information on doing this](https://github.com/duckduckgo/duckduckgo-documentation/blob/master/duckduckhack/spice/spice_advanced_frontend.md) this avoids using `.replace(/^\s*github\s+/, "");`

Ultimately this requires less code and removes the need for a regexp conditional statement.

puskin94:
@MrChrisW Thanks for the feedback! I learned a lot from this suggestions ;) a new commit is waiting your approval!

lib/DDG/Spice/Github.pm
```diff
@@ -16,10 +16,16 @@ triggers startend => "github";
 spice to => 'https://api.github.com/legacy/repos/search/$1?callback={{callback}}';
 spice proxy_cache_valid => '200 30d';
 
-handle query_lc => sub {
-    s/^github\s+|\s+github$//;
+handle remainder_lc => sub {
+
+	s/^github\s+|\s+github$//;
```
MrChrisW: This line can be removed as `github` will not be included in `$_` as we're only using the `remainder`

lib/DDG/Spice/Github.pm
```diff
@@ -16,10 +16,16 @@ triggers startend => "github";
...
+	return unless $_;
     return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
-    return;
+   	return;
+	
+	
+	return;
+
```
MrChrisW: Can you indent all code in the `handle` function using 4 spaces (soft tabs) as per the [code styleguide](https://duck.co/duckduckhack/code_styleguide)

share/spice/github/github.js
```diff
@@ -7,8 +7,10 @@
           return Spice.failed('github');
         }
 
-        var query = DDG.get_query()
-                    .replace(/^\s*github\s+/, "");
+        var script = $('[src*="/js/spice/github/"]')[0],
+                    source = $(script).attr("src"),
+                    query = source.match(/github\/([^\/]+)/)[1];
+        //console.log(query);
```
MrChrisW: Remove `console.log` :+1:

puskin94:
@MrChrisW ops, I had to delete this line but I forgot to do it :)

lib/DDG/Spice/Github.pm
```diff
@@ -16,10 +16,10 @@ triggers startend => "github";
...
 spice proxy_cache_valid => '200 30d';
 
-handle query_lc => sub {
-    s/^github\s+|\s+github$//;
-    return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
-    return;
+handle remainder_lc => sub {
+	return unless $_;
+	return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
+	return;
```
MrChrisW: Looks like you're still indenting with *TAB*, you should be able to switch your editor to indent using spaces when you hit the TAB key or just hit the spacebar 4 times :smiley:

puskin94:
@MrChrisW sorry, I just formatted my pc and I had to reconfigure all my settings ;)

MrChrisW:
@puskin94 No problem, happy to help! Thanks for making those changes so quick.

image 

Will test and merge shortly :+1:

puskin94:
@MrChrisW ahahah you made me laugh so much with that gif :) No problem! It is always nice to help improving DDG ;)

PR #1470, opened 2015-01-31
GoWatchIt: Don't display if we don't have any images for them

jagtalon:
This is a dirty fix, but the UI breaks when we display images that's not from us:

screen shot 2015-01-31 at 1 04 37 pm

Easiest fix is to remove them for now until we either have:

- Better images
- Resize their image so that it doesn't break the UI (this is also easy but requires extensive testing. Also GoWatchIt's images don't work well on retina screens)


jagtalon:
screen shot 2015-01-31 at 1 05 45 pm

moollaza:
LGTM :+1: 

Tested and working.

PR #1473, opened 2015-02-02
New spice to return words in specified part of speech

SueSmith:
This is my first attempt at a Spice so excuse any daft mistakes.. :wink: It uses Wordnik to return random words in particular parts of speech (noun, verb, adjective, pronoun, adverb, preposition and conjunction). If the query includes the part of speech in plural it returns five words, otherwise is returns one.

It gives people examples of words in a part of speech, which can be helpful when learning grammar. Rather than seeing a description of a grammatical idea, seeing examples is often the easiest way to understand what e.g. a noun, verb or whatever is.

Wordnik https://www.wordnik.com/

It's already being used and provides endpoints which allow you to specify the part of speech.

I couldn't see any that would allow you to specify the part of speech and would return JSON.

* noun example
* random adjective
* random verbs
* example adverbs

Anyone learning English grammar or who likes random stuff.

No.

I'm unsure what happens when triggers use the same words - see below, but it may interfere with the random name and random word answers.

Yes please!
* I'm not clear on whether this will interfere with existing Instant Answers with "random" in their triggers, e.g. the random name and random word triggers..? If so, should I alter my triggers to cover the full phrases e.g. "random noun" (rather than "random" with the handle function picking up the part of speech from the remainder)..? There would be a load of possible phrases, so in that case I'm wondering if it would be better to go down the regex route..? I could also extend this to cover other phrases e.g. "noun examples", which would also require a bit more processing, but the approach will depend on what I should do with the triggers generally so any guidance is appreciated!
* I'm totally new to Perl so please let me know if my code is inefficient or whatever :grimacing: 
* At the front end I've used the random word and rhyme Instant Answers as the basis for the layout, if it doesn't look right give me a shout!
* I've included the API key using the code in the dictionary Instant Answers, assuming you'll want to use the same key..? This is it: `{{ENV{DDG_SPICE_WORDNIK_APIKEY}}}` (Tested it with my own key..)
* I've built a minimum corpus count into the Wordnik query to weed out the more obscure words which might confuse people.. I've also only included seven parts of speech as the rest are a bit trickier to implement, but I could certainly look at extending this.

DuckDuckGo already uses the API.

random_single
random_list

## Checklist

```
[] Tested cross-browser compatibility
    - Ubuntu
        [x] Firefox

```

### DIFF
--- a/lib/DDG/Spice/RandPOS.pm
+++ b/lib/DDG/Spice/RandPOS.pm
@@ -1,0 +1,36 @@
+package DDG::Spice::RandPOS;
+# ABSTRACT: Returns a random word or list of words from the specified part of speech using Wordnik.
+
+use DDG::Spice;
+
+name 'Random Part of Speech Words';
+source 'Wordnik';
+description 'Returns one or more random words in the specified part of speech.';
+primary_example_queries 'random noun','example adjectives','adverb example';
+category 'language';
+topics 'words_and_games','everyday';
+code_url 'https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/RandPOS.pm';
+attribution github => ['https://github.com/SueSmith', 'SueSmith'],
+    web => ['http://benormal.info','Sue Smith'],
+    twitter => 'braindeadair';
+
+triggers start => 'random';
+triggers startend => 'example';
+
+spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
+spice to => 'http://api.wordnik.com/v4/words.json/randomWords?includePartOfSpeech=$1&minCorpusCount=10000&limit=$2&api_key={{ENV{DDG_SPICE_WORDNIK_APIKEY}}}&callback={{callback}}';
+
+handle remainder => sub {
+
+    return unless $_ =~ /^(noun|nouns|verb|verbs|adjective|adjectives|pronoun|pronouns|preposition|prepositions|conjunction|conjunctions|adverb|adverbs)$/i;
+
+    if($_ =~ /nouns|verbs|adjectives|pronouns|prepositions|conjunctions|adverbs/){
+        return substr($_, 0, -1), 5;
+    }
+    if ($_ =~ /noun|verb|adjective|pronoun|preposition|conjunction|adverb/){
+        return $_, 1;
+    }
+    return;
+};
+
+1;
--- a/lib/DDG/Spice/RandPOS.pm
+++ b/lib/DDG/Spice/RandPOS.pm
@@ -19,6 +19,7 @@ triggers startend => 'example';
 
 spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
 spice to => 'http://api.wordnik.com/v4/words.json/randomWords?includePartOfSpeech=$1&minCorpusCount=10000&limit=$2&api_key={{ENV{DDG_SPICE_WORDNIK_APIKEY}}}&callback={{callback}}';
+spice proxy_cache_valid => '418 1d';
 
 handle remainder => sub {
 
--- a/share/spice/rand_pos/content.handlebars
+++ b/share/spice/rand_pos/content.handlebars
@@ -1,0 +1,1 @@
+ {{#concat words sep=" "}}<a href="http://wordnik.com/words/{{word}}">{{word}}</a>{{/concat}}
--- a/share/spice/rand_pos/content.handlebars
+++ b/share/spice/rand_pos/content.handlebars
@@ -1,1 +1,1 @@
- {{#concat words sep=" "}}<a href="http://wordnik.com/words/{{word}}">{{word}}</a>{{/concat}}
+ {{#concat words sep=", "}}<a href="http://wordnik.com/words/{{word}}">{{word}}</a>{{/concat}}
--- a/share/spice/rand_pos/rand_pos.js
+++ b/share/spice/rand_pos/rand_pos.js
@@ -1,0 +1,31 @@
+(function(env){
+    "use strict";
+    
+    env.ddg_spice_rand_pos = function(api_result){
+        if (api_result.error) {
+            return Spice.failed('rand_pos');
+        }
+        var words = [];
+        var len = api_result.length;
+        for(var i=0; i<len; i++) {
+            words.push(api_result[i]);
+        }
+        Spice.add({
+            id: "rand_pos",
+            data: {words:words},
+            name: "Answer",
+            meta: {
+                sourceUrl: 'http://wordnik.com',
+                sourceName: 'Wordnik',
+                sourceIcon: true
+            },
+            templates: {
+                group: 'base',
+                options: {
+                    content: Spice.rand_pos.content,
+                    moreAt: true
+                }
+            }
+        });
+    }
+}(this));
--- a/t/RandPOS.t
+++ b/t/RandPOS.t
@@ -1,0 +1,105 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+use Test::More;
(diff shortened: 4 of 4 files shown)


MrChrisW:
Hi @SueSmith thanks a lot for making this! We really appreciate it. Code will be reviewed shortly :+1:
 
After a quick glance it may be possible to integrate your triggers/changes directly into the [RandWord](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/RandWord.pm), as they both use the same endpoint, just a different parameter `includePartOfSpeech`. 

Maybe we can add the `includePartOfSpeech` parameter if the user searches for noun, nouns, verb, verbs, adjective. I'll investigate further. :smile:

SueSmith:
Ah cool thanks for that info @MrChrisW - the only other thing mine does is to include a limit parameter if the search query includes plural, so if the query is e.g. "random nouns" it returns a list of five rather than a single one.. It also uses the "randomWords" endpoint rather than the "randomWord" one.. Happy to make whatever changes will help :smile_cat:

MrChrisW:
@moollaza Is it possible to modify the endpoint [like this](https://github.com/MrChrisW/zeroclickinfo-spice/blob/b0a27fa3e216a12a3d0bc3c05fbdb978a3a2aef5/lib/DDG/Spice/RandPOS.pm#L39-L41)? The only issue I'm having is the URL encoding after `spice to` is sent for fetching.  

This means we can use two different wordnik endpoints in a single IA `randomWord` and `randomWords`.

Otherwise we'll just use two IA's :smiley:

moollaza:
Hey guys, apologies for delayed response,

> @moollaza Is it possible to modify the endpoint like this? The only issue I'm having is the URL encoding after spice to is sent for fetching. (ignore any bad code)

Unfortunately @MrChrisW that won't work. The content returned from `handle` gets URL encoded later in the process so that's not going to work (the `&` will get encoded and then the params won't be recognized).

Adding support for multiple endpoints is definitely something I'd love to tackle but it's not a major concern right now.

In the interest of moving this along, lets just stick with two separate IA's. We can always merge them later if it becomes an option.

moollaza:
@SueSmith, to answer your questions:

>I'm not clear on whether this will interfere with existing Instant Answers with "random" in their triggers, e.g. the random name and random word triggers..?

Good question! No this won't affect other IA's. Multiple Spice IA's can use the same trigger words and all can return an answer. They'll each get their own Tab on the AnswerBar (beside the static Images and Videos tabs)

>If so, should I alter my triggers to cover the full phrases e.g. "random noun" (rather than "random" with the handle function picking up the part of speech from the remainder)..?

The only benefit to having more explicit triggers like `["random noun", "random verb", ...]` is that you know for sure the query applies to your IA and it means you don't have to do any work in your `handle` to validate the query.

If you wanted to prepare a .txt or .yml file with a list of all these queries, it can be `slurp`ed and used to generate an array of triggers which you can pass to `spice triggers => @triggers`. This method would essentially result in the same triggering performance as what you currently have though, so either approach is fine. (I suppose it would technically be slightly faster because running a regex against the query to check for other words is a more expensive operation)

>I'm totally new to Perl so please let me know if my code is inefficient or whatever :grimacing:

That's what we're here for! @MrChrisW  and our other Community Leaders give all PR's a first look at the code and then myself or another DDG Staff member will give the code a sign-off.

>At the front end I've used the random word and rhyme Instant Answers as the basis for the layout, if it doesn't look right give me a shout!

Not a problem. After the code has been approved our design team will also sign off on the look of the IA. If you're using a pre-defined template it's likely not going to cause any problems.

>I've included the API key using the code in the dictionary Instant Answers, assuming you'll want to use the same key..? This is it: {{ENV{DDG_SPICE_WORDNIK_APIKEY}}} (Tested it with my own key..)

Yup that's perfect, our system already has a WordNik API key that we'll sub into there.

>I've built a minimum corpus count into the Wordnik query to weed out the more obscure words which might confuse people.. I've also only included seven parts of speech as the rest are a bit trickier to implement, but I could certainly look at extending this.

Perfect, that sounds good for now :+1:

lib/DDG/Spice/RandPOS.pm
```diff
@@ -0,0 +1,36 @@
...
+attribution github => ['https://github.com/SueSmith', 'SueSmith'],
+    web => ['http://benormal.info','Sue Smith'],
+    twitter => 'braindeadair';
+
+triggers start => 'random';
+triggers startend => 'example';
+
+spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
+spice to => 'http://api.wordnik.com/v4/words.json/randomWords?includePartOfSpeech=$1&minCorpusCount=10000&limit=$2&api_key={{ENV{DDG_SPICE_WORDNIK_APIKEY}}}&callback={{callback}}';
+
```
moollaza: Seeing as this is a "random" Spice, we don't want to cache the API response.

You'll need to add `spice proxy_cache_valid => '418 1d'` to this like the RandWord Spice does: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/RandWord.pm#L18

Details: https://duck.co/duckduckhack/spice_advanced_backend#caching

moollaza:
@abeyang this one is ready for a design sign-off. It has the same design as our RandWord Spice, https://duckduckgo.com/?q=random+word&ia=answer, except this one has different triggers and gives different answers

abeyang:
I'm sorry to be nitpicky, but when I looked at the screenshot w/ 5 verbs, I found that I was trying to read it as a sentence. Can we add (non-linked) commas in-between? So it should be like:

> endanger, propping, scoffed, secures, volunteered

share/spice/rand_pos/content.handlebars
```diff
@@ -0,0 +1 @@
+ {{#concat words sep=" "}}<a href="http://wordnik.com/words/{{word}}">{{word}}</a>{{/concat}}
```
moollaza: as per @abeyang's suggestion, let's change the `sep` to `,`

moollaza:
>I found that I was trying to read it as a sentence. Can we add (non-linked) commas in-between? So it should be like:

@abeyang great suggestion, easy to fix and definitely an improvement. I'll make an issue for the Random Word Spice, to be updated as well.

MrChrisW:
@moollaza thanks for clearing that up. Just a suggestion! Using multiple endpoints seems like a simple thing to implement if URL encoding is the only issue, however I understand its not a priority.

SueSmith:
@moollaza Thanks for all this info! I've added the cache line and altered the separator - I had a comma in there originally but couldn't decide if it looked better with or without.. 

I've left the triggers as they are for the moment - if I used a file to slurp the exact phrases ('random nouns' etc) I think I'd need to alter the handle function to carry out quite a bit more processing (on the whole query rather than just the remainder), so that I could figure out which part of speech has been entered/ whether it's plural, not sure what impact that would have on efficiency.. In any case I'd like to extend it to cover more phrases at some point so could look at that later if it's ok for now..?

Hope that makes sense!

Cheers @MrChrisW @abeyang :smile_cat:

moollaza:
>In any case I'd like to extend it to cover more phrases at some point so could look at that later if it's ok for now..?

@SueSmith yup, this is great for now :+1:

MrChrisW:
@moollaza Ping! Pretty sure this is right to be merged :+1:

moollaza:
@SueSmith @MrChrisW whoops, I forgot about this one -- very sorry. It LGTM :+1:

moollaza:
I just realized this one fell through the cracks! Will set it up for deployment today. So sorry @SueSmith!

moollaza:
@SueSmith congrats! It's live: https://duckduckgo.com/?q=random+noun&ia=answer

PR #1481, opened 2015-02-03
Tile Variant Changes

sdougbrown:
*Requires an internal PR, please don't merge until that is released.*

We're trying to make it easier for all DDH devs to create great-looking instant answers.  The first step is to make a bunch of 'variants' available to you when crafting a text tile.  This should eliminate the need for you to take guesses at the correct line-height and sizes of your titles, snippets, footers, and more.

We're working out some of the kinks still, but this PR should convert all of the affected instant answers to the new method.


moollaza:
@jagtalon @sdougbrown can we use our image tool to generate pics and verify everything looks good?

jagtalon:
@moollaza yeah I think we can do that.

sdougbrown:
Internal PR merged, this is good to go. :smile:

PR #1474, opened 2015-02-02
SalesTax Instant Answer

MrChrisW:
**What does your Instant Answer do?**

Returns the sales tax of a state in the United States

**What is the data source for your Instant Answer? (Provide a link if possible)**

http://taxratesapi.avalara.com/docs

**Are there any other alternative (better) data sources?**

Yes, paid sources. 

**What are some example queries that trigger this Instant Answer?**

"sales tax for pennsylvania", "sales tax for pa"

**Which existing Instant Answers will this one supercede/overlap with?**

Yes, [SalesTax Goodie](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/917)
Fixing [#960](https://github.com/duckduckgo/zeroclickinfo-goodies/issues/960)

**Screenshots**

selection_175
selection_174

//CC @moollaza @javathunderman

PR #1484, opened 2015-02-05
Update sourceName removing FQDN extention

MrChrisW:
Possible improvement to `sourceName`. I've left timeanddate.com and xe.com as these are brand names including the extention.

https://github.com/duckduckgo/zeroclickinfo-spice/pull/1474#issuecomment-72935193

CC// @jagtalon


jagtalon:
Nice--didn't know we had that many :+1:

PR #1489, opened 2015-02-06
Fix vertical centering of UV Index value

moollaza:
/cc @jagtalon @abeyang @bjoern

Fixed the CSS. Tested on Windows 8 -- FF, Win 7 -- IE 8, OSX Yosemite -- Safari & Chrome. Should be good :)

2015-02-06_11h11_00


jagtalon:
Looks good on mobile as well. Would like to wait for what @chrismorast says about the favicon though and why it needs to be changed.

moollaza:
Alternative fix in #1491

Issue #1486, opened 2015-02-06
Books: Cloud icon not aligning with percentage


jagtalon:
![screen shot 2015-02-06 at 3 21 23 am](https://cloud.githubusercontent.com/assets/81969/6076278/56c2d4d8-adaf-11e4-931d-ffd6e0640fb3.png)

Using Chrome on a Mac

samph:
Before:

![screen shot 2015-02-07 at 10 56 16](https://cloud.githubusercontent.com/assets/468118/6091862/12d7ebdc-aeb8-11e4-8038-febb72ff1230.png)

After:
![screen shot 2015-02-07 at 10 56 06](https://cloud.githubusercontent.com/assets/468118/6091861/08f621f6-aeb8-11e4-87dc-b38a27e8575b.png)

jagtalon:
Fixed by @samph in #1494

PR #1494, opened 2015-02-07
Rating now aligns with cloud icon as per Issue #1486

samph:
Addressed Issue #1486 

Before:

screen shot 2015-02-07 at 10 56 16

After:
screen shot 2015-02-07 at 10 56 06

### DIFF
--- a/share/spice/book/book.css
+++ b/share/spice/book/book.css
@@ -64,7 +64,7 @@
     float: right;
     line-height: 1;
     font-size: 2em;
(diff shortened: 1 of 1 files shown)


MrChrisW:
@samph Thanks for fixing that. @jagtalon can sign off :+1:

jagtalon:
Nice! Looks like we still have trouble with some layouts but this is good!

screen shot 2015-02-19 at 3 59 14 pm

PR #1503, opened 2015-02-08
Updating Github Spice

joerussbowman:
This does the following:

 - Updates the Github Spice to use the current version of the Github API
 - Adds language detection to the query parser, adds detected languages as the language parameter passed to Github
 - Does remove the bare github search test, my testing shows it was working correctly showing Github status.

This is in lieu of the previous GithubLanguages Spice I was working on.

### DIFF
--- a/lib/DDG/Spice/Github.pm
+++ b/lib/DDG/Spice/Github.pm
@@ -33,7 +33,6 @@ handle query_lc => sub {
     
     my $query = $_;
     my $l; 
-    #if (/ ($langs) /|| /^($langs) / || / ($langs)$/) {
     if (/\b($langs)\b/) {
         $l = $1;
         $query =~ s/ ($langs) |^($langs) | ($langs)$//;
@@ -47,7 +46,9 @@ handle query_lc => sub {
         # Github. You specify language as a part of the raw query string
         # passed to the api like on the web form interface. 
         return "${query} language:\"${l}\"" unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
-    } 
+    } else {
+	return $query;
+    }
     
     return;
 };
--- a/lib/DDG/Spice/Github.pm
+++ b/lib/DDG/Spice/Github.pm
@@ -2,6 +2,7 @@ package DDG::Spice::Github;
 # ABSTRACT: Search for information about GitHub repositories
 
 use DDG::Spice;
+use String::Trim;
 
 primary_example_queries "github zeroclickinfo";
 description "Github info";
@@ -37,13 +38,9 @@ handle query_lc => sub {
         $query =~ s/ ($langs) |^($langs) | ($langs)$//;
 
         # make sure there is an actual query, and not just a language term search
-        for ($query) {
-            $query =~ s/^\s*//; 
-            $query =~ s/\s*$//;
-            if ($query eq "") {
-                return;
-            }
-        }
+        trim($query);  # need to add `use String::Trim`
+        s/\s+/ /g;     # squash down multiple spaces
+        return unless length $_;
 
         # These is no separate language parameter for the query to 
         # Github. You specify language as a part of the raw query string
--- a/lib/DDG/Spice/Github.pm
+++ b/lib/DDG/Spice/Github.pm
@@ -33,7 +33,8 @@ handle query_lc => sub {
     
     my $query = $_;
     my $l; 
-    if (/\b($langs)\b/) {
+    # can't use standard \b word boundary here. It will fail for languages with characters like c++
+    if (/ ($langs) / or /^($langs) / or / ($langs)$/) {
         $l = $1;
         $query =~ s/ ($langs) |^($langs) | ($langs)$//;
 
--- a/lib/DDG/Spice/Github.pm
+++ b/lib/DDG/Spice/Github.pm
@@ -33,35 +33,30 @@ spice proxy_cache_valid => '200 30d';
 
 handle query_lc => sub {
     s/^github\s+|\s+github$//;
+    if ($_ eq "") {
+        return;
+    }
     
     my $query = $_;
     my $l; 
-    # structuring the search this way gives the first langauge specified
-    # precedence. So have 'python javascript' passes the query
-    # 'javascript language:python' to github.
-    if (/($langs) /||/ ($langs)/ || /^($langs)/ || /($langs)$/) {
+    if (/ ($langs) /|| /^($langs) / || / ($langs)$/) {
         $l = $1;
-        $query =~ s/($langs) | ($langs)|^($langs)|($langs)$//;
+        $query =~ s/ ($langs) |^($langs) | ($langs)$//;
 
         # make sure there is an actual query, and not just a language term search
-        #for ($query) {
-        #    $query =~ s/^\s*//; 
-        #    $query =~ s/\s*$//;
-        #    if ($query eq "") {
-        #        # playing around with the main site, the live version does load
-        #        # something for just 'github bash' so go ahead and pass the query
-        #        return $_;
-        #    }
-        #}
+        for ($query) {
+            $query =~ s/^\s*//; 
+            $query =~ s/\s*$//;
+            if ($query eq "") {
+                return;
+            }
+        }
 
         # These is no separate language parameter for the query to 
         # Github. You specify language as a part of the raw query string
         # passed to the api like on the web form interface. 
-        return "${query} language:\"${l}\"";
+        return "${query} language:\"${l}\"" unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
     } 
-    
-    
-    
     
     return $_ unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
     return;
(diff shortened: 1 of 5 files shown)


lib/DDG/Spice/Github.pm
```diff
@@ -13,13 +13,45 @@ attribution web => ['http://dylansserver.com','Dylan Lloyd'],
...
+            $query =~ s/\s*$//;
+            if ($query eq "") {
+                return;
+            }
+        }
+
+        # These is no separate language parameter for the query to 
+        # Github. You specify language as a part of the raw query string
+        # passed to the api like on the web form interface. 
+        return "${query} language:\"${l}\"" unless /^jobs\b|\bjobs$|^status\b|\bstatus$/;
```
moollaza: I would suggest moving the `unless` to the top:

```perl
if ($_ eq "" || m/\bjobs\b|\bstatus\b/) {
    return;
}
```

Also I believe `\b` covers both `^` and `$` -- will double check!

t/Github.t
```diff
@@ -18,8 +18,7 @@ ddg_spice_test(
         caller => 'DDG::Spice::Github'
     ),
     
-    'github status' => undef,
-    'github' => undef
+    'github status' => undef
```
moollaza: Please add some test queries for the new functionality you've added :smile:

lib/DDG/Spice/Github.pm
```diff
@@ -13,13 +13,45 @@ attribution web => ['http://dylansserver.com','Dylan Lloyd'],
...
-    return unless $_;
+handle query_lc => sub {
+    s/^github\s+|\s+github$//;
+    if ($_ eq "") {
+        return;
+    }
+    
+    my $query = $_;
+    my $l; 
+    if (/ ($langs) /|| /^($langs) / || / ($langs)$/) {
```
moollaza: I think we should also accept the GitHub search syntax `lang: c++`

moollaza: Given that `\b` matches all 3 cases here, `if (\b$langs\b)` should suffice

lib/DDG/Spice/Github.pm
```diff
@@ -13,13 +13,45 @@ attribution web => ['http://dylansserver.com','Dylan Lloyd'],
...
+        $l = $1;
+        $query =~ s/ ($langs) |^($langs) | ($langs)$//;
+
+        # make sure there is an actual query, and not just a language term search
+        for ($query) {
```
moollaza: Don't think you need the loop here:

```perl
trim($query);  # need to add `use String::Trim`
s/\s+/ /g;     # squash down multiple spaces
return unless length $_;
```

lib/DDG/Spice/Github.pm
```diff
@@ -13,13 +13,45 @@ attribution web => ['http://dylansserver.com','Dylan Lloyd'],
...
 triggers startend => "github";
-spice to => 'https://api.github.com/legacy/repos/search/$1?callback={{callback}}';
+
+my @triggers = share("triggers.txt")->slurp;
+triggers startend => @triggers;
+
+chomp(@triggers);
+my $langs = join("|", map(quotemeta, @triggers));
+
+spice to => 'https://api.github.com/search/repositories?q=$1&sort=stars&callback={{callback}}';
```
moollaza: :+1:

moollaza:
@joerussbowman thanks a lot for this! Off to a great start, I've left some feedback :+1:

joerussbowman:
On the \b stuff, I found for some cases that didn't work out so well. c++ for example would end up witha query like ++ test lang: c, I think graphiz (dotviz) or something similar like that also had problems. That's why I did the regex the way I did.

I will work on the other suggestions later this week. Just got a new laptop and am setting it up.

moollaza:
>On the \b stuff, I found for some cases that didn't work out so well. c++ for example would end up witha query like ++ test lang: c,

Ah yes. You likely want to `sort` the list of languages from longest to shortest so it captures the larger ones first. `c` comes before `c++` in the file so it stopped after it matched `\bc\b`.

>I will work on the other suggestions later this week. Just got a new laptop and am setting it up.
Awesome, sound good! I'm jealous, what kinda laptop?!

joerussbowman:
For this statement:

"I think we should also accept the GitHub search syntax lang: c++"


Are you suggesting I filter out the language detection for when someone does something like "github redis lang: c++" as a search parameter duckduckgo so that the string passed to the github api is "redis lang: c++"? That seems like a bit of search site encapsulation? I'd think if someone is going to that level of using specific parameters for the github search they'd already be doing that on github? How would that affect the general search results that DuckDuckGo provides? 

I mean it should be easy enough to do, I can just look for "lang:" and stop parsing if it's there. I'm just concerned that might be a bit of overthinking the implementation?

joerussbowman:
Oh sorry, never replied. I got a sager np8651

joerussbowman:
Sorry for the delay, I have to admit I've actually been playing video games with my computer time since I got my new laptop. It has a real nice video card :)

moollaza:
@joerussbowman sorry about the delay, I'll take a look at this next week.

BTW, how's the Sager? I've heard good things about them!

joerussbowman:
It's an awesome machine, really well built. A bit overpowered and the case seems a bit flimsy compared to machines my old dell xps 13 and my wifes lenovo yoga 11s. But, I use those for being portable, this is more of a desktop replacement that lets me run a bunch of VMs and/or play some video games when the kids monopolize the tv

share/spice/github/github.js
```diff
@@ -62,10 +63,10 @@
 }(this));
 
 // Make sure we display only three items.
-Handlebars.registerHelper("GitHub_last_pushed", function(pushed) {
+Handlebars.registerHelper("GitHub_last_pushed", function(pushed_at) {
```
moollaza: @joerussbowman we now have moment.js available to use. If you wrap `Spice.add()` in `DDG.require('moment.js', function() { ... });` you'll have `moment()` available to you which should allow you to get rid of this helper, or at least simplify it.

lib/DDG/Spice/Github.pm
```diff
@@ -13,13 +14,43 @@ attribution web => ['http://dylansserver.com','Dylan Lloyd'],
             email => ['dylan@dylansserver.com','Dylan Lloyd'];
 
 triggers startend => "github";
-spice to => 'https://api.github.com/legacy/repos/search/$1?callback={{callback}}';
+
+my @triggers = share("triggers.txt")->slurp;
+triggers startend => @triggers;
```
moollaza: After a little more consideration, I'm afraid that given the number of dictionary words that exist in the triggers.txt I'm not sure we should allow the triggers to also be the languages. It's likely that this will cause a lot of false-positive triggering.

Let's stick with `github` as the trigger and if a language exists in the query, use that to specify the language of the result repos.

share/spice/github/triggers.txt
```diff
@@ -0,0 +1,315 @@
...
+ioke
+io
+irc log
+isabelle
+jade
+java server pages
+json5
+jsoniq
+jsonld
+json
```
moollaza: I feel like JSON, HTTP and similar should be removed from this list. It's more likely that someone is looking for a JSON parser in another language/platform then code written in JSON.

joerussbowman:
I'll poke at these soon. Been doing some research for my next home project idea. I should hopefully be able to knock this stuff out this weekend.

jagtalon:
@joerussbowman 

- username is rendered as `[object Object]`

screen shot 2015-03-23 at 3 37 25 pm

- Clicking on "More at ..." doesn't work because query is encoded:

screen shot 2015-03-23 at 3 38 17 pm

- I think I'd like it if we specified if it was a language search because it's indistinguishable from the normal github search (cc @moollaza):

screen shot 2015-03-23 at 3 37 25 pm

joerussbowman:
I'll look at the first 2 when I can, sorry I should have caught those myself.

On the last one, to make sure I understand: If a language is detected and we append that to the query then you want me to changed the "Showing.." to reflect that it was limited by language? This is the guthub Spice, so with this change going live it is the "normal github search" so just making sure it's correct to get to that level of detail. 

Should I change it to something like "Showing 29 Git Repositories for 'test' written in c++"?

jagtalon:
@joerussbowman for the last one yeah that's what I was thinking although shorter like "Showing 29 Git Repositories (C++)" but I'm not sure if other people agree. @abeyang @moollaza how do you think we should show filters in general?

moollaza:
For now,

>"Showing 29 Git Repositories (C++)"

works.

Soon we'll have the dropdown which I think we'll be able to use the language. Ultimately it would be awesome if we could allow the user to requery with a different language using the dropdown.

jagtalon:
@moollaza thanks I agree. @joerussbowman yup, let's go with that! Should be good to merge right after.

joerussbowman:
The only thing I haven't done is the moment.js stuff. How strict of a requirement is that? I'm about to switch gears into a couple of ideas I have and the output is working as is.

jagtalon:
@joerussbowman not that important--as long as it works we can just make another PR for it.

jagtalon:
@joerussbowman looks good! :+1: 

screen shot 2015-03-26 at 3 13 31 pm

joerussbowman:
sweet! Thanks :)

PR #1513, opened 2015-02-10
use new ddg-wrapped jsonp api endpoints

josephwegner:
Solves the gzip issue on my API - now the DDG api responses come pre-wrapped with the JSONP callback. It's a bit brittle and coupled, so things will blow up if the callback naming format ever changes, but we can cross that bridge when we get there.


moollaza:
@josephwegner thanks a lot!

jagtalon:
@josephwegner nice! Tested and it works:

screen shot 2015-02-10 at 12 55 37 pm

jagtalon:
Just modified the `spice from` to `(.*)/(.*)` because nginx complains otherwise

josephwegner:
:open_mouth: that's got to be your local right? You couldn't have deployed that fast....

jagtalon:
@josephwegner oh yeah my local instance--should be up sometime this week though :)

PR #1516, opened 2015-02-10
Movie.pm - update example queries

MariagraziaAlastra:
Since the triggers changed, the example queries must be changed as well.


jagtalon:
:+1: Nice!

PR #1520, opened 2015-02-12
add metadata to repo

jdorweiler:
dzil grabs the metadata from the community platform and adds it to `share/spice/meta/`
@russellholt


lib/DDG/Spice/Meta.pm
```diff
@@ -0,0 +1,8 @@
+package DDG::Spice::Meta;
```
moollaza: I take it we need this for the share dir to work?

jdorweiler: Yes

russellholt:
:cactus:

PR #1521, opened 2015-02-12
Delete Translate.pm

moollaza:
This file is no longer used/needed.
/cc @MrChrisW


MrChrisW:
@moollaza Oops missed this file :grinning: good catch! LGTM :+1:

PR #1531, opened 2015-02-13
QuandlFundamentals Spice

brianrisk:
# Spice Pull Request Template

Using this template will help us better understand your Instant Answer and assist you when necessary.  Simply copy and paste the markdown below into the description of your GitHub pull request and complete as appropriate.

Returns fundamental values for companies given the presence of a ticker symbol (e.g. "AAPL", "msft") and indicator description (e.g. "revenue", "total assets", "book value").

Provides a live, current value for searches where the value is primary information sought:  "YHOO net income"

https://www.quandl.com/SF1/documentation/about

I am a developer with Quandl and have a deep understanding of its data sets and API.

Frankly, no.  Other offerings (Bloomberg, Morningstar, SEC) do not offer this open access via API.  The API access and data license are what makes this spice possible.

AAPL revenue
sales FB
what is the latest YHOO earnings per share

 investors and financial advisors

No

This will overlap with 'Stocks' which, at the moment, incorrectly returns price data when fundamentals data is requested.  QuandlFundamentals will correctly handle the fundamentals queries while letting through queries where the appropriate answer is price.

No, thanks.

An API key is required.  There will be no costs.

Abraham Thomas @athomasq

screen shot 2015-02-13 at 6 21 19 pm

## Checklist

```

    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 7
        [] IE 8

    - Mac OSX
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [X] Safari

    - iOS (iPhone)
        [X] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [] Google Chrome
        [] Opera
```


moollaza:
@brianrisk thanks a lot for making this! We'll be sure to review it shortly and give you some feedback. Would you mind emailing me an API key for testing? moollaza@duckduckgo.com

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,84 @@
...
+}
+
+# defining our triggers
+triggers any => @trigger_keys;
+
+# to set an environmental variable:
+# duckpan env set <name> <value>
+
+# set spice parameters
+spice to => 'http://quandl.com/api/v1/datasets/SF1/$1_MRQ.json?auth_token={{ENV{QUANDL_KEY_SF1}}}&rows=2';
```
moollaza: @brianrisk can you please update the API Key placeholder to: `{{ENV{DDG_SPICE_QUANDL_APIKEY}}}` -- that's our canonical format.

brianrisk:
Done @moollaza

lib/DDG/Spice/Quandl/HomeValues.pm
```diff
@@ -0,0 +1,90 @@
...
+attribution web => ["https://www.quandl.com", "Quandl"],
+            twitter => "quandl";
+            
+# load our trigger phrases
+my @trigger_lines = share('home_values_triggers.txt')->slurp;
+# hash to associate phrases with URL codes
+my %trigger_phrases = ();
+# array to preserve the order of the trigger phrases from the text file
+my @trigger_keys;
+foreach my $line (@trigger_lines) {
```
moollaza: We can simply almost all of this down to a single line if we convert `triggers.txt` to a YAML file. E.g:

```yaml
total assets: ASSETS
current assets: ASSETSC
...
```

See: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SalesTax.pm#L30, https://github.com/duckduckgo/zeroclickinfo-spice/blob/9d370de370b4094caa8802dac896b6c337ff619e/share/spice/sales_tax/zipcodes.yml

lib/DDG/Spice/Quandl/HomeValues.pm
```diff
@@ -0,0 +1,90 @@
...
+handle sub {
+
+    # split query phrase by spaces
+    my @words = split / /, $_;
+    
+    # go through each word of the query to see if it is a valid region
+    # for now, a region will only be a zip code
+    my $region;
+    my $indicator_type;
+    foreach my $word (@words) {
```
moollaza: We can use a simple regex for this, much fast too:

```perl
if ($_ =~ m/\b(\d{5})\b/){
    my $zipcode = $1;
}
```

moollaza: the `\b` make sure there's a word boundary on either side of the numbers, so we don't simply match 5 digits of a larger number.

Alternatively, to support longer zipcode you could use something like http://stackoverflow.com/a/2577239/819937

lib/DDG/Spice/Quandl/HomeValues.pm
```diff
@@ -0,0 +1,90 @@
...
+
+# to set an environmental variable:
+# duckpan env set <name> <value>
+
+# set spice parameters
+spice to => 'http://quandl.com/api/v1/datasets/ZILLOW/$1.json?auth_token={{ENV{DDG_SPICE_QUANDL_APIKEY}}}&rows=2';
+spice wrap_jsonp_callback => 1;
+spice proxy_cache_valid => "418 1d";
+
+handle sub {
```
moollaza: using `handle query_lc => {` will make sure you're working with the lowercased version of the query, with leading/trailing spaces removed.

moollaza: alternatively `handle query` will preserve the letter casing.

share/spice/quandl/fundamentals/fundamentals_triggers.txt
```diff
@@ -0,0 +1,54 @@
+total assets	ASSETS
```
moollaza: Can we alpha-sort this list? (and the others?) It would make it a little easier to read/update.

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,84 @@
...
+attribution web => ["https://www.quandl.com", "Quandl"],
+            twitter => "quandl";
+            
+# load our trigger phrases
+my @trigger_lines = share('fundamentals_triggers.txt')->slurp;
+# hash to associate phrases with URL codes
+my %trigger_phrases = ();
+# array to preserve the order of the trigger phrases from the text file
+my @trigger_keys;
+foreach my $line (@trigger_lines) {
```
moollaza: Same here, switching triggers to YAML would likely be better.

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,84 @@
...
+# defining our triggers
+triggers any => @trigger_keys;
+
+# to set an environmental variable:
+# duckpan env set <name> <value>
+
+# set spice parameters
+spice to => 'http://quandl.com/api/v1/datasets/SF1/$1_MRQ.json?auth_token={{ENV{DDG_SPICE_QUANDL_APIKEY}}}&rows=2';
+spice wrap_jsonp_callback => 1;
+spice proxy_cache_valid => "418 1d";
```
moollaza: Does the data update in real time? If not, can we cache responses for a bit? Would certainly speed things up!

moollaza: and reduce the load on your servers

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,84 @@
...
+
+handle sub {
+
+    # split query phrase by spaces
+    my @words = split / /, $_;
+    
+    # load our list of tickers
+    my %tickers = map { trim($_) => 0 } share('tickers.txt')->slurp;
```
moollaza: best to load this and create the variable outside of the `handle` function, otherwise we'll be slurping the file every time the IA is triggered instead of just once when the package loads on our machines

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,84 @@
...
+
+    # go through each word of the query to see if it is a valid ticker
+    my $ticker;
+    foreach my $word (@words) {
+        my $ucword = uc $word;
+        # save ticker and exit loop if found
+        if (exists $tickers{$ucword}) {
```
moollaza: I noticed the first ticker symbol is simply "A" (and a few others are also dictionary words) -- I'm a little weary as that's likely to create a lot of false positives because we're only checking the presence of the ticker and trigger *anywhere* in the query.

For example "price of a dog" triggers this IA currently. The combination of the trigger "price" and stock symbol "A" are likely to occur in a large subset of queries so we may need to narrow the trigger space of this to prevent a lot of unnecessary API calls.

The simplest approach is to only allow queries where the trigger and stock symbol are the only words, e.g "APPL ebit" or relax that slightly with a regex that allows for queries that are also formed as questions, e.g. https://github.com/duckduckgo/zeroclickinfo-spice/blob/9d370de370b4094caa8802dac896b6c337ff619e/lib/DDG/Spice/Nutrition.pm#L18

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,84 @@
...
+            $ticker = $ucword;
+            last;
+        }
+    };
+    
+    # only return if we found a ticker in the search query
+    my $query = lc $_;
+    if ($ticker) {
```
moollaza: `return unless $trigger` is a little better because we don't need to use an `if` block for the next part

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,84 @@
...
+        my @chunks = split /\t/, $line;
+        # add to our hash
+        $trigger_phrases{$chunks[0]} = $chunks[1];
+        # add to our array
+        push @trigger_keys, $chunks[0];
+    }
+}
+
+# defining our triggers
+triggers any => @trigger_keys;
```
moollaza: Another way to reign in the triggering (and reduce the query space) would be to use `triggers startend` so that we're only working with queries that begin or end with one of our trigger words/phrases.

lib/DDG/Spice/Quandl/Fundamentals.pm
brianrisk: Ah, I didn't realize that caching was going on with the backend.

share/spice/quandl/fundamentals/fundamentals_triggers.txt
brianrisk: Unfortunately, no.  The code takes ordering into account to give preference to more specific phrases.  So a search with "current assets" will return the ASSETSC identifier, whereas simply searching with "assets" will return ASSETS.  If the list were alphabetized, the "assets" line would trigger even if people searched on "current assets".

moollaza: @brianrisk thanks for explaining.

>The code takes ordering into account to give preference to more specific phrases.

By more specific I take it you mean longer strings? I would suggest then that we alpha sort this list, and then when we slurp the file and create our trigger array, we sort the array by length.

E.g. ` @sorted = sort { length $a <=> length $b } @elements `

lib/DDG/Spice/Quandl/HomeValues.pm
brianrisk: This is a good suggestion.  The home price database only takes 5 digit zips, so I'll go with the first regex you sugggested.

lib/DDG/Spice/Quandl/Fundamentals.pm
brianrisk: Caching is disabled so that we can track calls for analytics and internal invoicing purposes.

brianrisk:
Hi @moollaza, I'm not understanding the errors produced by the latest commit:

Dubious, test returned 2 (wstat 512, 0x200)
No subtests run

moollaza:
@brianrisk if you scroll up a bit from the very bottom of the output to [line 1688](https://travis-ci.org/duckduckgo/zeroclickinfo-spice/jobs/53062872#L1688)

You'll see the output from the failed test:
```
# Failed test 'Testing query expensive homes 11235'

# at /home/travis/perl5/perlbrew/perls/5.14/lib/site_perl/5.14.4/DDG/Test/Spice.pm line 50.

# Structures begin differing at:

# $got->{call} = '/js/spice/quandl/home_values/ZIP_ALLHOMES_11235'

# $expected->{call} = '/js/spice/quandl/home_values/ZIP_TOPTIER_11235'

# Looks like you failed 1 test of 11.
```

So your test file says it should be calling TOPTIER but it's actually calling ALLHOMES. --- If the alpha sorted trigger files is causing too much trouble we can skip it for now. Apologies for the hassle!

moollaza:
BTW you can also run your tests locally with either `duckpan test` (runs all tests in repo) or `prove -Ilib t/mytestfile.t` (run from root of repo)

brianrisk:
@moollaza, don't worry:  your sorted trigger idea is very good; this will be a a feature of the Quandl IAs.  Thanks for your help decoding the test results; the "prove" command is very helpful as it is much more succinct.  The IA is now passing all tests.

To sum up the changes:  I implemented the YAML file for both fundamentals and homeValues with the triggers being the keys sorted by length.  For fundamentals, the trigger is startend to help avoid collision with searches are not seeking fundamental data.  Further, I removed a few ticker symbols and one fundamental trigger that would have produced too many erroneous results (e.g. ticker CASH).

one question: in the handle function, do we have access to a variable that is the trigger string?  If so, this would help trim the code.

moollaza:
>one question: in the handle function, do we have access to a variable that is the trigger string? If so, this would help trim the code.

@brianrisk great question! I didn't think we could easily do it, but after spending a little more time looking though our code I learned that it's actually possible to get the original query and remove the `$remainder` from it, to produce what should be our trigger.

```perl
use String::Trim;
...
handle remainder => sub {

my $remainder = $_;

my $query = $req->query;

(my $trigger = $query) =~ s/$_//ig; # copy $query into $trigger, then run a string replace 
trim($trigger); # it's likely there will be leading/trailing spaces, so trim it! 
```

share/spice/quandl/fundamentals/quandl_fundamentals.css
```diff
@@ -0,0 +1,74 @@
+/*
+    Document   : 
+    Created on : 2015-02-12 08:21 PM
+    Author     : brianrisk
+    Description: styles based on the 'Stocks' spice for visual consistency
+*/
+
+#quandl h1.quandl-header {
```
moollaza: please use `zci--quandl_fundamentals` instead of `#quandl` for namespacing the CSS. That's the canonical method. Then you can also remove the `<div id="quandl">`

share/spice/quandl/home_values/quandl_home_values.css
```diff
@@ -0,0 +1,75 @@
+/*
+    Document   : zeroclickinfo-spice/share/spice/quandl/home_values/quandl_home_values.css
+    Created on : 2015-02-17 08:25 PM
+    Author     : brianrisk
+    Description:
+    Purpose of the stylesheet follows.
+    To change this template use Tools | Templates.
+*/
+
+#quandl h1.quandl-header {
```
moollaza: Same here, `.zci--quandl_home_values` should be used

share/spice/quandl/home_values/content.handlebars
```diff
@@ -0,0 +1,20 @@
+
+<div id="quandl">
+    <h1 class="quandl-header">
```
moollaza: primary headers should have our `zci__header` class

share/spice/quandl/home_values/content.handlebars
```diff
@@ -0,0 +1,20 @@
...
+        <a href="{{url}}" title="{{urlTitle}}">
+            {{name}}
+        </a>
+    </h1>
+    <div class="quandl-date">
```
moollaza: Subheader should similarly use the `zci__subheader` class

share/spice/quandl/fundamentals/quandl_fundamentals.css
brianrisk: I can do this, but this is not conducive to code re-use.

The plan is to make more IAs for various data sources on Quandl.  These IAs will use many of the same resources such as the logo, styles and handlebars.  As resources shared between IAs are not an option at the time, the backup plan is to simply copy these files into each Quandl IA share directory.

If we have to use a different namespace for each IA (zci--quandl_fundamentals, zci-qundl_home_values, etc.), this makes the Quandl IA code more difficult to manage -- especially as we develop more IAs.

Is what you propose a necessary fix, or could we make an exception with this?

moollaza: @brianrisk understood, thanks for explaining.

@sdougbrown any thoughts on this? I don't believe the ID selector will cause problems because the div is created by the template

One alternative is to use a CSS selector like:

```csss
div[class^='zci--quandle'] {
    ...
}
```

moollaza:
@brianrisk these are looking great!

@abeyang this is ready for a design review. There are two IAs here,

1. Fundamentals
    - https://ddh7.duckduckgo.com/?q=aapl+revenue&ia=fundamentals
    - https://ddh7.duckduckgo.com/?q=aapl+earnings+per+share&ia=fundamentals
2. HomeValues
    - https://ddh7.duckduckgo.com/?q=19087+cheap+homes&ia=homevalues
    - https://ddh7.duckduckgo.com/?q=90210+homes+prices&ia=homevalue

share/spice/quandl/home_values/content.handlebars
brianrisk: Are there pre-defined styles for zci__header, or should I simply rename the .quandl-header class in my css to zci__header?  Are there DDG docs on css?

brianrisk:
Thank you @moollaza!

The last link you supplied should be:
https://ddh7.duckduckgo.com/?q=90210+home+prices&ia=homevalues

moollaza:
Whoops, fixed!

abeyang:
Hi @brianrisk , I agree with @moollaza -- they're shaping up quite nicely!

I was wondering if it's possible to make it more inline with the Stock IA. So, for **revenue**, it would look more like this:

screen shot 2015-03-06 at 10 53 23 am

And for **home values**:
screen shot 2015-03-06 at 10 54 18 am

Please note that I'm prettifying the dates -- I think @moollaza can help you with that rather easily.

LMK if you have any questions!

brianrisk:
Hi@abeyang, thanks so much for this feedback!

You and I are on the same page about the Stocks spice -- the Quandl spice IA css files are almost exact copies of the Stocks css.  This is noted in quandl_fundamentals.css.

Your suggestion to make the displayed dates more readable is very good.  I have implemented this change.  I have also added the words "up", "down", or "same": after the percent:

screen shot 2015-03-06 at 4 09 52 pm

For your suggestion to move portions of the header into a sub-header, I totally understand the desire to do this.  This makes sense for the Stocks spice as stock price data is consistently delivered on a daily schedule from one data source.  However, for a few reasons, this would not be the best route for the Quandl IAs.  The first reason is that Quandl provides a diverse array of data: stock fundamentals, diabetes rates, population data, etc.  These data have variable time reporting, so begin date and end date need to be specified.  Fundamental data could be other than quarterly (e.g. annually).  Housing data could be variable time range.  The second is that the titles are taken straight from the data source (in this case Sharadar for stock fundamentals and Zillow for house prices).  These are not consistently parseable; an attempt to do so could unknowingly produce bad output.  Passing the title straight from the data source to the displayed title preserves the integrity of the data and greatly increases code maintainability / testability.  Is keeping the titles as-is something you and your team could consider?

abeyang:
@brianrisk let me go over the titles/subtitles with @moollaza . 

In the meantime, can we still take out the "from date"? In the case of revenue, it's implicit that it's quarterly. What's more, as is for all other IAs, if users are interested in more info, they can always click on the attribution link.

brianrisk:
@abeyang: great point about the attribution link -- I've gone ahead and removed the from date.

@moollaza: thanks!  It is surprisingly not well documented how to print "pretty" dates in native JS.

lib/DDG/Spice/Quandl/Fundamentals.pm
```diff
@@ -0,0 +1,74 @@
...
+    my @words = split / /, lc $_;
+    my $wordsSize = scalar @words;
+    
+    # exit if query is less than two words
+    if ($wordsSize < 2) {return;}
+
+    # Only valid if ticker symbol is at beginning or end
+    my $ticker;
+    if (exists $tickers{$words[0]}) {$ticker = uc $words[0];}
+    else {
```
moollaza: this can actually be an `elsif` no following, `else` is required:

```perl
if (...){
    ...
} elsif (...){
    ...
}
```

moollaza:
@brianrisk apologies for the delay, it's a little disappointing to hear the titles come pre-formatted like that. Is there any way you can break down more of the data on your end so it's more flexible for us?

I understand that you'd like to build as many IA's as possible with as little effort as possible, which is what we hope to achieve for our devlopers by using templates. It seems like we'll be using this layout (and styling) more often so we'll be introducing a template which will allow you to simply specify the data for the built-in template and then you won't need to worry about creating an HTML template or writing CSS that will need to be copied. I think that should greatly reduce the complexity for these IAs, but we'd still like to ensure each of them are up to our design standards.

We won't have the templates ready for Friday, so as an interim fix, could we write a couple simple regexs to split the titles in half (at the ` - `).

If we do that I think we can leave one half for the title, and move the rest to the subtitle to achieve a slightly improved look. My biggest concern is the length of the header's and that they wrap onto a second line which makes it harder to read.

I was thinking we could go for something like this:

2015-03-10_10h08_57
2015-03-10_10h06_54

I think that would be simple enough to implement and achieves most of the look our design team was aiming for.

How does that sound @brianrisk ?

/cc @abeyang

lib/DDG/Spice/Quandl/Fundamentals.pm
brianrisk: Good point; I've committed that change

brianrisk:
@moollaza, we want a good DDG user experience, so if a couple of non-reusable code bits have to go in, that is fine.  I have implemented and pushed this change.  The solution I implemented was in the JS files.  It splits at the '-' char; the difference is that for fundamentals the first element is the header, for home values, it is the second element.

moollaza:
@brianrisk thanks a lot this is looking great!

@abeyang thoughts?

- https://ddh7.duckduckgo.com/?q=19087+2+bedroom+homes&ia=homevalues
- https://ddh7.duckduckgo.com/?q=MSFT+eps&ia=fundamentals

abeyang:
@brianrisk It's looking better! Just a few minor tweaks:

First, because the byline is so long now, can you please format the date so that it shows the **shortened month** (eg, "Jan" or "Nov" instead of January/November)?

Second, let's create a bigger separation between header and byline (we're going to apply this to the Stock IA eventually as well):

```css
#quandl h1.quandl-header {
  margin-bottom: 0.5em;
}
```

Thanks!

share/spice/quandl/fundamentals/quandl_fundamentals.css
sdougbrown: I'd prefer a simple class name to the id or the crazy regex selector.  (`.quandl` instead of `#quandl`). I really don't foresee any need for us to create a `.quandl` class internally so there's no legitimate risk of conflict. Overall I don't think it'd hurt anything and it does make sense to remove the tab-specific selectors to make the styling more reusable.

brianrisk:
@abeyang I've added the margin spacing and changed month display to abbreviated form.
@sdougbrown the Quandl css is now selected via class instead of id

moollaza:
@brianrisk @abeyang 

Updated: https://ddh7.duckduckgo.com/?q=19087+2+bedroom+homes&ia=homevalues

Doesn't look quite right though, @brianrisk I think the switch to class had an unexpected affect on the header size, it's a little big now ;)

Also, the months didn't change for some reason? Not sure why, if you switch to moment.js the formatting should be much easier:

```js
moment(api_result.to_date).format("MMM Do, YYYY");
```

brianrisk:
@moollaza css selector is now back to id from class.  The month change should be going through, it is showing up properly on the duckpan server.  Perhaps your staging servers didn't get the latest push?

moollaza:
@brianrisk @abeyang 

Updated: https://ddh7.duckduckgo.com/?q=expensive+homeS+90210&ia=homevalues

Looking good!

abeyang:
Looks good! :+1:

moollaza:
Minor adjustments here: https://github.com/duckduckgo/zeroclickinfo-spice/pull/1627

moollaza:
@brianrisk congrats! It's live: https://duckduckgo.com/?q=19087+expensive+homes&ia=homevalues

Thanks again for taking the time to contribute! We really appreciate it :+1:

brianrisk:
@moollaza So excited!  Celebrations are going on over here.  Thank you for all of your help ( @abeyang too!)

PR #1538, opened 2015-02-16
movie.js: remove the use of is_retina because it's being deprecated.

jagtalon:
We're moving to `DDG.is3x` and `DDG.is2x` now.


moollaza:
tested. LGTM :+1:

bsstoner:
nice, thank you!

Issue #1542, opened 2015-02-17
RXInfo: Relevancy


yegg:
Bad relevancy here -- is this going through the is_relevant function?

https://duckduckgo.com/?q=tablet+of+ahkmenrah&t=iphone&iax=1&ia=rxinfo

https://duck.co/ia/view/rx_info

MrChrisW:
@yegg [rx_info.js] (https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/rx_info/rx_info.js) isn't using `DDG.isRelevant` I'll see if it can be improved.

moollaza:
Fixed in #1550

PR #1550, opened 2015-02-18
RxInfo: Update to use isRelevant to avoid returning invaild results

MrChrisW:
Fixing #1542 

Currently the RxInfo Spice triggers on small English two letter/three letter words: `aa ad ae am an as at be by do go ha he hi id if in is it me no of ok on or pi so to up us we`. 

Examples:
https://duckduckgo.com/?q=tablet+of+blue+color
https://duckduckgo.com/?q=tablet+is
https://duckduckgo.com/?q=capsule+age

Use DDG.isRelevant and call Spice.failed if no relevant results are found

### DIFF
--- a/lib/DDG/Spice/RxInfo.pm
+++ b/lib/DDG/Spice/RxInfo.pm
@@ -17,8 +17,13 @@ triggers startend => 'pill', 'rxinfo', 'capsule', 'tablet', 'softgel', 'caplets'
 spice to => 'http://rximage.nlm.nih.gov/api/rximage/1/rxbase?resolution=300&includeIngredients=true&parse=$1';
 spice wrap_jsonp_callback => 1;
 
+my $triggerWords = join "|", share('triggerWords.txt')->slurp(chomp => 1);
+
 handle remainder => sub {
-    return $_ if $_;
+    my $attributes = 0;
+    return unless $_;   
+    $attributes = 1 if (/$triggerWords/);
+    return $_,$attributes;
     return;
 };
 1;
--- a/lib/DDG/Spice/RxInfo.pm
+++ b/lib/DDG/Spice/RxInfo.pm
@@ -15,15 +15,14 @@ attribution web => ['http://www.medicosconsultants.com', 'Medicos Consultants, L
 triggers startend => 'pill', 'rxinfo', 'capsule', 'tablet', 'softgel', 'caplets';
 
 spice to => 'http://rximage.nlm.nih.gov/api/rximage/1/rxbase?resolution=300&includeIngredients=true&parse=$1';
+spice from => '(.*?)/(\d)';
 spice wrap_jsonp_callback => 1;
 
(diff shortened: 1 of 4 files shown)


moollaza:
@MrChrisW good catch! The goal of this IA was to be able to search for a pill by description though, rather than the name, so queries like "round blue pill" or "oval pill" should be triggering this. Checking the name for relevancy would likely cause none of the results to show in these cases.

I think we can continue with your relevancy checking (it should be put into a relevancy block though :smile: ) and we'll need to add another type of relevancy check that ensures some basic info is present.

The first type of relevancy checking should check the relevancy of the drug name or NDC number against the query. The second should be used so that when a color/shape is given we ensure that a picture and description exists for the result.

Make sense?

https://duck.co/duckduckhack/spice_displaying#relevancy-blocks

MrChrisW:
@moollaza Sure. I've reviewed the original PR and understand the need for this to trigger on ambiguous searches like "yellow pill" as the API support this. 

As you said, the problem is these types of queries will not function if we add a relevancy block checking the name & NDC.

A fix would be to have a list of all shapes/colours, if the query contains any we skip the "name" relevancy check & simply ensure an image & description exists. 

I'd prefer to implement the list of shapes/colours in Perl, then simply return an argument to the front-end if the query contains any, if the argument is detected we skip relevancy checking in JS. Sound ok? 

I've noticed there's actually quite a few Spices doing relevancy checking the way I originally implemented in this PR. loop + array + isRelevant(); these should probably be updated, if using relevancy blocks is best practise (I'll get right on it :laughing: ) 

Thanks :+1:

moollaza:
>I'd prefer to implement the list of shapes/colours in Perl, then simply return an argument to the front-end if the query contains any, if the argument is detected we skip relevancy checking in JS. Sound ok? 

@MrChrisW sounds good! 

>I've noticed there's actually quite a few Spices doing relevancy checking the way I originally implemented in this PR. loop + array + isRelevant(); these should probably be updated, if using relevancy blocks is best practise (I'll get right on it :laughing: ) 

@MrChrisW sadly you're right, there's lots of opportunities for improvement, but please don't feel like you need to do all this work. I often try to leave some "low-hanging fruit" type tasks that aren't exactly important (the relevancy checking is there so results are OK, but the implementation could be improved) for newcomers so that they have something to do. If no one tackles it after a while then we can decide to do it ourselves, but again it's not really hurting other than potentially setting a bad example for others to follow (ie. they didn't use a relevancy block, I won't either)

MrChrisW:
@moollaza This should be right to review :+1:

moollaza:
LGTM :+1:

moollaza:
@MrChrisW I think we can still add a little more query checking to the Perl to verify the query meets our expected format. Potentially we should just remove the trigger word "tablet" as it's very popular now with regards to electronics.

MrChrisW:
@moollaza :+1: Sure, open to any suggestions to improve relevancy.

PR #1551, opened 2015-02-19
Remove unused local vars in amazon spice js

bsstoner:
The `loadedRatingsData` var isn't used anymore, and the `items` local var isn't necessary since it's only referenced in one place now.

@moollaza @jagtalon


jagtalon:
Looks good! :+1:

PR #1558, opened 2015-02-21
BBC: Handle Radio 4 schedule (Fix #1557)

### LINKED ISSUE #1557: BBC: Radio 4 schedule not working
A search for "what's on bbc radio 2" lists programmes, but "what's on bbc radio 4" does not.

BBC Radio 4 has two variations on its schedule, i.e.:

http://www.bbc.co.uk/radio4/programmes/schedules/fm/today.json
http://www.bbc.co.uk/radio4/programmes/schedules/lw/today.json

I *think* this is a special case.  Probably best to default to FM, but allow requests for LW explicitly.

diocles:
Radio 4 comes in two flavours (FM and LW), which look very much
like areas in the Programmes API.  Default to FM.


moollaza:
LGTM :+1:  @diocles Thanks for taking the time to contribute! We really appreciate it. 

We invite you to stick around and comment/help on other PR's & Issues or even submit an Instant Answer!

PR #1560, opened 2015-02-21
Update Template: Example.pm

MrChrisW:
+ Caching is enabled by default - remove `spice is_cached => 1;` 
+ Include `spice to`

@moollaza


moollaza:
@MrChrisW I can't believe I was missing the `spice to` param!

Let's leave stick with only a `$1` because if they have a `$2` they'd also need a `spice from` which arguably we could also put in.

The reason why we chose to put in the `spice is_cached` was because most devs weren't aware that there is caching or what that caching param was doing so the hope was that they'd see that and consider if they do/don't need caching and which kind of caching it is that's needed.

MrChrisW:
@moollaza Sure, no problem!  A Spice doesn't work very well without the `spice to` param :laughing:

moollaza:
LGTM :+1: 

Thanks a ton @malbin

PR #1566, opened 2015-02-24
Use provider name for Env APIKEY variable

moollaza:
The canonical naming convention is to reference the provider name, not the Spice name. We're using this APIKEY in other Instant Answers so it makes sense to reference the Provider and not the IA name.

Note: There is a corresponding change to the close-source code

/cc @jagtalon


jagtalon:
:+1:

PR #1567, opened 2015-02-24
HOTFIX: CourseBuffet: API endpoint is down because it's using HTTPS

jagtalon:
HTTP endpoint not working anymore. HTTPS is working so we're moving there.

I had to change the image endpoint as well so that it wouldn't break:

screen shot 2015-02-24 at 2 38 44 pm

screen shot 2015-02-24 at 2 42 13 pm


moollaza:
Tested and LGTM :+1:

jagtalon:
Thank you

PR #1568, opened 2015-02-24
BBC: Use header_service_type for name field

diocles:
This was an already-declared variable set to either 'TV' or 'Radio'
depending on which type of station the user asked for.


moollaza:
@diocles good catch!

moollaza:
Thanks again!

PR #1570, opened 2015-02-24
Travis: Use container-based platform

diocles:
See: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

It's faster!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 ---
+sudo: false
 before_install:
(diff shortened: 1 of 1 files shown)


moollaza:
@diocles thanks a lot -- we actually attempted this previosly but we need to wait because it creates problems as the .travis.yml file in the root of the repo gets overwritten at build time by the TravisCI Dist::Zilla Plugin.

See https://github.com/duckduckgo/zeroclickinfo-spice/pull/1456/files

PR #1574, opened 2015-02-27
JS Keycodes

andrey-p:
Thought a JS keycodes IA is something that's needed. There's a [goodie](https://github.com/duckduckgo/zeroclickinfo-goodies/issues/414) that does a similar thing but it hasn't been released yet. (Further discussion between me and @mintsoft [here](https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1013#issuecomment-75055319).)

Displays Javascript keycodes.

It saves a few clicks for a very common (for front-end devs) search topic.

The existing Goodie and [this page](www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes), which is the first result for "js keycodes" (and I've been using it for *years*).

There's the [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) on this too.

"keycode tab", "javascript keycode j", "javascript keycodes"

Javascript developers.

Nope.

The [JsKeycodes](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/JsKeycodes.pm) goodie

Turns out there's a few weird cross-browser / cross-platform (possibly even cross-keyboard and cross-locale?) differences - they're the ones highlighted in the massive MDN table linked above. I decided to go for the data that was already there, but I was wondering whether it'll be a good idea to add a disclaimer mentioning there could be some differences.

None.

2015-02-27-150657_1366x768_scrot
2015-02-27-151102_1366x768_scrot
2015-02-27-151114_1366x768_scrot

```
[X] Tested cross-browser compatibility (see note above though)

- Linux
      [X] Chromium
      [X] Firefox   

- Windows XP
      [X] IE 8
```

------

> IA Page: https://duck.co/ia/view/js_key_codes

### DIFF
--- a/lib/DDG/Spice/JsKeycodes.pm
+++ b/lib/DDG/Spice/JsKeycodes.pm
@@ -8,7 +8,7 @@ spice is_cached => 1;
 
 # Metadata
 name "Javascript Keycodes";
-source "www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes";
+source "MDN";
 description "Get the Javascript keycode for a key";
 primary_example_queries "keycode j", "javascript keycode tab";
 secondary_example_queries "javascript keycodes";
--- a/lib/DDG/Spice/JsKeycodes.pm
+++ b/lib/DDG/Spice/JsKeycodes.pm
@@ -10,7 +10,7 @@ spice is_cached => 1;
 
 # Metadata
 name "Javascript Keycodes";
-source "";
+source "www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes";
 description "Get the Javascript keycode for a key";
 primary_example_queries "keycode j", "javascript keycode tab";
 secondary_example_queries "javascript keycodes";
--- a/lib/DDG/Spice/JsKeycodes.pm
+++ b/lib/DDG/Spice/JsKeycodes.pm
@@ -4,8 +4,6 @@ package DDG::Spice::JsKeycodes;
 use DDG::Spice;
 use Text::Trim;
 
-use Data::Dumper;
-
 spice is_cached => 1;
 
 # Metadata
@@ -21,114 +19,20 @@ code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/
 attribution github => ["https://github.com/andrey-p/", "Andrey Pissantchev"];
 
 # Triggers
-triggers any => "keycode",
-    "keycodes";
+triggers any => "js keycode",
+    "javascript keycode",
+    "js keycodes",
+    "javascript keycodes",
+    "js key code",
+    "javascript key code",
+    "js key codes",
+    "javascript key codes";
+
 
 spice call_type => 'self';
 
-my %keys = map { $_ => undef } (
-    'backspace',
-    'tab',
-    'enter',
-    'shift',
-    'ctrl',
-    'alt',
-    'pause',
-    'break',
-    'caps lock',
-    'escape',
-    'page up',
-    'page down',
-    'end',
-    'home',
-    'left arrow',
-    'up arrow',
-    'right arrow',
-    'down arrow',
-    'insert',
-    'delete',
-    '0',
-    '1',
-    '2',
-    '3',
-    '4',
-    '5',
-    '6',
-    '7',
-    '8',
-    '9',
-    'a',
-    'b',
-    'c',
-    'd',
-    'e',
-    'f',
-    'g',
-    'h',
-    'i',
-    'j',
-    'k',
-    'l',
-    'm',
-    'n',
-    'o',
-    'p',
-    'q',
-    'r',
-    's',
-    't',
-    'u',
-    'v',
-    'w',
-    'x',
-    'y',
-    'z',
-    'space',
-    'numpad 0',
-    'numpad 1',
-    'numpad 3',
-    'numpad 4',
-    'numpad 5',
-    'numpad 6',
-    'numpad 7',
-    'numpad 8',
-    'numpad 9',
-    '*',
-    '-',
-    '.',
-    'f1',
-    'f2',
-    'f3',
-    'f4',
-    'f5',
-    'f6',
-    'f7',
-    'f8',
-    'f9',
-    'f10',
-    'f11',
-    'f12',
-    'num lock',
-    'scroll lock',
-    ';',
-    ,
-    ',',
-    '/',
-    '\\',
-    '(',
-    ')',
-    'quote'
-);
-
-my $is_valid_key;
-
 # Handle statement
 handle remainder_lc => sub {
-    # display only if query contains "javascript" or "js"
-    # or the remainder is a valid key
-    # to avoid conflict with any other uses of the word "keycode"
-    return unless $_ =~ s/js|javascript//i or exists $keys{$_};
-
     return call;
 };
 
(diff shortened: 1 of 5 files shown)


lib/DDG/Spice/JsKeycodes.pm
```diff
@@ -0,0 +1,135 @@
...
+);
+
+my $is_valid_key;
+
+# Handle statement
+handle remainder_lc => sub {
+    # display only if query contains "javascript" or "js"
+    # or the remainder is a valid key
+    # to avoid conflict with any other uses of the word "keycode"
+    return unless $_ =~ s/js|javascript//i or exists $keys{$_};
```
MrChrisW: To avoid calling this Spice on any search for `keycode` can the `js`,`javascript` triggers be added to `triggers` instead of the guard. 

This will also allow you to return the keycode and avoid using `.replace` on [L58](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1574/files#diff-235b2e5ea069d75a9003d5ce023271e9R58) in js_keycodes.js by grabbing the returned result using [source.match](https://duck.co/duckduckhack/spice_advanced_frontend#accessing-the-query-remainder) - the docs you wrote!

andrey-p: > To avoid calling this Spice on any search for keycode can the js,javascript triggers be added to triggers instead of the guard.

@MrChrisW My approach was to trigger on searches like `keycode [valid keyname]` (but not searches like `keycode pro`) - do you think it'll be more useful to allow only searches with `js` and `javascript` in the query?

> This will also allow you to return the keycode and avoid using .replace on L58 in js_keycodes.js by grabbing the returned result using source.match - the docs you wrote!

Yeah, I was pretty excited about using that as well but I couldn't get it to work for this spice - I think having `call_type => 'self'` means there's no separate JSONP call to the server.

MrChrisW: @andrey-p Bugger! That's right, unfortunately there's nothing passed to the front end when using `'self'`

Regarding the triggering, I think it's probably a good idea to only trigger when js or javascript is contained in the query (feel free to disagree! :smile: ) my reasoning behind this is because these keycodes are Javascript specific, they're not [Ascii decimal](http://www.asciitable.com/) keycodes for example. 

You can also improve the triggering by assigning a list of triggers that you've generated with `map` to `triggers startend =>` this will avoid calling the `handle` for useless queries (`keycode pro`) and thus improve overall performance. 

You can find a good example of this in the [RandPOS](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/RandPOS.pm#L22-L31) IA!

mintsoft: > Regarding the triggering, I think it's probably a good idea to only trigger when js or javascript is contained in the query (feel free to disagree! :smile: ) my reasoning behind this is because these keycodes are Javascript specific, they're not Ascii decimal keycodes for example.
I concur with this; it's better to be more precise and expand later than too broad and give users irrelevant information.

andrey-p: > my reasoning behind this is because these keycodes are Javascript specific, they're not Ascii decimal keycodes for example. 

Ah, that's an excellent point. I'll change it so that it always checks for `js` or `javascript` in the query.

> You can also improve the triggering by assigning a list of triggers that you've generated with map to triggers startend => this will avoid calling the handle for useless queries (keycode pro) and thus improve overall performance. 

That's a good trick to keep in mind, thanks. Might not end up using it as I might remove the list of key names from the backend - will have a play around with it and see.

moollaza:
@andrey-p this is looking pretty good -- I noticed a few keys didn't register around my numpad, but that may be my keyboard?

However the `-` and `=` keys beside my backspace didn't seem to register either?

@abeyang any design thoughts? My biggest concern is the mobile layout because the table is large and wide

share/spice/js_keycodes/content.handlebars
```diff
@@ -0,0 +1,25 @@
+{{#if result}}
+<p class="keycode-result">The keycode for <code>{{result.key}}</code> is: {{result.code}}</p>
+{{else}}
+<p class="keycode-result">Press a key in the textbox to check its keycode:</p>
+<input type="text" class="keycode-input"/>
+{{/if}}
+<a class="btn btn--primary show-keycodes" href="#">Show all codes</a>
+
+{{! hidden when the page first renders }}
+<table class="all-keycodes">
```
moollaza: Might be best to add the class "is-hidden" here so that it doesn't need to wait for the Spice's CSS to load in order to hide it, or you could use inline css but the former is better :smiley:

share/spice/js_keycodes/js_keycodes.js
```diff
@@ -0,0 +1,151 @@
...
+        Spice.add({
+            id: "js_keycodes",
+            name: "Keycodes",
+            data: data,
+            meta: {},
+            templates: {
+                group: 'base',
+                options: {
+                    content: Spice.js_keycodes.content
+                }
```
moollaza: I would suggest you put the JS below into an `onShow` [event](https://duck.co/duckduckhack/spice_displaying#onshow-codefunctioncode)

andrey-p:
@moollaza Added changes to do with hiding the table with `is-hidden` and moving stuff to `onShow`.

> I noticed a few keys didn't register around my numpad, but that may be my keyboard?
>
> However the - and = keys beside my backspace didn't seem to register either?

Yeah, that's the thing I'm not 100% what to do about... Turns out there's some minor but very annoying differences between the way different keyboards, browsers and OSes handle keycodes. For instance [this page on MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) lists about 3 different keys that can have a keycode of 186!

I was thinking of adding a link to the MDN article and a disclaimer that some key combinations are ambiguous, maybe marking the offending entries in the table in some way. What do you think?

moollaza:
@andrey-p I think the `More at` should probably link to MDN -- a disclaimer would be good too!

jagtalon:
@moollaza looks like it's ready for review!

lib/DDG/Spice/JsKeycodes.pm
```diff
@@ -0,0 +1,39 @@
...
+    "javascript key code",
+    "js key codes",
+    "javascript key codes";
+
+
+spice call_type => 'self';
+
+# Handle statement
+handle remainder_lc => sub {
+    return call;
```
moollaza: We should check here that if there's a remainder it's relevant. I.e. just a single word -- if we could make sure it's exactly an expected keycode that would be awesome. We could build a hash (perhaps in a separate text file in /share/) that contains all the names of the keys and if the remainder is one of those, or non-existant this AI can return, otherwise it should ignore the query.

We typically use YAML and load the strings into a hash for a fast lookup. This is a good example: https://github.com/duckduckgo/zeroclickinfo-spice/pull/1737/files

share/spice/js_keycodes/js_keycodes.js
```diff
@@ -0,0 +1,155 @@
...
+
+                keyName = keyCode.key;
+
+                $input.val(keyName);
+                $result.html("The keycode for <code>" + keyName + "</code> is: " + code);
+            });
+        }
+
+    };
```
jagtalon: Please call `ddg_spice_js_keycodes` at the end like https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/flash_version/flash_version.js#L34. :)

gautamkrishnar:
:+1:

jagtalon:
Talked to @andrey-p and @moollaza and we agreed that this would be better as a cheat sheet (albeit a modified cheat sheet that has the same features as this one). Closing this in favor of that. :neckbeard:

PR #1580, opened 2015-02-28
Remove MetaYAML Plugin, already using newer [MetaJSON]

moollaza:
/cc @jagtalon


jagtalon:
@moollaza nice--so you mean META.yml isn't actually being used?

moollaza:
@jagtalon it seems that META.json is meant to replace the META.yml file/format

jagtalon:
:+1:

PR #1582, opened 2015-03-01
BBC: link to schedule

diocles:
At the moment with the BBC results, the "More at" link goes to the BBC homepage.  This PR makes that link go to the relevant schedule page.


moollaza:
@diocles this is excellent! Thanks a lot, we really appreciate it and I hope you'll keep up the great work!

PR #1587, opened 2015-03-03
Rewrite of Indeed Jobs IA. Fixes #1584

### LINKED ISSUE #1584: Indeed (Jobs): Rescue and update PR
Before the DDG redesign a PR was started for Job search using the Indeed.com API.

We'd love to see this go live and we're asking if anyone out there would be interested in bringing that PR back to life? The code will need to be re-written as it was created before the changes to the Spice API and before our templates so the implementation should be **much** simpler now using the `text` template group.

Any takers out there?!

Here's the original PR:
https://github.com/duckduckgo/zeroclickinfo-spice/pull/333

tagawa:
Gets 10 jobs posts from indeed.com

Shows location, company and how recent the job posting is.

https://ads.indeed.com/jobroll/xmlfeed

Suggested by @moollaza: https://duck.co/ideas/idea/604/jobs-listings-from-indeed-com

GitHub jobs and other job sites, but I believe indeed.com aggregates a lot of these.

"web developer jobs"
"jobs in boston"
"tax accountant job in san francisco"

Adults looking for a job

Yes: https://duck.co/ideas/idea/604/jobs-listings-from-indeed-com
and fixes #1584 

Probably supersedes github_jobs

1. An Indeed Jobs publisher ID is required (DDG_SPICE_INDEED_JOBS_PUBLISHER). Does DDG have one?
2. In Firefox I'm unable to click the tiles to go to indeed.com. Is it just me? Is this a DDG template problem? I've not got any custom code so this is really confusing me.

The API is free but an API key (publisher ID) is required, as well as attribution.

Contributed IAs a couple of years ago.

ddg_spice_indeed_jobs

## Checklist

```

    - Ubuntu 14
        [X] Chromium
        [X] Firefox *
        [X] Opera
        [X] Midori (WebKit)

    - Android
        [X] Firefox *
        [X] Native Browser
        [] Google Chrome
        [X] Opera
* Note: On Firefox the display is OK but unable to click the tiles. See comment above.
```

------
> IA Page: https://duck.co/ia/view/indeed_jobs

### DIFF
--- a/lib/DDG/Spice/IndeedJobs.pm
+++ b/lib/DDG/Spice/IndeedJobs.pm
@@ -1,0 +1,39 @@
+package DDG::Spice::IndeedJobs;
+# ABSTRACT: Search for jobs on Indeed.
+# Reviews DDG_SPICE_INDEED_JOBS_PUBLISHER environment variable
+# https://ads.indeed.com/jobroll/xmlfeed
+
+use DDG::Spice;
+
+primary_example_queries "java jobs";
+secondary_example_queries "perl jobs in san francisco";
+description "Indeed jobs";
+name "Indeed Jobs";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/IndeedJobs.pm";
+topics "special_interest";
+category  "reference";
+attribution github => ['https://github.com/parker','parker'],
+            twitter => ['http://twitter.com/pseidel','pseidel'];
+
+triggers any => "job", "jobs";
+
+spice to => 'http://api.indeed.com/ads/apisearch?publisher={{ENV{DDG_SPICE_INDEED_JOBS_PUBLISHER}}}&v=2&useragent=DuckDuckGo&userip=1.2.3.4&q=$1&l=$2&co=$3&format=json&callback={{callback}}';
+spice from => '([^/]+)/(.*?)/([^/]*)';
+spice proxy_cache_valid   => "418 1d";
+
+
+handle query => sub {
+    my $country = 'US';
+    if($loc->country_code) {
+        $country = $loc->country_code;
+    }
+    my $spice_location = $loc->city . ', ' . $loc->region_name;
+    if (/(?:\s*(?:i\s+|we\s+)?(?:need|want|deserve|seek|get)\s+(?:an?\s+)?)?(?:(.+)\s+)?(?:jobs?|work|employment|internship)(?:\s+(?:in|near\s+)?\s*(.+))?$/i) {
+        my $query = $1 || ' ';
+        my $location = $2 || $spice_location || ' ';
+        return $query, $location, $country;
+    }
+	return;
+};
+
+1;
--- a/lib/DDG/Spice/IndeedJobs.pm
+++ b/lib/DDG/Spice/IndeedJobs.pm
@@ -33,7 +33,6 @@ handle query => sub {
     
     if (/(?:\s*(?:i\s+|we\s+)?(?:need|want|wanna|deserve|seek|get|find)\s+(?:a\s+|an?\s+)?)?(?:(?<query>.+)\s+)?(?:jobs?|work|employment|internship)(?:\s+(?:in|near|around\s+)?\s*(?<location>.+))?$/i) {
         if ($1 || $2) {
-            #my $query = $1 || ' ';
             my $query = $+{query} || ' ';
             my $location = $+{location} || $spice_location || ' ';
             return $query, $location, $country;
--- a/lib/DDG/Spice/IndeedJobs.pm
+++ b/lib/DDG/Spice/IndeedJobs.pm
@@ -20,7 +20,7 @@ attribution github  => ['parker', 'Parker'],
 
 triggers any => "job", "jobs";
 
-spice to => 'http://api.indeed.com/ads/apisearch?publisher={{ENV{DDG_SPICE_INDEED_JOBS_PUBLISHER}}}&v=2&useragent=DuckDuckGo&userip=1.2.3.4&q=$1&l=$2&co=$3&format=json&callback={{callback}}';
+spice to => 'http://api.indeed.com/ads/apisearch?publisher={{ENV{DDG_SPICE_INDEED_APIKEY}}}&v=2&useragent=DuckDuckGo&userip=1.2.3.4&q=$1&l=$2&co=$3&format=json&callback={{callback}}';
 spice from => '([^/]+)/(.*?)/([^/]*)';
 spice proxy_cache_valid   => "418 1d";
 
--- a/lib/DDG/Spice/IndeedJobs.pm
+++ b/lib/DDG/Spice/IndeedJobs.pm
@@ -12,8 +12,11 @@ name "Indeed Jobs";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/IndeedJobs.pm";
 topics "special_interest";
 category  "reference";
-attribution github => ['https://github.com/parker','parker'],
-            twitter => ['http://twitter.com/pseidel','pseidel'];
+attribution github  => ['parker', 'Parker'],
+            twitter => ['pseidel', 'Parker'],
+            github  => ['tagawa', 'Daniel Davis'],
+            twitter => ['ourmaninjapan', 'Daniel Davis'],
+            web     => ['http://daniemon.com/', 'Daniel Davis'];
 
 triggers any => "job", "jobs";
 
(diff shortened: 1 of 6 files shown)


pfirsichbluete:
Travis failed the rainfall tests (due to an unrelated change). A commit to fix those failures is at https://github.com/duckduckgo/zeroclickinfo-spice/pull/1589

zekiel:
thanks a million @pfirsichbluete

moollaza:
@tagawa apologies for the long delay! This is looking great. I have a couple suggestions though:

- Let's set the `item.snippet` as the `description` and pass the location to the footer template
- Let's use the new `variants` to space things out accordingly:
    
    ```js
    normalize: function(item) {
        return {
            url: item.url,
            title: item.jobtitle,
            subtitle: item.company,
            description: DDG.strip_html(item.snippet)
        };
    },
    templates: {
        group: 'text',
        detail: false,
        item_detail: false,
        options: {
            footer: Spice.indeed_jobs.footer
        },
        variants: {
            tileTitle: '2line',
            tileSnippet: 'large',
            tileFooter: '2line'
        }
    }
    ```

    ```html
    <div class="one-line">
        {{formattedLocationFull}}
    </div>
    <span class="tile__time one-line text--secondary">
        {{formattedRelativeTime}}
    </span>
    ```

The result looks like this:

2015-03-24_18h03_37

/cc @abeyang @chrismorast any thoughts?

tagawa:
Thanks for the suggestion. I've made this change:
* Full description show in view
* Location shown in footer
* Uses new variants 

I also changed the IA display name to "Indeed Jobs" rather than just "Jobs" to avoid duplication with GitHub Jobs (which will also need to be changed).

moollaza:
> I also changed the IA display name to "Indeed Jobs" rather than just "Jobs" to avoid duplication with GitHub Jobs (which will also need to be changed).

Actually both should just have "Jobs" as their `name`. We're trying to avoid using any brand names for our tabs (i.e. Videos vs YouTube, Software vs GitHub).

lib/DDG/Spice/IndeedJobs.pm
```diff
@@ -0,0 +1,43 @@
...
+category  "reference";
+attribution github  => ['parker', 'Parker'],
+            twitter => ['pseidel', 'Parker'],
+            github  => ['tagawa', 'Daniel Davis'],
+            twitter => ['ourmaninjapan', 'Daniel Davis'],
+            web     => ['http://daniemon.com/', 'Daniel Davis'];
+
+triggers any => "job", "jobs";
+
+spice to => 'http://api.indeed.com/ads/apisearch?publisher={{ENV{DDG_SPICE_INDEED_JOBS_PUBLISHER}}}&v=2&useragent=DuckDuckGo&userip=1.2.3.4&q=$1&l=$2&co=$3&format=json&callback={{callback}}';
```
moollaza: One more small fix, lets change the APIKEY placeholder to our canonical wording: `DDG_SPICE_INDEED_APIKEY`

tagawa:
Hi again @moollaza. Appreciate the clarification - I've made those changes.
The name policy means there would be two tabs with the name "Jobs" so maybe the GitHub one could be changed to "IT Jobs" (or similar).

moollaza:
>The name policy means there would be two tabs with the name "Jobs" so maybe the GitHub one could be changed to "IT Jobs" (or similar).

You're correct, our policy is to prevent duplicate Instant Answers like this scenario so we're going to reduce the trigger space of the GitHub Jobs IA by changing the trigger to "github jobs"

lib/DDG/Spice/IndeedJobs.pm
```diff
@@ -0,0 +1,43 @@
...
+attribution github  => ['parker', 'Parker'],
+            twitter => ['pseidel', 'Parker'],
+            github  => ['tagawa', 'Daniel Davis'],
+            twitter => ['ourmaninjapan', 'Daniel Davis'],
+            web     => ['http://daniemon.com/', 'Daniel Davis'];
+
+triggers any => "job", "jobs";
+
+spice to => 'http://api.indeed.com/ads/apisearch?publisher={{ENV{DDG_SPICE_INDEED_APIKEY}}}&v=2&useragent=DuckDuckGo&userip=1.2.3.4&q=$1&l=$2&co=$3&format=json&callback={{callback}}';
+spice from => '([^/]+)/(.*?)/([^/]*)';
```
moollaza: @tagawa it's best to put optional checks at the end and only return them if they're defined otherwise you'll end up with a double slash `//` which gets automatically changed to `/` by NGINX.

```perl
spice from => '([^/]+)/([^/]+)/?([^/]+)?';
```

lib/DDG/Spice/IndeedJobs.pm
```diff
@@ -0,0 +1,43 @@
...
+spice proxy_cache_valid   => "418 1d";
+
+handle query => sub {
+    my $country = 'US';
+    if($loc->country_code) {
+        $country = $loc->country_code;
+    }
+    my $spice_location = $loc->city . ', ' . $loc->region_name;
+    if (/(?:\s*(?:i\s+|we\s+)?(?:need|want|wanna|deserve|seek|get|find)\s+(?:a\s+|an?\s+)?)?(?:(.+)\s+)?(?:jobs?|work|employment|internship)(?:\s+(?:in|near|around\s+)?\s*(.+))?$/i) {
```
moollaza: Might help to use named captures here: `(?:a\s+|an?\s+)?)?(?:(?<query>.+)\s+)?`, `my $query = $+{query}`

lib/DDG/Spice/IndeedJobs.pm
moollaza: Sorry this doesn't actually apply here -- you're returning `' '` not `''` so it's OK

abeyang:
Hi @tagawa . This IA looks good! There are some things we'll need to modify on our end (the template itself), but as far as yours stand, it's good! :+1:

lib/DDG/Spice/IndeedJobs.pm
```diff
@@ -0,0 +1,43 @@
...
+        if ($1 || $2) {
+            my $query = $1 || ' ';
+            my $location = $2 || $spice_location || ' ';
+            return $query, $location, $country;
```
moollaza: I think we either need to handle the country a little bitter, or ignore it for now. if I try to search for job in a city outside my country, I'm going to get really weird results or none. e.g. "jobs in philadelphia" sets the country to `CA` because I'm in Canada, which sends some mixed signals to Indeed.

We should either check for the presence of a country and get the country code using `Locale::Country` (a few other Spice do this) or we should try and pass the whole location along to Indeed (I'm not sure how well they handle that?)

moollaza:
Up on DDH8, https://ddh8.duckduckgo.com/?q=computer+programmer+jobs+in+philadelphia

/cc @abeyang

tagawa:
Thanks @abeyang and @moollaza &mdash; pleased to hear it's looking good with just a couple of things to tweak.

lib/DDG/Spice/IndeedJobs.pm
tagawa: Looks like the Indeed API is only for the US, e.g. for "Paris" it defaults to Paris, Texas and there are no results for Paris, France. I agree we should ignore ```$country``` completely but should we also restrict the IA to searches from the US only?

moollaza: I believe it supports all the countries they have websites for. It definitely had results for Canada when the country was set to CA, and I just tried Brazil and India and they have results there too. I'd like to keep it open to all users because I can imagine lots of people outside the US look for jobs there.

I think the easiest route is to mimic the Time Spice which uses a large list of the biggest cities for most countries. https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Time.pm#L23

That should allow us to get great coverage for the majority of queries. If needed, we can add internal "deep" triggering.

abeyang:
Just curious @tagawa --

1. Are we able to show the most recent job listings first?
2. Does Indeed return different job types (eg, full-time, part-time, contract, etc)?

tagawa:
@abeyang:
1. Yes, the IA now shows most recent jobs first.
2. Yes, it's possible to filter by full-time, part-time, contract, temporary and internship. However, I'm having trouble adding more placeholders to the Spice To URL. Is there a max limit of 3?

lib/DDG/Spice/IndeedJobs.pm
tagawa: Still wondering about the best way to handle countries. You're right that they support more than just the US but they can't handle the whole location without country being a separate parameter.

I think most practical would be a hash lookup of capitals/countries, and perhaps a separate check for the country name in the query. This still wouldn't catch your Philadelphia (Canada) example but it would be better than nothing.

moollaza:
> However, I'm having trouble adding more placeholders to the Spice To URL. Is there a max limit of 3?

No, there's no limit. But any missing/empty params will be collapsed (i.e. `spice/my_spice/$1//$3` -> `spice/my_spice/$1/$3`) which will throw off your regex. You'll need to return a ` ` for empty params or join params into a single string (`return $foo, "$bar-$baz"` -> `/foo/bar-baz/`) and then your regex will need to capture what exists.

FYI: We're bringing front-end filtering to the UI very soon, which will allow users to filter the original returned results, and also re-query the API with added filters

tagawa:
Thanks @moollaza. The value of ```$4``` is not being recognized so I'll try out your suggestions.
Front-end filtering sounds very cool &mdash; looking forward to it.

moollaza:
Everything LGTM :+1:

abeyang:
I agree with @moollaza.  Just a head's up, we are in the process of updating the look+feel for these kind of jobs IAs, so this might get yet another update in the near future. Thanks, @tagawa !

tagawa:
Thank you both. Job-related templates sounds good so I'll look out for those.

lib/DDG/Spice/IndeedJobs.pm
moollaza: >I think most practical would be a hash lookup of capitals/countries, and perhaps a separate check for the country name in the query. 

@tagawa I just realized we didn't implement this yet -- that approach sounds good to me!

jagtalon:
@tagawa @moollaza now live! https://duckduckgo.com/?q=javascript+jobs+nyc&ia=jobs

screen shot 2015-05-12 at 12 35 41 pm

tagawa:
Looks great! Thanks for the heads-up @jagtalon.

PR #1596, opened 2015-03-06
Remove all trigger words except for "weather"

tommytommytommy:
The weather IA is overtriggering and producing some odd results when we pass Forecast an extracted location that it does not fail on. To try to address this, I would like to limit the keyword triggering for this IA to only "weather," and then handle all other potential weather/forecast-related queries with an internal trigger that looks at search results to infer the user's intent.

### DIFF
--- a/lib/DDG/Spice/Forecast.pm
+++ b/lib/DDG/Spice/Forecast.pm
@@ -12,21 +12,6 @@ secondary_example_queries "weather 12180";
 topics "everyday", "travel";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Forecast.pm";
 
-# my @temperatures = qw(temp temperature);
-# my @connectors   = qw(in for at);
-# my @descriptors  = qw(current);
-# 
-# my @temps_triggers;
-# 
-# foreach my $temp (@temperatures) {
-#     foreach my $conn (@connectors) {
-#         push @temps_triggers, $temp . ' ' . $conn;    # temperature for, temp at
-#         foreach my $desc (@descriptors) {
-#             push @temps_triggers, join(' ', $desc, $temp, $conn), $desc . ' ' . $temp;    # current temperature in; current temp
-#         }
-#     }
-# }
-
 my @forecast_words = qw(weather);
 my @triggers = (@forecast_words);
 triggers startend => @triggers;
--- a/lib/DDG/Spice/Forecast.pm
+++ b/lib/DDG/Spice/Forecast.pm
@@ -12,23 +12,23 @@ secondary_example_queries "weather 12180";
 topics "everyday", "travel";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Forecast.pm";
 
-my @temperatures = qw(temp temperature);
-my @connectors   = qw(in for at);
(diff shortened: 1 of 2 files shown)


MrChrisW:
Hey @tommytommytommy ! Better update the [tests](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/t/Forecast.t#L82) as well :+1:

tommytommytommy:
@MrChrisW Thanks for the heads-up! I commented out the tests and triggers in the IA instead of deleting them because they might be good examples for others that are exploring what they can do with triggers in their own Spice. If it's preferable, I am happy to delete all of the commented out sections to clean up the code.

lib/DDG/Spice/Forecast.pm
```diff
@@ -12,28 +12,30 @@ secondary_example_queries "weather 12180";
...
+
+my @forecast_words = qw(weather);
+my @triggers = (@forecast_words);
 triggers startend => @triggers;
 
 spice from => '([^/]*)/?([^/]*)';
 spice to => 'http://forecast.io/ddg?apikey={{ENV{DDG_SPICE_FORECAST_APIKEY}}}&q=$1&callback={{callback}}';
 
+# skip synonyms for "weather" because of overtriggering
+my @weather_synonyms =qw(forecast);
```
moollaza: hehe maybe we should change the name of the IA too, then?

lib/DDG/Spice/Forecast.pm
```diff
@@ -12,28 +12,30 @@ secondary_example_queries "weather 12180";
...
-        push @temps_triggers, $temp . ' ' . $conn;    # temperature for, temp at
-        foreach my $desc (@descriptors) {
-            push @temps_triggers, join(' ', $desc, $temp, $conn), $desc . ' ' . $temp;    # current temperature in; current temp
-        }
-    }
-}
-
-my @forecast_words = qw(forecast forcast weather);
-my @triggers = (@forecast_words, @temps_triggers, 'meteo');
+# my @temperatures = qw(temp temperature);
```
moollaza: Might as well remove this, we can always revert the commit to get it back.

t/Forecast.t
```diff
@@ -12,24 +12,24 @@ my $loc = test_location('de');
...
-    DDG::Request->new(
-        query_raw => 'forecast',
-        location => $loc
-    ) => test_spice(
-        "/js/spice/forecast/" . uri_escape_utf8(${\$loc->loc_str}) . "/current",
-        call_type => 'include',
-        caller => 'DDG::Spice::Forecast',
-        is_cached => 0
-    ),
+    #     DDG::Request->new(
```
moollaza: Same here, let's just remove this to simplify things. The docs and (hopefully) all other IA's should be providing good examples.

lib/DDG/Spice/Forecast.pm
tommytommytommy: Haha...well, since we're still using proper noun Forecast as the provider, let's leave it for now :)

tommytommytommy:
@moollaza Comments removed!

lib/DDG/Spice/Forecast.pm
moollaza: :+1:

moollaza:
@tommytommytommy LGTM :+1:

Issue #1581, opened 2015-03-01
Keybase: Should provide info when no hit is found.


ScreapDK:
~~The [Keybase IA](https://duck.co/ia/view/keybase) when trying to get key based on GitHub username.~~
~~E.g. a search for [keybase github:mwolny](https://duckduckgo.com/?q=keybase+github%3Amwolny&ia=keybase) doesn't trigger the Instant Answers.~~

~~At the same [keybase github:jeresig](https://duckduckgo.com/?q=keybase+github%3Ajeresig&ia=keybase).~~

~~Switching between space and colon doesn't make any difference.~~

~~Instant Answer details:~~
~~[Keybase IA](http://duck.co/ia/view/keybase)~~

...the above is what I've been write for the last 5 minutes... but, I have discovered that what I thought was a bug, is actually just an example of not giving useful feedback upon error.

The reason "keybase github:mwolny" doesn't work, while "keybase github:jeresig" does is that the GitHub user `mwolny` is not connected to any Keybase.io user. The DDG SERP just doesn't tell me.

E.g. if the search [keybase github:ScreapDK](https://duckduckgo.com/?q=keybase+github%3AScreapDK&ia=keybase) doesn't work because GitHub user ScreapDK is not on Keybase, then the Instant Answer should report something like:
"Sorry, @ScreapDK on **GitHub** does not appear to be on Keybase."

MrChrisW:
@ScreapDK Usually if the answer doesn't return a result the default is to not display the IA.

Although this is a great case to update the [example queries](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Keybase.pm#L19-L20), we shouldn't include queries in `
secondary_example_queries` that don't trigger. 

I'd also like to update the `primary_example_queries` to include a better result for the first query, as currently the primary query doesn't really show off the IA as it excludes the *description, location and a nice profile image*! I think a better query would make this look nicer https://duck.co/ia/view/keybase

moollaza:
@ScreapDK as @MrChrisW said, generally if we don't have something good to show, we don't show anything at all. It's arguable that given how specific the query and triggers are for this IA that we could, but then you get into difficult territory when people search things that aren't meant to be a keybase user search. E.g. "keybase website", "keybase faq", "keybase financial"

ScreapDK:
Sorry for this late reply! 

[...]

PR #1597, opened 2015-03-06
Keybase Hotfix: Check returned status

MrChrisW:
The Keybase IA attempts to check if an object (`api_result.them`) returned by the keybase API is an array however if this doesn't exist it throws an exception as we're checking an undefined object.

This is currently in production: https://duckduckgo.com/?q=keybase+moollaza
selection_258

I've also updated the example_queries as discussed in #1581


jagtalon:
@MrChrisW oof. Thanks. Making a quick test and merging.

jagtalon:
:+1: good catch @MrChrisW

PR #1601, opened 2015-03-06
Update regex for amazon.jp review urls

b2ddg:
The regex for Amazon reviews needs to include '.jp' to match results from amazon.jp


moollaza:
LGTM :+1:

moollaza:
actually, @b2ddg it looks like amazon.jp redirects to .co.jp which we have. So is this still necessary?

Issue #1554, opened 2015-02-19
Switch custom date formatting to moment.js


jagtalon:
We recently added support for http://momentjs.com/, and it would be cool if we move things like https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/hacker_news/hacker_news.js#L63.

![screen shot 2015-02-19 at 1 31 09 pm](https://cloud.githubusercontent.com/assets/81969/6273242/b514424c-b83b-11e4-8485-d0f4baa9c2c1.png)

If possible, of course. This might be a huge task that needs a bunch of testing so it's best to do one PR for each IA so that it's easier to review.

Usage of moment.js in DuckDuckGo is as follows:

``` javascript
DDG.require("moment.js", function() {
    console.log(moment("9/23/2015").format("dddd, MMMM Do YYYY")); // Wednesday, September 23rd 2015
});
```

mattr555:
YES!
![gif](http://www.reactiongifs.com/wp-content/gallery/yes/dwight_yes.gif)

There are many "fuzzy date" functions in the codebase that should be replaced. Some more that I found (in no particular order):

+ [airlines](https://github.com/duckduckgo/zeroclickinfo-spice/blob/73b4091f04695ce104b94c0bfe38e0e47ee86dc9/share/spice/airlines/airlines.js#L145)
+ [flights](https://github.com/duckduckgo/zeroclickinfo-spice/blob/73b4091f04695ce104b94c0bfe38e0e47ee86dc9/share/spice/flights/route/flights_route.js#L309)
+ [tvmaze](https://github.com/duckduckgo/zeroclickinfo-spice/blob/dc7bddb4f42b0d52cafcfa6e2cb8b176decd5ad0/share/spice/tvmaze/nextepisode/tvmaze_nextepisode.js#L96)
+ [launchbug](https://github.com/duckduckgo/zeroclickinfo-spice/blob/4e83783722e32b27adeb66072c1780f59d3610ed/share/spice/launchbug/launchbug.js#L9)
+ [time](https://github.com/duckduckgo/zeroclickinfo-spice/blob/837c18804528b96784fb2d7397d781a2749ded0e/share/spice/time/time.js#L33)
+ [github_jobs](https://github.com/duckduckgo/zeroclickinfo-spice/blob/6502229e8f2126df93168104650c9cf81cc7968b/share/spice/github_jobs/github_jobs.js#L40)
[...]

PR #1605, opened 2015-03-07
RedditSearch: switch to moment.js

mattr555:
For #1554


moollaza:
LGTM :+1: 

It actually fixed some dates that I see are way off in production.

PR #1606, opened 2015-03-07
Whois: Support for moment.js

MrChrisW:
:warning: There's currently an issue with using the existing function to parse dates resulting in `undefined` being returned, see [live example](https://duckduckgo.com/?q=whois+see.se&ia=whois) :warning: 

**What this does:** See #1554

+ Update prettifyTimestamp function to support moment.js date parseing
+ Wrap in require moment.js function

**Screenshots**

### Current
selection_264

________________
### Updated
selection_265

### DIFF
--- a/share/spice/whois/whois.js
+++ b/share/spice/whois/whois.js
@@ -92,9 +92,9 @@
             'title': api_result.domainName,
             'Registered to': get_first_by_key(contacts, 'name'),
             'Email': get_first_by_key(contacts, 'email'),           
-            'Last Updated': api_result.updatedDate,
+            'Last Updated': prettifyTimestamp(api_result.updatedDate),
             'Expires On': prettifyTimestamp(api_result.expiresDate),
-            'Registrar': prettifyTimestamp(api_result.registrarName),
+            'Registrar': api_result.registrarName,
             'Name Servers': nameServers
         };
 
--- a/share/spice/whois/whois.js
+++ b/share/spice/whois/whois.js
@@ -6,21 +6,23 @@
         return Spice.failed('whois');
     }
 
-    // get the search query
-    var script = $('[src*="/js/spice/whois/"]')[0],
-        source = $(script).attr("src"),
-        query = source.replace('/js/spice/whois/','')
+    DDG.require('moment.js', function(){       
+        // get the search query
+        var script = $('[src*="/js/spice/whois/"]')[0],
+            source = $(script).attr("src"),
+            query = source.replace('/js/spice/whois/','')
 
-    // all the data is stored in WhoisRecord
-    api_result = api_result.WhoisRecord;
+        // all the data is stored in WhoisRecord
+        api_result = api_result.WhoisRecord;
 
-    // fail if the domain name the api returns does not match the searched domain
-    if(api_result.domainName != query) {
-        return Spice.failed('whois');
-    }
-    
-    // decide which template to show show_available or show_whois
-    (is_domain_available(api_result)) ? show_available(api_result) : show_whois(api_result);  
+        // fail if the domain name the api returns does not match the searched domain
+        if(api_result.domainName != query) {
+            return Spice.failed('whois');
+        }
+        
+        // decide which template to show show_available or show_whois
+        (is_domain_available(api_result)) ? show_available(api_result) : show_whois(api_result);
+    });  
 }
 
  // show message saying that the domain is available.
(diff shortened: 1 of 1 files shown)


share/spice/whois/whois.js
```diff
@@ -113,16 +115,25 @@
...
         if(!timestamp) { return; }
-        var dateObj = DDG.getDateFromString(timestamp),
-            monthArr = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"],
-            day = dateObj.getDate(),
-            month = monthArr[dateObj.getMonth()+1],
-            year = dateObj.getFullYear();
-        if(day<10) {day='0'+day};
-        return month + " " + day + ", " + year;
+
+        //When using the expiresDateNormalized (example: 2020-09-14 00:00:00 UTC); need to use custom date formatting
```
moollaza: Can you please explain why that is? I don't understand the need/problem that this code solves.

share/spice/whois/whois.js
```diff
@@ -113,16 +115,25 @@
...
+        //(if !api_result.updatedDate && !api_result.expiresDate) 
+
+        var customDate = moment(timestamp, "YYYY-MM-DD HH:mm Z"), // formatting date: 2020-09-14 00:00:00 UTC
+            autoDate = moment(timestamp); // using auto parse
+
+        //Check which date is vaild
+        if(autoDate.isValid()) {
+            return autoDate.format("MMM MM, YYYY");
```
moollaza: Should `MM` be `DD`?

share/spice/whois/whois.js
```diff
@@ -113,16 +115,25 @@
...
-            day = dateObj.getDate(),
-            month = monthArr[dateObj.getMonth()+1],
-            year = dateObj.getFullYear();
-        if(day<10) {day='0'+day};
-        return month + " " + day + ", " + year;
+
+        //When using the expiresDateNormalized (example: 2020-09-14 00:00:00 UTC); need to use custom date formatting
+        //(if !api_result.updatedDate && !api_result.expiresDate) 
+
+        var customDate = moment(timestamp, "YYYY-MM-DD HH:mm Z"), // formatting date: 2020-09-14 00:00:00 UTC
```
moollaza: I think the comment has an extra `:00` ?

moollaza: Small nit, but it seems like we're doing extra work in creating the `customDate` before we know if we'll actually need it? Let's create this in the `else` to fix that.

moollaza:
@MrChrisW looking good! Just need some clarification on the updated `prettifyTimestamp`

share/spice/whois/whois.js
MrChrisW: @moollaza Sure, was awaiting this question :wink: 

Looking at the following [conditional block](https://github.com/MrChrisW/zeroclickinfo-spice/blob/whois-updates/share/spice/whois/whois.js#L76-L81) you can see we failback to using the `DateNormalized` values if we're unable to find `updatedDate` or `expiresDate` values.
 
moment.js cannot automatically parse the normalized dates so we need to provide a custom format `YYYY-MM-DD HH:mm:ss Z` [line 125](https://github.com/MrChrisW/zeroclickinfo-spice/blob/whois-updates/share/spice/whois/whois.js#L125)

MrChrisW:
@moollaza Sure. No problem. Another issue I've noticed with the Whois IA is the `updatedDateNormalized` & `expiresDateNormalized` can contain different **date values** then the dates contained in `updatedDate` & `expiresDate`, this can result in the IA displaying differing dates/incorrect dates.

**Possible be an API issue**. [API documentation](https://www.whoisxmlapi.com/documentation/whoisapi_documentation/noNamespace/element/WhoisRecord.registryData.html) .

Example (duckduckgo.com whois) - `updatedDate`: https://gist.github.com/MrChrisW/b550074caa9f8de98e5e#file-gistfile1-txt-L3

Example (duckduckgo.com whois) - `updatedDateNormalized`: https://gist.github.com/MrChrisW/b550074caa9f8de98e5e#file-gistfile1-txt-L110

share/spice/whois/whois.js
MrChrisW: Yes

MrChrisW:
@moollaza Updated, removed function `prettifyTimestamp` logic now [happens here](https://github.com/MrChrisW/zeroclickinfo-spice/blob/whois-updates/share/spice/whois/whois.js#L80-L88)

moollaza:
@MrChrisW this is looking really good. One small bug though:

2015-03-17_00h55_47

I'm not quite sure what the cause is...

MrChrisW:
@moollaza No problem, i'll investigate shortly. I'm lodging an issue with the API provider regarding the difference in dates mentioned.

MrChrisW:
@moollaza Ok, I've looked at the data returned by the API and it's now returning inconsistant date formats (updatedDate, createdDate) for some queries (a little frustrating). 

I think my previous ilitiration of checking if the date is vaild then proceeding to parse it would work, I'll test against a few domains to ensure we get consistant results. Will update shortly :+1:

moollaza:
>it's now returning inconsistant date formats (updatedDate, createdDate) for some queries (a little frustrating). 

Argh. That's really annoying.

>I think my previous ilitiration of checking if the date is vaild then proceeding to parse it would work, I'll test against a few domains to ensure we get consistant results. Will update shortly :+1:

Awesome! Ping me when ready :smile:

MrChrisW:
@moollaza Ping pong :point_left: [Updated](http://duckpan2.mrchris.me/?q=whois+twitter.com&ia=whois)

share/spice/whois/whois.js
```diff
@@ -66,8 +73,10 @@
         }
 
         // check we have 2 nameserver hostnames
-        if(nsObj && nsObj.hostNames[0] && nsObj.hostNames[1]) {
-            var nameServers = [nsObj.hostNames[0].toLowerCase(),nsObj.hostNames[1].toLowerCase()].join(' ');
+        if(nsObj.hostNames) {
+            if(nsObj.hostNames.length >= 2) {
+                var nameServers = [nsObj.hostNames[0].toLowerCase(),nsObj.hostNames[1].toLowerCase()].join(' ');
```
jagtalon: Small nitpick that doesn't affect anything. This looks like it was already defined above.

jagtalon:
@MrChrisW looks like the issues above have been addressed. I have style issues but I can just put that in the Issues. :) :+1:

share/spice/whois/whois.js
MrChrisW: Stop nitpicking @jagtalon :laughing: Unfortunately if you try checking if  `nsObj.hostNames[0-1]` exists and it doesn't you'll end up with an `undefined` typeError so we first check the length. :+1:

moollaza: @MrChrisW FWIW when grabbing deep elements of objects that may be undefined it might help to use our [DDG.getProperty](https://duck.co/duckduckhack/spice_js_api#getpropertyobj-pathname) function. It will catch the undefined and return null

MrChrisW: @moollaza Thanks for the reference :+1: 
I was actually testing out the use of a try catch block for the above check, in an act of desperation!! I'll make use of the `getProperty` method next time :laughing:

PR #1617, opened 2015-03-09
Npm: Include more Info (fixes #1530)

### LINKED ISSUE #1530: NPM: Show more important information about the package!
When I search something like `npm phantomjs` I get this

screen shot 2015-02-13 at 4 50 00 pm

But I'm not sure if that's the best one partly because I can't see how popular it is. So I end up clicking on the link anyway.

killerfish:
fixes #1530 
@jagtalon @moollaza I think this is the most useful information we can have here, where do you guys think dependencies should be shown? i included both places, also let me know what other info you think should be there.
SS below 
npmupdate

### DIFF
--- a/share/spice/npm/npm.js
+++ b/share/spice/npm/npm.js
@@ -14,11 +14,34 @@
                 sourceName: "npmjs",
                 sourceUrl: 'http://npmjs.org/package/' + api_result.name
             },
-            
             normalize: function(item) {
+                var boxData = [{heading: 'Package Information:'}];
+                
+                boxData.push({
+                    label: "Author: " + item.author.name
+                });
+ 
+                boxData.push({
+                    label: "Project Homepage: " + item.homepage
+                });
+
+                boxData.push({
+                    label: "Repository: " + item.repository.url
+                });
+
+                var dependencies = $.map(item.dependencies, function(val, key) {
+                    return key + " (" + val + ")";
+                }).join(", ");
+
+                boxData.push({
+                    label: "Dependencies: " + dependencies
+                });
+
                 return {
                     title: item.name + " " + item.version,
-                    subtitle: item.description
+                    subtitle: item.description,
+                    infoboxData: boxData,
+                    dependency: dependencies
                 }  
             },
 
--- a/share/spice/npm/npm.js
+++ b/share/spice/npm/npm.js
@@ -17,31 +17,56 @@
             normalize: function(item) {
                 var boxData = [{heading: 'Package Information:'}];
                 
-                boxData.push({
-                    label: "Author: " + item.author.name
-                });
+                if (item.author) {
+                    boxData.push({
+                        label: "Author",
+                        value: item.author.name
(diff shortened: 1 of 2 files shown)


share/spice/npm/npm.js
```diff
@@ -14,11 +14,34 @@
                 sourceName: "npmjs",
                 sourceUrl: 'http://npmjs.org/package/' + api_result.name
             },
-            
             normalize: function(item) {
+                var boxData = [{heading: 'Package Information:'}];
+                
+                boxData.push({
+                    label: "Author: " + item.author.name
```
moollaza: These should actually use `label` and also `value` to get the intended look:

```js
{
    label: "Author",
    value: item.author.name
}
```

share/spice/npm/npm.js
```diff
@@ -14,11 +14,34 @@
...
+                });
+ 
+                boxData.push({
+                    label: "Project Homepage: " + item.homepage
```
moollaza: These ones can use the `url` field, to generate a clickable link:

```js
{
    label: "Author",
    url: item.author.name
}
```

share/spice/npm/npm.js
```diff
@@ -14,11 +14,34 @@
...
+                boxData.push({
+                    label: "Repository: " + item.repository.url
+                });
+
+                var dependencies = $.map(item.dependencies, function(val, key) {
+                    return key + " (" + val + ")";
+                }).join(", ");
+
+                boxData.push({
+                    label: "Dependencies: " + dependencies
```
moollaza: This one can actually provide a hash for `value`:

```js
{
    label: "Dependencies",
    value: {
        content: {
            label: //optional
            value: 
        }
}
```

moollaza:
@killerfish thanks for doing this! There's a lot of undocumented functionality for the infobox that I've highlighted. (I swear we're working on updating the docs!)

The use of the infobox is definitely great, nicely done. I'm not a huge fan of the source button though. I think the repo and project homepage links are sufficient.

The only other things I think we should add to the infobox would be the license type and maybe the supported node engines.

I think we can also remove the dependencies from the body, or try to render them with the `record` template so they're easier to see/read.

killerfish:
This seems good, i think, dependencies with record view might crowd the body, so they suit better in infobox. Hmm, i think source button might serve its purpose for some people (since we wont be always getting homepage and repo links).  
SS below
npmupdate

moollaza:
@abeyang any thoughts on the button? I'll put this up on a dev machine soon..

moollaza:
@killerfish I spoke with the design team. I didn't realize at first that the button is to download the source. I'm not sure that's really all that necessary so let's move the link to the infobox. We should definitely add an infobox link to the `homepage` though. I think that would be most useful.

killerfish:
`homepage` is included, though some packages dont have them. Updated design, SS below
npmup

moollaza:
LGTM :+1:

PR #1619, opened 2015-03-10
Octopart: Switch from api v2 to v3 (fixes #1595)

### LINKED ISSUE #1595: Octopart: Migrate to API v3 -- API v2 deprecating March 20!
Octopart will be officially turning off Octopart API v2 on **March 20, 2015** and **will not be functional** after March 20. 

Let's update it fast!

Here's the documentation for v3: https://octopart.com/api/docs/v3/overview

Code: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Octopart.pm#L18

https://duck.co/ia/view/octopart

killerfish:
fixes #1595 
@moollaza let me know if the switch to v3 slows down spice calls, ill see if it could be optimized.

### DIFF
--- a/share/spice/octopart/octopart.js
+++ b/share/spice/octopart/octopart.js
@@ -47,7 +47,7 @@ function ddg_spice_octopart (api_result) {
         meta: {
             itemType: 'Parts',
             sourceName: 'Octopart',
-			sourceUrl : 'http://octopart.com/partsearch#search/requestData&q=' + api_result.request.q
+            sourceUrl : 'http://octopart.com/search?q=' + api_result.request.q
         },
         normalize: function(item) {
             var img = DDG.getProperty(item, "item.imagesets.0.small_image.url");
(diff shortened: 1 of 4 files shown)


share/spice/octopart/octopart.js
```diff
@@ -50,18 +50,18 @@ function ddg_spice_octopart (api_result) {
 			sourceUrl : 'http://octopart.com/partsearch#search/requestData&q=' + api_result.request.q
```
moollaza: This is actually currently broken, can you please fix?

Should be: `http://octopart.com/search?q=`

killerfish: done

jagtalon:
Nice! Gotta get this in before the 20th!

moollaza:
@killerfish thanks for getting this going! I gave it a try and unfortunately it wasn't working -- seems like some of the params you were using no longer exist.

I noticed the code code be simplified and cleaned up a bit and pushed to a new PR #1651 -- would you be interested in continuing that? I noticed that the detail pane isn't showing so something still needs to be fixed...

moollaza:
Actually i think I manged to just about finish it all up...

moollaza:
@killerfish apologies -- I was actually not using API v3 when testing this which is why it seemed broken. @jagtalon pointed that out to me while testing my PR. I've changed my code to use what you had before and I've added some other updates as well.

PR #1620, opened 2015-03-10
Use DDG.toHTTP/S where applicable (fixes #1552)

### LINKED ISSUE #1552: Replacing https to http should just use DDG.toHTTP
Lines such as https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/sound_cloud/sound_cloud.js#L55 should be replaced with `DDG.toHTTP`. Might exist in other spices as well.

killerfish:
fixes #1552


moollaza:
LGTM :+1:

PR #1622, opened 2015-03-10
nutrition relevancy

### LINKED ISSUE #1616: Nutrition: Regex and relevancy need to be improved
Right now it looks like anything with one of our trigger words is causing the IA to try and return our results. I suspect that our regex is letting too many things through given all the optional matching groups. I haven't taken a close look but this needs to be fixed!

For example, any query with "iron" in it currently triggers this IA:

- https://duckduckgo.com/?q=iron+man&ia=nutrition 
- https://duckduckgo.com/?q=ram+iron+scrap+metal&ia=nutrition

The regex should likely solve most of the problems but we should also try adding a relevancy block if it still over triggers.

jdorweiler:
Removed some of the ambiguous trigger words so things like `carbon fiber wrap` don't trigger.
Internal pr will take care of the rest of the relevancy. 
fixes #1616


jdorweiler:
https://ddh10.duckduckgo.com/?q=carbon+fiber+wrap&ia=nutrition

Some examples:
https://ddh10.duckduckgo.com/?q=fat+in+kale&ia=nutrition
https://ddh10.duckduckgo.com/?q=fiber+in+popcorn&ia=nutrition

moollaza:
@jdorweiler this is looking pretty good -- but it seems that it still triggers when any one of our trigger words exists in the query: https://ddh10.duckduckgo.com/?q=protein+shake&ia=nutrition

I noticed that our `$question_regex` actually has a `?` after every clause, so it's not enforcing anything (not sure if that was by design though)

jdorweiler:
@moollaza I think that was by design so things like `protein chicken` worked.

moollaza:
@jdorweiler maybe we should mimic the Forecast spice and remove all the triggers, except 1 or 2 like "nutrition" and let Deep do the rest?

jdorweiler:
@moollaza good idea :+1: 

https://ddh10.duckduckgo.com/?q=protein+shake&ia=about

russellholt:
LGTM, merging now to reduce over triggering.

PR #1627, opened 2015-03-11
Add zci__header, zci__subheader and text-primary classes for better colours

moollaza:
Added in our DDG text classes. Darkens header and subheader colours slightly:

2015-03-11_19h49_34

- https://ddh7.duckduckgo.com/?q=aapl+revenue&ia=fundamentals
- https://ddh7.duckduckgo.com/?q=19087+3+bedroom+home&ia=homevalues

//cc @brianrisk @abeyang

### DIFF
--- a/share/spice/quandl/fundamentals/content.handlebars
+++ b/share/spice/quandl/fundamentals/content.handlebars
@@ -1,18 +1,16 @@
-
 <div id="quandl">
-    <h1 class="quandl-header">
+    <h1 class="quandl-header zci__header">
         <a href="{{url}}" title="{{urlTitle}}">
             {{header}}
         </a>
     </h1>
-    <div class="quandl-date">
+    <div class="quandl-date zci__subheader">
         <span>{{subheader}} as of {{to_date}}</span>
     </div>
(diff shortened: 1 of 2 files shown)


abeyang:
Looks good! :+1:

brianrisk:
@moollaza a subtle, but good change!  I just did a test on the production site (as I'm eagerly checking for when the code is live).  My search on "aapl revenue" returned the Stock spice IA.  Will the Quandl IAs supersede these results?

moollaza:
@brianrisk right now priority is given to the IA that returns first. We have the option of giving IA's higher priority which we can do if Quandl isn't getting the spotlight when it should be.

Now that the IA is merged I'll set it up for deployment and I'm aiming to get it deployed by tonight.

moollaza:
@brianrisk to clarify however it's behaving right now on DDH7 is how it'll act in production. You'll notice that when you select the Quandl Spice the URL changes to specify that, so in a pinch you can provide a link to a search that'll auto show the given IA based on the `&ia=` param, assuming the IA actually triggers.

PR #1630, opened 2015-03-12
Parse Amazon Review Count

sdougbrown:
Currently being passed to the template as a formatted string, this should remove commas and allow the template to abbreviate the review counts for extremely high numbers.

to @bsstoner


sdougbrown:
We're thinking of going about this a slightly different way.

PR #1634, opened 2015-03-13
BrainyQuote: Cut at 150 chars (fixes #805)

### LINKED ISSUE #805: Brainy Quotes: ellipsis cuts off too much of the quote
for example 'john kennedy quote' gives

> As we express our gratitude, we must never forget that the highest appreciation is... 

...is what? what is it?!

https://duck.co/ia/view/brainy_quote

killerfish:
fixes #805 
Cutting off at 200, was breaking the design, so i set this at 150 chars.

### DIFF
--- a/share/spice/brainy_quote/content.handlebars
+++ b/share/spice/brainy_quote/content.handlebars
@@ -1,6 +1,6 @@
 <div class="brainy__quote--container">
   <div class="brainy__quote">
-    {{ellipsis q 85}}
+    {{ellipsis q 150}}
   </div>
   <div class="brainy__attribution">
     {{person}}


moollaza:
@killerfish thanks for this! Please post a screenshot though, for PR's like this that update the design/look.

killerfish:
I do that, usually, just that in this case, it doesnt really update the design as such, just accommodates longer quote texts.

moollaza:
> it doesnt really update the design as such, just accommodates longer quote texts.

True, but I'm eager to see the before/after effect of it, and a pic would help me (and others) quickly understand what the change does. I'll likely end up testing it in DuckPAN anyways to make sure, but being able to see things immediately is always nice :smiley:

killerfish:
quote

moollaza:
@killerfish thanks for the pic. This LGTM, but I wasn't actually able to test it in DuckPAN because of UTF8 quote character in the CSS. Did you not run into that problem?

killerfish:
Nope, works fine for me.

moollaza:
LGTM :+1:

PR #1632, opened 2015-03-12
Reviving old branch - Ideas/wikinews

iambibhas:
I took one of the branches mentioned in #1535 and worked on it a little. Updated the frontend to latest format and added handlebar template. This is how it looks - 

wikinews

I'm looking for feedback on if what I did was right, if not what to do to improve. Then I can work on the other branches.

Issue #1535, opened 2015-02-16
Revive Old IA Branches


moollaza:
https://github.com/duckduckgo/zeroclickinfo-spice/branches/stale

Previously we have pushed the "abandoned" PRs to a branch for others to work off of and they've been sitting there for a while now untouched.

Most of the branchs [here](https://github.com/duckduckgo/zeroclickinfo-spice/branches/stale) have Spice IAs that are nearly complete, but in many cases need to be updated to meet the current Spcie practices (using built-in Templates and the new Spice API)

If you're interested in picking up some of these IA's feel free to let us know and submit a PR :smile:

tagawa:
Re. the exoplanets IA, the original API (exoapi.com) seems dead but NASA have a public API that could probably be used: http://exoplanetarchive.ipac.caltech.edu/docs/program_interfaces.html

iambibhas:
I looked at the [ideas/firefox-plugins](https://github.com/duckduckgo/zeroclickinfo-spice/tree/ideas/firefox-plugins) branch. addons.mozilla.org API now returns XML as mentioned [here](https://developer.mozilla.org/en-US/docs/addons.mozilla.org_%28AMO%29_API_Developers%27_Guide). They probably changed it since the last time that branch was used.

moollaza:
@iambibhas it looks like that branch was going to use XML (which isn't allowed), but I just tried adding `?format=json` to an example URL in their docs and it worked! :smiley: 

https://services.addons.mozilla.org/en-US/firefox/api/1.5/addon/1865?format=json

iambibhas:
Cool! Not sure why it isn't documented. So I Can work on it. Lemme give it a try.

iambibhas:
Looks like JSON response isn't available for the search endpoint.
```
{
    "msg": "Not implemented yet."
}
```

https://services.addons.mozilla.org/en-US/firefox/api/1.5/search/memory/all/1?format=json

iambibhas:
This old IA(stale branch `upc`) specifies the API Key in plain text. That should be removed, right? https://github.com/duckduckgo/zeroclickinfo-spice/blob/upc/lib/DDG/Spice/Upc.pm#L10

moollaza:
@iambibhas that's not one of DDG's keys, it's the contributor's, but yes if you plan to bring back that IA, you should get a new key. 

The contributor isn't using that key though for anything else. He was a DDG intern :)

[...]

PR #1642, opened 2015-03-16
Reviving old branch - Ideas/wikinews

iambibhas:
I took one of the branches mentioned in #1535 and worked on it a little. Updated the frontend to latest format and added handlebar template. This is how it looks - 

wikinews_at_duckduckgo_-_2015-03-16_22_17_31

This is the new PR against master branch. The old one #1632 got a bit out of hand because I opened it against the stale `ideas/wikinews` branch.


share/spice/wikinews/content.handlebars
```diff
@@ -0,0 +1,6 @@
+<h5>Latest Articles from Wikinews</h5>
```
moollaza: Looks like we can remove this file?

jagtalon:
@moollaza looks like @iambibhas made his changes.

moollaza:
@iambibhas I can't remember what we discussed, but just wanted to double check -- is there any chance we can get summaries/snippets or photos for this Spice?

iambibhas:
@moollaza no. :| the api just provides page id, title and timestamp.

moollaza:
@abeyang I think this is ready for a design pass. It's really unfortunate, but this is about all the data we can get from Wiki News...

It just dawned on me that perhaps the list view would be better for this IA?

abeyang:
I think this is a decent foundation, but agree with @moollaza that it would be nice if they can open up their API a bit more (this is bare bones at the moment).

With that said, @iambibhas let's make the tiles thinner and a little bit taller. First, let's modify the css (prefixed with your own namespace, of course):

``` css
.tile--c, .tile--m {
  width: 14em;
  height: 11em;
}
```

And instead of using the class `tile__title--3sm` for the titles, let's use these classes: 
`tile__title--4 tx--17`

In the end, it should look something like this:

screen shot 2015-04-28 at 4 11 08 pm

Thanks!

iambibhas:
@abeyang I added that css change. Also, I'm currently using `tileTitle: "3line-small"` for title. What do I need to do to add that `tile__title--4 tx--17` class instead? Here is how it looks now -

moollaza:
@iambibhas you'll want to use the new `elClass` property to pass along the `tx--17` and `variants` to change the`tileTitle` to `4line`

Details: https://duck.co/duckduckhack/spice_displaying#directly-specifying-classes

iambibhas:
@moollaza doing this is adding the `tx--17` class but not the `tile__title--4` -

```
    templates: {
        group: 'text',
        options: {
            footer: Spice.wikinews.footer
        },
        detail: false,
        item_detail: false,
        variants: {
            tileTitle: "4line",
        },
        elClass: {
            tileTitle: 'tx--17'
        },
    }
```

iambibhas:
The example for `4line` in the [doc](https://duck.co/duckduckhack/spice_templates_overview) is actually using `3line-small` - https://duckduckgo.com/?q=live+shows+in+london&ia=concerts

moollaza:
@iambibhas good catch, sorry about that

@abeyang it looks like we removed the `4line` title internally -- was that expected?

moollaza:
@iambibhas alright, we've re-implemented the `tileTile: 4line` variant -- this should be working again!

iambibhas:
@moollaza it isn't working. :( I ran `duckpan -e -v server` to clear the cache but it's still not appending the `tile__title--4` class.

moollaza:
@iambibhas sorry I spoke to soon, it hasn't been pushed live yet. Should be going on soon though...

moollaza:
LGTM :+1: 

@iambibhas `4line` is now working. I'll add it in with a new PR.

Issue #1591, opened 2015-03-04
RedditSubSearch: Switch to trigger words


moollaza:
Seeing as the regex it uses looks for `/r/` or `r/` (or `subreddit`) we should be able to use a word trigger because the "words" of a query are not only split on spaces, but also slashes, dashes, periods, and other punctuation.

As a result, `triggers startend => "subbredit", "r"` followed by a `handle` that checks the query against our current trigger regex should work in exactly the same way, but it's much more efficient because we're not using a regex trigger.

https://duck.co/ia/view/reddit_sub_search

puskin94:
@moollaza I only have one question: How can I check the triggers without a regex if a word trigger has not spaces?
`r/ pizza` and `/r/ pizza` works because there is a space between the trigger and the query, but `r/pizza` cannot be triggered because there is no space and the trigger is not recognized. Am I right? I'm playing with this IA but I can't write a working solution ;)

moollaza:
@puskin94 hmm it looks like there's possibly a bug in the DuckDuckGo library. The triggering does work as expected, which you can see if you turn on the Block Trace debug: `DDG_BLOCK_TRACE=1 duckpan query`

You'll have to change the IA to simply `return $_` in the handle because it doesn't have a `remainder` (which is the bug and why you likely didn't get a result).

If you switch it to `handle query_lc` and run the original regex in the `handle` to ensure we have a properly formatted query, this should work as expected.

jagtalon:
Fixed by @puskin94

PR #1643, opened 2015-03-16
Reddit: Switch from regex to word trigger

puskin94:
@moollaza I changed the trigger system about this issue: https://github.com/duckduckgo/zeroclickinfo-spice/issues/1591 . Now it works properly but I don't know if this is the best solution

### DIFF
--- a/lib/DDG/Spice/RedditSubSearch.pm
+++ b/lib/DDG/Spice/RedditSubSearch.pm
@@ -29,10 +29,7 @@ handle query_lc => sub {
 	return $_ if $_;
 	return unless $_ =~ qr#^(?:subreddit|/?r/?)\s*(\w+)$|^(\w+)\s+subreddit$#i;
 
-	if ($1) {
-		return $1;
-	}
-	return $2;
+	($1) ? return $1 : return $2;
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/DDG/Spice/RedditSubSearch.pm
```diff
@@ -16,16 +16,21 @@ attribution twitter => ["https://twitter.com/mithrandiragain","Gary Herreman"],
...
-handle matches => sub {
-	if($1) {
-		return $1;
-	} elsif($2) {
-		return $2;
-	}
-    return;
+handle query_lc => sub {
+
+	$_ =~ s/^(subreddit|r)|(subreddit|r)$//i;
```
moollaza: Not sure what you're trying to achieve with these replacements. The regex below does all the work for you!

```perl
handle query_lc => sub {
    return unless m/<regex>/);
    my $remainder = $1 || $2;
    return $remainder;
}
```
Something like that should be fine. The regex could possibly do with some improvements, or rather a named capture so we don't have to check for both match groups.

puskin94:
@moollaza I don't know why but I have the power to solve problems with complex solutions, especially if the problem require a simple one

puskin94:
@moollaza Travis failed due not my code:
`The command "dzil authordeps | grep -ve '^\W' | xargs -n 5 -P 10 cpanm  --quiet   --notest  --skip-installed" failed and exited with 123 during .` 
Other tests are passed

moollaza:
LGTM :+1:

moollaza:
Thanks @puskin94

PR #1644, opened 2015-03-16
Removed ambiguous tickers and triggers

brianrisk:
@moollaza After seeing which API requests were called over the past few days, I have noticed that certain ticker symbols and certain trigger phrases were being called in combinations that implied the users did not truly desire fundamentals data with those queries.

A removed trigger is "price".  While "AAPL price" strongly implies a fundamentals response, "FORD price" does not.  In the end, I decided that "price" as an indicator would produce more inappropriate queries than not.

I also removed many ticker symbols that had common meanings when paired with indicators.  For example those submitting "HOME income" and "HOME revenue" queries are most likely interested in making money at home.

Note that many tickers are English words, and I did not remove those that, when paired with indicator triggers, are not typical English phrases.  However, I will continue to monitor requests to see if any do not make sense.

### DIFF
--- a/share/spice/quandl/fundamentals/fundamentals_triggers.yml
+++ b/share/spice/quandl/fundamentals/fundamentals_triggers.yml
@@ -26,7 +26,6 @@ ebt:	EBT
 eps:	EPS
 equity ratio:	DE
 equity:	EQUITY
-events:	EVENTS
 exchange:	NCFX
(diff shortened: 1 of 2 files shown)


moollaza:
@brianrisk thanks a lot for monitoring this! LGTM :+1:

PR #1650, opened 2015-03-17
Weather triggering

tommytommytommy:
Hello DDH community,

I would like to limit the weather trigger further until we are able to filter queries through the upcoming places detection logic, because we can generate some silly results (see http://www.reddit.com/r/techsnap/comments/2w0c27/odd_duckduckgo_weather_bug/). In combination with the internal trigger, we should still return Forecast results for most relevant queries. Once places detection goes live, we can remove this change and implement all of the relevancy and triggering features here again.

I modified the tests but left most of the code inside the Spice block in anticipation of using it again (would be happy to remove it if it's cleaner).


moollaza:
@tommytommytommy this LGTM :

My only thought is that we should simply comment out everything for now and add in something like:

```perl
spice is_cached => 0;

triggers start => "weather"
handle remainder => sub {
    return if $_;
    return $loc->loc_str;
}
```

moollaza:
LGTM :+1:

moollaza:
(Travis failing for unrelated reasons...)

PR #1653, opened 2015-03-18
Movies: Simplify handle function, remove old comments

moollaza:
Noticed this was a little outdated because we previously had many more triggers. There's no need for the logic we currently have in the `handle`

/cc @jagtalon


iambibhas:
How about adding the trigger `movie`?

moollaza:
@iambibhas meant to reply to you on IRC. This used to have lots of triggers but relevancy was and issue. We've since removed most of them and rely on an internal triggering method, based on the presence of certain domains in our organic links. We call this "deep triggering". 

I noticed it wasn't triggering is some cases that it should though and made a task for the team to investigate what's going on.

iambibhas:
I wish I could work more closely. These all looks fascinating.

jagtalon:
@moollaza a lot cleaner--thanks! :+1: Hope we get deep stuff on open source soon. :)

oisodse

PR #1659, opened 2015-03-19
Disable prettyprint of API response

moollaza:
Saves us about 20Kb per request!

/cc @jagtalon


MrChrisW:
@moollaza :+1: Oops! That was left in there for debugging purposes, nice catch :smile:

PR #1661, opened 2015-03-19
limit time to capitals

### LINKED ISSUE #1657: Time: Favouring City over Country causes relevency issues
For the query ["time in whitby, england"](https://duckduckgo.com/?q=time+in+whitby%2C+england&ia=time) we're showing the time for Whitby, Canada. Likewise ["time in whitby, new zealand"](https://duckduckgo.com/?q=time+in+whitby%2C+new+zealand&ia=time) shows Whitby, Canada.

In both cases Time and Date does not have a result for the correct city, but they have results for locations that are very close and in the same time zone. We should be favouring results that match the same country as the one specified in the query.

["time in Paris, Canada"](https://duckduckgo.com/?q=time+in+paris%2C+canada&ia=time) is another example. 

In all these cases TimeAndDate **doesn't** have an exact match for the city and country, but they give a result in the correct country that's in the same timezone (looking at a map of the results)

jdorweiler:
Fixes #1657 
This limits the time IA to only return if it finds a known capitol from the capitols.yml file.  There is also an internal pr to go with its one. 
Bolth PRs are installed on ddh4 for testing. https://ddh4.duckduckgo.com/?q=seattle+time&ia=time
@moollaza @russellholt

### DIFF
--- a/lib/DDG/Spice/Time.pm
+++ b/lib/DDG/Spice/Time.pm
@@ -35,9 +35,8 @@ handle query_lc => sub {
     if (my $caps = $capitals->{$q}) {
         # These are internally sorted by population, so assume they want the big one for now.
         $q = string_for_search($caps->[0]);
+        return $q;
     }
-
-    return $q;
 };
 
 sub string_for_search {
--- a/share/spice/time/time.js
+++ b/share/spice/time/time.js
@@ -9,12 +9,16 @@
         var script = $('[src*="/js/spice/time/"]')[0],
             source = $(script).attr("src"),
             // Query is normalized as we'll normalize the generated strings.
-            query = decodeURIComponent(source.match(/time\/([^\/]+)/)[1]).toLowerCase(),
+            // if we have a comma separated query it is in the form:
+            // "town, state, country"  but state is optional
+            query = decodeURIComponent(source.match(/time\/([^\/]+)/)[1]).toLowerCase().split(','),
             chosen;
 
         for(var i = 0; i < api.locations.length; i++) {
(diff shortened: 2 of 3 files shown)


moollaza:
LGTM :+1:

moollaza:
@jdorweiler nicely done, this fixes most of the problems I noticed!

a couple remaining edge cases:

- https://ddh4.duckduckgo.com/?q=time+in+whitby+united+kingdom (no idea what's happening here?)
- https://ddh4.duckduckgo.com/?q=time+in+port+perry (Port Perry is a small town 20 mins from me -- I think our isRelevant is passing this one through?)

jdorweiler:
@moollaza sorry it was still running the old Time.pm module.  I updated it.  
The api doesn't seem to have most smaller cities so those two that you found shouldn't show now. 
Basically, if you go to their site and search for a city and it's not in the top few results then we should be showing Time for that query.

moollaza:
@jdorweiler ah okay. Testing again.

moollaza:
- https://ddh4.duckduckgo.com/?q=time+in+oshawa (should be triggered by deep)
- https://ddh4.duckduckgo.com/?q=time+in+hamilton&ia=time (interestingly the first organic is more relevant but that may just be because Hamilton Canada is more popular than Bermuda?)

moollaza:
@jdorweiler pretty awesome increase in relevancy! Nicely done.

jdorweiler:
Hamilton is the capital of bermuda :smile: 

I see an improvement I can make to deep to get the others though.

moollaza:
>Hamilton is the capital of bermuda

Haha well then...I guess the IA is right :)

>I see an improvement I can make to deep to get the others though.

Awesome!

PR #1662, opened 2015-03-21
simplify tornodes, reduce custom design elements and css

moollaza:
I simplified the code a bit using newly provided Moment.js, and I moved towards a more canonical structure so you can more easily see the `Spice.add()` call and the object we pass it.

I also removed the unused `detail` template.

Regarding the design, I remove a bunch of the CSS in favour of leveraging our new template variants. @abeyang I know this is a slight departure from your original design, but I feel this is a bit more in line with the direction we're now headed?

2015-03-21_01h23_42
2015-03-21_01h24_11

---

Any thoughts?

/cc @jagtalon @abeyang @mogigoma


share/spice/tor_node/tor_node.js
```diff
@@ -79,33 +150,47 @@
...
-	base.templates.options.content = "record";
-	base.templates.options.keySpacing = true;
-        base.meta.sourceUrl = node.url;
+        base.templates = {
+            group: "text",
+            options: {
+                content: "record",
+                keySpacing: true
+            },
+            meta: {
```
MrChrisW: Can we add 
```
itemType: "Nodes"
```

selection_298

share/spice/tor_node/tor_node.js
```diff
@@ -65,8 +136,8 @@
         addField("IPs", node.addrs.join(", "));
```
MrChrisW: Can we update this to be:
```javascript
var ip_key = node.addrs.length > 1 ? "IPs" : "IP";
addField(ip_key, node.addrs.join(", "));
```

share/spice/tor_node/tor_node.js
```diff
@@ -115,72 +200,13 @@
...
-            templates: {
-                options:{
-                    moreAt: true
-                }
-            }
-        };
-    };
-
     function stripPort(addr) {
         return addr.replace(/:[0-9]+$/, "");
```
MrChrisW: This function needs to be updated to remove the brackets from an IPv6 address, these are not required if the [port number is removed](http://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers)

Example IPv6:

Before: `[2a01:4f8:100:112c::57]:9001`
After: `2a01:4f8:100:112c::57`

share/spice/tor_node/tor_node.js
```diff
@@ -79,33 +150,47 @@
...
+        };
 
         base.normalize = function(node) {
             // Handle the node's locale.
             node.icon = DDG.settings.region.getSmallIconURL(getCountryCode(node.country));
 
             // Handle the node's uptime history.
-            node.uptime = timeFromNow(node.last_restarted);
-            node.downtime = timeFromNow(node.last_seen);
+            node.uptime = moment(node.last_restarted).fromNow(true);
```
MrChrisW: One last suggestion would be to use `sort_fields` and `sort_default:` to sort the tiles by uptime. I've experimented with this, it looks like you'd need to create a new date object maybe in Unix timestamp format then compare the two within `sort_fields`

jagtalon:
Ping @moollaza

abeyang:
Sorry @moollaza -- already whipped this in Photoshop before I got this running on my server. Please take a stab first, and I can help you with whatever css is necessary...

tor-v1 1

tor-v1 1-fonts

tor-v1 1-margins

Hopefully, the *elclasses* work here, so here's a quick/dirty cheatsheet:
- `#666` -> `.tx-clr--slate-light`
- `#999` -> `.tx-clr--grey`
- `14px` -> `.tx--14`

share/spice/tor_node/tor_node.js
moollaza: :+1:

share/spice/tor_node/tor_node.js
moollaza: :+1:

share/spice/tor_node/tor_node.js
moollaza: :+1:

share/spice/tor_node/tor_node.js
moollaza: :+1:

moollaza:
@abeyang here's how it looks: 

2015-04-14_22h29_36

moollaza:
Looks like the red and green status colours were off, updated to meet styleguide red/green:

2015-04-14_22h34_56

moollaza:
@abeyang @MrChrisW thanks for the great feedback -- should be implemented now.

share/spice/tor_node/tor_node.js
```diff
@@ -6,26 +6,103 @@
...
+                itemType: "Nodes"
+            },
+            templates: {
+                options:{
+                    moreAt: true
+                }
+            },
+            sort_fields: {
+                'uptime' : function (a,b) {
+                    return moment(a.last_restarted).isBefore(b.last_restarted) ? -1 : 1;
```
MrChrisW: @moollaza Nice! I like this :+1:

share/spice/leak_db/leak_db.js
```diff
@@ -24,11 +24,10 @@
...
-                group: 'base',
+                group: 'list',
                 options:{
                     content: 'record',
-                    moreAt: true,
-                    rowHighlight: true
+                    moreAt: true
                 }
             }
         });
```
MrChrisW: Don't think this should be in here :smiley:

share/spice/tor_node/tor_node.js
```diff
@@ -6,26 +6,101 @@
...
 
-        // Put together the base of the Spice.
-        var base = get_shared_spice_data(api_result);
-        if (base === null) {
-            return Spice.failed("tor_node");
-        }
+        // Filter out the trigger from the query.
+        var words = DDG.get_query().split(" "),
+            reg   = /tor|onion/i,
+            query = (reg.test(words[0])) ? words.pop() : words.shift();
```
MrChrisW: This is interesting, should it be updated to use the more "standard approach"? Don't think it makes any difference... only thing I can think of is the triggers changing in future... meaning the `reg` var will need updating (edge case)

```js
        var script = $('[src*="/js/spice/tor_node/"]')[0],
            source = $(script).attr("src"),
            query = decodeURIComponent(source.match(/tor_node\/([^\/]+)/)[1]);
```

abeyang:
Sweet. Can we just tweak the margins of the snippet a bit:

```css
.tile--tor_node .tile__content {
  margin: 1.181em 0;
}
```

Also, not sure if this is just some anomaly with my dev server, but it off when I'm viewing this with my retina MBP:

http://cl.ly/image/3A19233w2z2H

moollaza:
>Also, not sure if this is just some anomaly with my dev server, but it off when I'm viewing this with my retina MBP:

@abeyang good catch! The `border-radius` is too small, will update to `100%`

moollaza:
Updated:

![2015-04-28_10h47_42](https://cloud.githubusercontent.com/assets/873785/7372319/1cf900e0-ed94-11e4-929b-3f7b7f2ee268.png

PR #1667, opened 2015-03-22
Reviving old branch - Ideas/deb

iambibhas:
One of the stale branches from #1535. Using the current frontend format and changed the API being used. Debian now has an API that returns JSON of the versions.


lib/DDG/Spice/DebVersion.pm
```diff
@@ -0,0 +1,21 @@
...
+triggers any => "debian version", "debian versions", "deb ver", "debian ver";
+
+# spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
+spice to => 'http://sources.debian.net/api/src/$1/';
+
+spice wrap_string_callback => 1;
+
+handle remainder => sub {
+    my $DebPackage = lc $_;
+    $DebPackage =~ s/\s+$//g;
```
moollaza: small nit but you can `use Text::Trim` above and then simply `trim()` here

lib/DDG/Spice/DebVersion.pm
```diff
@@ -0,0 +1,21 @@
+# Get the version of a package in various debian repositories
+package DDG::Spice::DebVersion;
+
+use DDG::Spice;
+
+triggers any => "debian version", "debian versions", "deb ver", "debian ver";
```
moollaza: Probably best to use `startend` otherwise we'll trigger this on every query with these words in it.

lib/DDG/Spice/DebVersion.pm
```diff
@@ -0,0 +1,21 @@
...
+
+# spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
+spice to => 'http://sources.debian.net/api/src/$1/';
+
+spice wrap_string_callback => 1;
+
+handle remainder => sub {
+    my $DebPackage = lc $_;
```
moollaza: We should verify that our debian version is actually a deb version and not a multi-word string or something else we shouldn't bother sending to the API.

share/spice/deb_version/deb_version.js
```diff
@@ -0,0 +1,47 @@
...
+                };
+
+                if (Object.keys(data.record_data).length === 0) {
+                    return Spice.failed('deb_version');
+                }
+
+                return data;
+            },
+            templates: {
+                group: 'base',
```
moollaza: Let's use the `text` template -- base is to be avoided unless absolutely necessary for a custom tile.

share/spice/deb_version/deb_version.js
```diff
@@ -0,0 +1,47 @@
...
+    env.ddg_spice_deb_version = function(api_result_str){
+        var api_result = JSON.parse(api_result_str);
+
+        if (api_result.error) {
+            return Spice.failed('deb_version');
+        }
+
+        Spice.add({
+            id: "deb_version",
+            name: "Debian Package Versions (" + api_result.package + ")",
```
moollaza: This `name` is a little too long. `Software` is more appropriate. I would move the package name to the `title`.

iambibhas:
@moollaza for some reason some commits by someone else for the spices whois, bitcoin, airlines etc showed up on this branch. Not exactly sure how. Any idea?

jagtalon:
@iambibhas @moollaza I think I'd like a simpler title for it:

screen shot 2015-03-26 at 1 03 02 pm
screen shot 2015-03-26 at 1 03 14 pm

Instead of:

screen shot 2015-03-26 at 1 00 19 pm

What do you think?

jagtalon:
@iambibhas looks like just pulling and merging master should fix it. Did the same and it's now correctly showing only 3 files changed:

screen shot 2015-03-26 at 1 06 13 pm

lib/DDG/Spice/DebVersion.pm
```diff
@@ -0,0 +1,24 @@
...
+
+triggers startend => "debian version", "debian versions", "deb ver", "debian ver";
+
+# spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
+spice to => 'http://sources.debian.net/api/src/$1/';
+
+spice wrap_string_callback => 1;
+
+handle remainder => sub {
+    my $query = lc trim $_;
```
jagtalon: Nitpicking but probably cleaner to use `handle remainder_lc` so that `$_` is already lowercased for you.

lib/DDG/Spice/DebVersion.pm
```diff
@@ -0,0 +1,24 @@
+# Get the version of a package in various debian repositories
+package DDG::Spice::DebVersion;
+
+use Text::Trim;
+use DDG::Spice;
+
+triggers startend => "debian version", "debian versions", "deb ver", "debian ver";
+
+# spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
+spice to => 'http://sources.debian.net/api/src/$1/';
```
jagtalon: Looks like this URL is redirecting to HTTPS--we don't support redirects in production so best to switch to https.

lib/DDG/Spice/DebVersion.pm
```diff
@@ -0,0 +1,24 @@
...
+
+spice wrap_string_callback => 1;
```
jagtalon: Best to use `wrap_jsonp_callback` since it's already in JSON format.

share/spice/deb_version/deb_version.js
```diff
@@ -0,0 +1,48 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_deb_version = function(api_result_str){
+        var api_result = JSON.parse(api_result_str);
```
jagtalon: No need to parse JSON if we use `wrap_jsonp_callback`

share/spice/deb_version/deb_version.js
```diff
@@ -0,0 +1,48 @@
...
+
+        if (api_result.error) {
```
jagtalon: Check if `api_result` exists before accessing the `error` property, i.e.,

```javascript
if(!api_result || api_result.error) {
    ...
}
```

On that note, does the API actually return an `error` property when there's an error?

lib/DDG/Spice/DebVersion.pm
moollaza: we should really remove that option altogether...

lib/DDG/Spice/DebVersion.pm
```diff
@@ -0,0 +1,24 @@
+# Get the version of a package in various debian repositories
+package DDG::Spice::DebVersion;
+
+use Text::Trim;
+use DDG::Spice;
+
+triggers startend => "debian version", "debian versions", "deb ver", "debian ver";
+
+# spice from => '([^/]+)/?(?:([^/]+)/?(?:([^/]+)|)|)';
```
moollaza: Might as well remove this comment :)

moollaza:
@iambibhas @jagtalon agreed on the design. We should set the `title` to the name of the package, and `subtitle` to something like "Debian Package Versions".

lib/DDG/Spice/DebVersion.pm
iambibhas: TBH, I didn't know `remainder_lc` existed. Is it documented?

lib/DDG/Spice/DebVersion.pm
iambibhas: It's not redirecting to HTTPS for me. :confused: But HTTPS works and it's always the better choice. Will change it.

share/spice/deb_version/deb_version.js
iambibhas: It returns this if the package isn't found - 
```
{
    "error": 404
}
```

lib/DDG/Spice/DebVersion.pm
jagtalon: Aaaah that's my bad. It was HTTPSEverywhere doing it. Should be good. :)

iambibhas:
@jagtalon pulling and merging master on the ideas/deb branch didn't really help. :( Should I just take the files for this spice and create a new PR on top of the master branch?

moollaza:
>Should I just take the files for this spice and create a new PR on top of the master branch?

@iambibhas that would probably be best?

iambibhas:
@moollaza done. let's close this one and work on #1697. Sorry for messing with the stale branches with rebase. :|

PR #1668, opened 2015-03-23
World Bank, improved home prices

brianrisk:
@moollaza:  Some nice new additions for you.

1)  World Bank:  I'm starting small with this one - about 20 property triggers so that we can monitor how this goes in terms of quality results before diving right in.  The spice only makes a "spice to" call when one of these properties if paired with a valid country name or country code.  This fully replaces the previous Population spice while improving upon it by generalizing it to multiple World Bank data sets and also improving the final data display with percent change.  In these commits I have deleted the Population spice as it is now redundant.

Example searches:
world population
female population in the united states
france urban population
food production in argentina

2) Home value searches are improved in that metro names and state names can be used to identify regions.  Care was taken to remove cities with duplicate names (e.g. "Springfield") and to structure the spice so that city names that contained state names (e.g. "new york city") would call the city indicator.

Example searches:
new york city two bedroom houses
new york 2 bedroom homes

------
> IA Page: https://duck.co/ia/view/quandl_world_bank


moollaza:
@brianrisk thanks a lot! We'll take a look shortly and give you some feedback.

share/spice/quandl/world_bank/quandl_world_bank.js
```diff
@@ -0,0 +1,100 @@
...
+* To change this template use Tools | Templates.
+*/
+
+
+(function (env) {
+    'use strict';
+    env.ddg_spice_quandl_world_bank = function(api_result){
+
+        if (!api_result) {
+            return Spice.failed('quandl_world_bank: no results');
```
moollaza: these should just be passing in the `id`: `Spice.failed('quandl_world_bank')`

share/spice/quandl/world_bank/quandl_world_bank.js
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        var result = api_result;
+        
+        // we need two data points to get percent change
+        if (result.data.length < 2) {
+            return Spice.failed('quandl_world_bank: not enough data points');
```
moollaza: same here -- we're expecting the `id` to map it to other internal checks for the given IA

share/spice/quandl/world_bank/quandl_world_bank.js
```diff
@@ -0,0 +1,100 @@
...
+        result.url = 'https://quandl.com/' + result.source_code + "/" + result.code;
+
+        // add title tag for link:
+        result.urlTitle = 'View more World Bank data at Quandl';
+        
+        var recentValue = result.data[0][1];
+        var previousValue = result.data[1][1];
+        
+        // month array to make string version of date
+        var months = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
```
moollaza: Better to handle all this with moment.js: `moment(my_date).fromNow()`

brianrisk:
@moollaza I've corrected the IDs (in the other Quandl spices as well).

Regarding Moment.js, for information workers, they appreciate knowing precise times for data points.  The rounding applied with fromNow() giving results like "4 years ago" would obfuscate that level of detail.  Do you think we could keep this the way it is?

share/spice/quandl/world_bank/world_bank_primary.yml
```diff
@@ -0,0 +1,22 @@
...
+stocks traded:	CM_MKT_TRAD_CD
+urban population:	SP_URB_TOTL
+labor force:	SL_TLF_TOTL_IN
+food production:	AG_PRD_FOOD_XD
+telephone lines:	IT_MLT_MAIN
+national expenditure :	NE_DAB_TOTL_CN
+shipping connectivity:	IS_SHP_GCNW_XQ
+shipping connectivity index:	IS_SHP_GCNW_XQ
+shipping index: IS_SHP_GCNW_XQ
+droughts:	EN_CLC_MDAT_ZS
```
moollaza: if you add in `rainfall` we can also deprecate the Rainfall Spice :)

brianrisk: Done!  Note that this often returns 0% change (e.g. "Argentina rainfall").  Just having "0%" looks odd.  What do you think about adding the word "change" after the percent change?

moollaza:
@brianrisk looks like your Branch is a little behind Master. Could your try rebasing your changes so I can easily merge this once it's ready?

The code looks pretty good, I'll put it on a dev machine for the design team to give it a look.

Thanks again!

brianrisk:
@moollaza rebased.  About the code:  thanks for the compliment.  I feel a basic pattern is forming with these spice's where there is an object (country, company, zip code) and an indicator (rainfall, revenue, avg house price).  However, each database is just different enough that I'm glad these are broken into modules.

lib/DDG/Spice/Quandl/WorldBank.pm
```diff
@@ -0,0 +1,77 @@
...
+spice proxy_cache_valid => "418 1d";
+
+handle sub {
+
+    my $query = lc $_;
+    
+    # find which secondary trigger was used
+    my $secondary;
+    for my $trigger (@secondary_keys) {
+        if ( $query =~ /\b$trigger\b/ ) {
```
moollaza: Seeing as the keys are sorted, we can actually join them outside the handle to build a regex. We can then use that for a 1-line check if any `secondary_keys` match our query:

```perl
my $secondary_qr = join "|", @secondary_triggers;
return unless my $secondary = m/\b($secondary_qr)\b/;
```

brianrisk:
@moollaza I tried you suggestion which was:  
return unless my $secondary = m/\b($secondary_qr)\b/;

unfortunately that resulted in $secondary being undefined.  I got it working in two lines:
$query =~ m/\b($secondary_qr)\b/;
return unless my $secondary = $1;

I think though that this regex will probably be slower than the loop because the regex is considering all \b boundaries whereas the loop knows how the words should be grouped.

moollaza:
>unfortunately that resulted in $secondary being undefined. I got it working in two lines:
$query =~ m/\b($secondary_qr)\b/;
return unless my $secondary = $1;

Sorry, my mistake! What you have is good, I think the more Perl-y way to write it would be:

```perl
return unless $query =~ m/\b($secondary_qr)\b/;
my $secondary = $1;
...
```

Your way it totally fine though :+1: 

------

>I think though that this regex will probably be slower than the loop because the regex is considering all \b boundaries whereas the loop knows how the words should be grouped.

I was curious to know for sure which is faster so I tested a benchmark and it looks like the single-large-regex method is considerably (5000%+) faster than the loop approach. I think it's because creating a regex is an expensive operation and the loop builds and optimizes hundreds of regexs as it loops over them.

**Disclaimer:** There's a good chance I've made a mistake in the benchmark, but I think I've done it right? I've haven't benchmarked Perl code before :)

> Here's the test code: https://gist.github.com/moollaza/1b02a74847d7aae578d9

lib/DDG/Spice/Quandl/WorldBank.pm
```diff
@@ -0,0 +1,70 @@
...
+
+    my $query = lc $_;
+    
+    # find which secondary trigger was used
+    $query =~ m/\b($secondary_qr)\b/;
+    return unless my $secondary = $1;
+     
+    # find which primary trigger was used
+    my $primary;
+    for my $trigger (@primary_keys) {
```
moollaza: With my benchmark results in mind, I believe we should use a single large regex here too :)

brianrisk:
@moollaza I can't argue with a 50x speedup!  I implemented your more your more Perl-y approach and did similar modifications of the Fundamentals and HomeValues spices -- removing looped regex in favor of a single regex on a joined array.

moollaza:
@brianrisk thanks! I think this is ready for a design signoff :+1: 

/cc @abeyang (will put it on a server in a minute)

brianrisk:
@moollaza @abeyang cool!  What do you think about adding the word "change" after the percent change number?  This will help specifically for 0% cases.  For example "US rainfall" is confusing as the 0% does not seem to be related to anything.

lib/DDG/Spice/Quandl/HomeValues.pm
```diff
@@ -37,34 +46,47 @@ spice wrap_jsonp_callback => 1;
...
-    $_ =~ m/\b(\d{5})\b/;
+    # is it a 5-digit zip code?
+    $query =~ m/\b(\d{5})\b/;
     if ($1) {
         $region = $1;
         $indicator_type = "ZIP";
     }
     
+    # is it a metropolitan area?
+    if (! defined $region) {
```
moollaza: No big deal, but an `if .. elsif` would work here:

```perl
if ($query =~ m/../) {
     $region = $1;
     $indicator_type = ...;
} elsif ($query =~ m/../) {
   ...
} elsif ($query =~ m/../) {
   ...
}
```

moollaza:
Sorry for the delay guys! It's finally up on a testing machine: https://ddh8.duckduckgo.com/?q=world+population&ia=worldbank

/cc @abeyang

moollaza:
@brianrisk looks like we may want to round off the percent in the "female population" example to 2 decimal places?

moollaza:
@brianrisk https://ddh8.duckduckgo.com/?q=brazil+average+annual+rainfall&ia=worldbank

Looks like the data is incorrectly formatted here? Giving a dollar amount for rainfall. Also it seems like everything shows a 0% change for rainfall -- do you have any examples where it has changed?

share/spice/quandl/world_bank/world_bank_primary.yml
```diff
@@ -0,0 +1,25 @@
...
+floods: EN_CLC_MDAT_ZS
+extreme temperaturs:	EN_CLC_MDAT_ZS
```
moollaza: small typo here, should be `extreme temperatures`

moollaza:
@brianrisk one other thing I noticed is that we're not checking if a year was specified in the query -- which can lead to somewhat irellevant results: https://ddh8.duckduckgo.com/?q=canada+stocks+traded+2011&ia=worldbank

- Can we get the data for a specified year?
    - if not, we should probably ignore queries that specify a year...
- Can we get the data for a range? (to eventually produce a graph)

abeyang:
Looking good. But can we get rid of the *negative sign*? The "down arrow" already implies this.

http://cl.ly/image/3n2w450R0q3R

share/spice/quandl/world_bank/world_bank_primary.yml
brianrisk: Thank -- typo fixed

brianrisk:
@moollaza I've addressed the issues you have pointed out.
1) Dollar signs removed.  Sorry about that; those were carry-over from previous finance spices
2) If a year is specified, no result is returned.  We could perhaps do something more sophisticated later, but for the time being this will save returning wrong year data.
3) 0% change in rainfall seems to be across the board.  I checked the World Bank site and none of the examples have any change from 2012 to 2014: http://data.worldbank.org/indicator/AG.LND.PRCP.MM
4) all numbers are now rounded to 2 decimal places.

@abeyang I feel the "down arrow" is a nice graphic element that will help people more quickly understand this is a lower value.  However, I don't think everyone will necessarily pick up on it or notice it.  The negative sign leaves nothing to the imagination.  This is easily changed though, so if you feel strongly about this I can get rid of the negative.

brianrisk:
@moollaza Yes, we can get date ranges.  Charts can definitely be a future development!

abeyang:
@brianrisk I get you, but having *both* is what makes it confusing. =T

Can you get rid of the negative and keep the down arrow? It'll also be consistent with some of our other IAs, like stocks:

screen shot 2015-04-13 at 12 31 15 pm

Thanks!

moollaza:
@brianrisk I think this is good for merge -- looks like there's some conflicts right now though, could you please resolve those and then I'll be happy to merge this in, thanks!

brianrisk:
@moollaza that rebase took longer than expected!  I think everything is fine on the Quandl end.. Astrobin.t seems to be the showstopper.  I'll check it out.

brianrisk:
screen shot 2015-04-13 at 10 43 16 pm
To me this is confusing.  as there is no triangle between the two numbers, it's difficult to tell what the primary result is.

moollaza:
>that rebase took longer than expected!

Sorry about that! Lots of things are changing in Spice :)

>I think everything is fine on the Quandl end.. Astrobin.t seems to be the showstopper. I'll check it out.

Good catch, not sure how we let that typo in!

moollaza:
@abeyang any thoughts on that example? @brianrisk had suggested adding "change" to the 0% in those cases so it's more clear what's going on.

brianrisk:
@moollaza @abeyang: something like this:

screen shot 2015-04-14 at 2 33 17 pm

I don't want the user thinking, "0% of what?" or that there was no rain whatsoever.

moollaza:
Updated on DDH8

- https://ddh8.duckduckgo.com/?q=world+population&ia=worldbank
- https://ddh8.duckduckgo.com/?q=US+rainfall&ia=worldbank
- https://ddh8.duckduckgo.com/?q=+food+production+in+argentina&ia=worldbank
- https://ddh8.duckduckgo.com/?q=female+population+in+canada&ia=worldbank

/cc @abeyang

abeyang:
Hi @brianrisk , let's just stick with the *up arrow* followed by 0%. So:

> 715 ^ 0%

brianrisk:
@abeyang that's some good thinking!  Here's the result:

screen shot 2015-04-14 at 7 50 39 pm

moollaza:
LGTM :+1:

moollaza:
@brianrisk congrats! It's live: https://duckduckgo.com/?q=france+urban+population&ia=geography

brianrisk:
Love it!  Thanks so much for your efforts on this @moollaza and @abeyang.  I hope to soon have an update to this spice which adds more World Bank indicators to the yml file.  Zaahir, keep me in mind if you have any ideas about time series chart displays.

PR #1673, opened 2015-03-23
Newint - support for moment.js

iambibhas:
Added support as mentioned in #1554.

### DIFF
--- a/share/spice/newint/newint.js
+++ b/share/spice/newint/newint.js
@@ -6,41 +6,41 @@
             return Spice.failed('newint');
         }
 
-        Spice.add({
-            id: "newint",
-            name: "Magazines",
-            data: api_result,
-            meta: {
-                itemType: "New Internationalist magazines",
-                sourceName: "New Internationalist",
-                sourceUrl: 'https://digital.newint.com.au/issues'
-            },
-            normalize: function(item){
-                var month_names = new Array("January", "February", "March", 
-                    "April", "May", "June", "July", "August", "September", 
-                    "October", "November", "December");
-                var editors_letter_strip_tags = item.editors_letter_html.replace(/<h[1-6]>(.*?)<\/h[1-6]>/g, "<p>$1. </p>");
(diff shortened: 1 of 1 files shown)


share/spice/newint/newint.js
```diff
@@ -6,41 +6,40 @@
...
+                id: "newint",
+                name: "Magazines",
+                data: api_result,
+                meta: {
+                    itemType: "New Internationalist magazines",
+                    sourceName: "New Internationalist",
+                    sourceUrl: 'https://digital.newint.com.au/issues'
+                },
+                normalize: function(item){
+                    var editors_letter_strip_tags = item.editors_letter_html.replace(/<h[1-6]>(.*?)<\/h[1-6]>/g, "<p>$1. </p>");
```
moollaza: Looks like we can use `DDG.strip_html` and `DDG.strip_href` here?

share/spice/newint/newint.js
```diff
@@ -6,41 +6,40 @@
...
+                    editors_letter_strip_tags = editors_letter_strip_tags.replace(/<a(.*?)>(.*?)<\/a>/g, "$2");
+                    return {
+                        release_date: moment(item.release).format('MMMM YYYY'),
```
moollaza: :+1:

share/spice/newint/newint.js
iambibhas: It seems to replace just the header tags and anchors, and leaves everything else. So maybe `strip_href`, not `strip_html`, that is if we want to keep the formatting as it is. Or we could just pass it through `strip_html` and show the text only.

moollaza: @iambibhas wow I just noticed a few lines down we're running it through `DDG.strip_html` so I think these two lines are actually redundant. We should be able to simply remove them :)

iambibhas: Ugh, I also totally missed it. :| Yeah, we can just remove them.

PR #1676, opened 2015-03-23
Content Layout Updates

sdougbrown:
This depends on an internal PR.

We're adding some new built-in templates (list), and simplifying some spacing and alignment concerns so this PR is just two update a few affected spices.

to @bsstoner


sdougbrown:
Internal PR merged. :+1:

moollaza:
@sdougbrown it looks like the change isn't merged to Master internally yet -- might want to revert this until then?

PR #1678, opened 2015-03-24
Seasons: improve logic and triggering

moollaza:
Simplified the code, improved triggering and added more tests.

Using a larger regex to ensure queries match the required form. This fixes the problem of triggering on queries that have trailing words. e.g. "summer solstice tomorrow"

/cc @jagtalon @mayo

### DIFF
--- a/lib/DDG/Spice/Seasons.pm
+++ b/lib/DDG/Spice/Seasons.pm
@@ -87,7 +87,7 @@ my %aliases = (
 # Handle statement
 handle query_lc => sub {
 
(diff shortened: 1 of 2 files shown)


jagtalon:
@moollaza thank you. Looks good but lemme just do a quick test

0-nfjuyrk4woqj_xzz

jagtalon:
These could just be issues but here are some notes:

- Would like it to work for these queries too:
    - when is spring?
    - when is summer?
    - when does spring start?
    - spring in the us
    - spring in the ph

- Maybe we should use the timezone of the place like in W|A? Might not be important though

screen shot 2015-03-24 at 4 57 30 pm

jagtalon:
:+1:

jagtalon:
https://github.com/duckduckgo/zeroclickinfo-spice/issues/1683

PR #1680, opened 2015-03-24
spice/newint - removed unnecessary HTML formatting

iambibhas:
We were running the whole thing through `strip_html` anyway.


jagtalon:
@iambibhas ooh nice catch. Yeah def unnecessary.

jagtalon:
:+1:

PR #1681, opened 2015-03-24
require a space between flight and  flight number

jdorweiler:
Check for a space in between the flight and flight number.  Otherwise queries like "sha512" will trigger airlines.
@moollaza @russellholt

### DIFF
--- a/lib/DDG/Spice/Airlines.pm
+++ b/lib/DDG/Spice/Airlines.pm
@@ -19,7 +19,7 @@ attribution web => [ 'https://www.duckduckgo.com', 'DuckDuckGo' ],
 spice to => 'https://duckduckgo.com/flights.js?airline=$1&flightno=$2&callback={{callback}}';
 spice from => '(.*?)/(.*)';
 
-triggers query_lc => qr/^(\d+)\s*(.*?)(?:[ ]air.*?)?$|^(.*?)(?:[ ]air.*?)?\s*(\d+)$/;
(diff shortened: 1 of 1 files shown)


moollaza:
@jdorweiler I believe we want to allow those kinds of queries to trigger it though? I always use "aa102" as a test: https://duckduckgo.com/?q=aa102&ia=flights

jdorweiler:
Forgot to mention there's an internal PR to handle the other queries.

moollaza:
Aah. Well then. LGTM :+1:

jagtalon:
@jdorweiler what does the internal PR do?

jdorweiler:
@jag Test it out here: https://ddh4.duckduckgo.com/?q=lufthansa+flight+to+phl&ia=flights

jagtalon:
@jdorweiler So queries such as "aa102" will now be triggered internally when we find airline websites in the SERP is that it?

moollaza:
LGTM :+1:

moollaza:
@jdorweiler I just noticed that "AA102" isn't show a result anymore -- it looks like it triggers the Spice but makes a call for `/spice/airlines/AA1/02`, instead of `/spice/airlines/AA/102`

jdorweiler:
@moollaza looks okay to me. Do you still see it?

moollaza:
@jdorweiler LGTM now.

PR #1682, opened 2015-03-24
Updated 'plural form of...' line and "audio unavailable' padding

chrismorast:
cc @moollaza @sdougbrown @bsstoner


jagtalon:
@chrismorast looks good! :+1: 

screen shot 2015-03-24 at 4 27 12 pm

moollaza:
@jagtalon @chrismorast actually, we need to wrap the "Plural form of" section in an `{{if plural_form}}` check

On the lice site, we don't show "Plural form of" because `word` is `cats` in this case (Hence the subtitle looks broken)

chrismorast:
Ahh, got it.

PR #1685, opened 2015-03-24
New IA for Sunrise/Sunset timing

iambibhas:
Work in progress. Shows the sunset/sunrise times for the capitals of countries.

 - Tested with TimeAndDate test locations. Need to test with more locations.
 - UI may need work. Showing plain text now.
 - Not sure if the `capitals.yml` file inside `share/time` can be reused. For now copied it to this spice's folder.

------
> IA Page: https://duck.co/ia/view/sun_rise_set


lib/DDG/Spice/SunRiseSet.pm
```diff
@@ -0,0 +1,43 @@
+package DDG::Spice::SunRiseSet;
+# ABSTRACT: Returns package information from npm package manager's registry.
+
+use DDG::Spice;
+use YAML::XS qw( Load );
+
+primary_example_queries "sunset Oslo", "sunrise Oslo";
+secondary_example_queries "Oslp sunset time", "Oslo sunrise time";
+description "Provides the sunrise/set time of country, city or state searched";
+name "Time";
```
moollaza: Don't forget to change this too :)

share/spice/sun_rise_set/sun_rise_set.js
```diff
@@ -0,0 +1,59 @@
...
+                country: location.geo.country.name,
+                sunrise_time: sunrise_time + ' AM',
+                sunset_time: sunset_time + ' PM'
+            },
+            meta: {
+                sourceName: "timeanddate.com",
+                sourceUrl: 'http://www.timeanddate.com/worldclock/results.html?query=' + location.geo.name
+            },
+            templates: {
+                group: 'base',
```
moollaza: Let's use the `text` template and then we can supply a title (and subtitle if necessary) from the JS. We can copy the design from the existing SunInfo Goodie (which we'll deprecate after merging this) and re-use the image assets we have there as well.

iambibhas:
Brought the templates from the goodie. This is how it looks now -

iambibhas:
I changed the css class name to `zci--sun_rise_set` and was *just* going to push the changes. :D

moollaza:
>I changed the css class name to zci--sun_rise_set and was just going to push the changes. :D 

Awesome!

moollaza:
@chrismorast do you still have the original PNG's we made for this? Are we still OK with using SVG? (This is now a Spice so choosing 1x, 2x, 3x images is quite easy if we want to go that route)

iambibhas:
Would using PNG instead of SVG provide any significant benefit? SVG should work on all browsers, right?

iambibhas:
Tested with more locations, seems to work, but just for country capitals. Would the deep triggering handle other locations?

iambibhas:
I worked a little on using regex triggers and searching for user location in the capital list when only `sunset` or `sunrise` is searched. Should I push that?

moollaza:
> I worked a little on using regex triggers
@iambibhas we prefer only to use those when absolutely necessary (e.g. when you need to match on numbers) Word trigger are much more efficient and should be used. 

> searching for user location in the capital list when only sunset or sunrise is searched.

That sounds pretty good -- you can probably borrow code from the SunInfo Goodie for that as well.

iambibhas:
> We get the best performance and relevancy by using relatively specific word triggers and then using a regex in the handle to make sure the query is something we should be acting on.

That's good to know. I'll modify it a little and push something.

> you can probably borrow code from the SunInfo Goodie for that as well.

Looking into it now. Thanks.

iambibhas:
@jagtalon this one is working and ready to be tested. If any changes are needed, let me know.

chrismorast:
Sorry, just saw the request for PNGs/SVGs. Here you go.
icon_sunrise
icon_sunset

moollaza:
@iambibhas please update the template to load the svg image (or fallback to png when needed).

You should be able to check if `Modernizr.svg` is true for SVG support. If that doesn't work, you can use our own `is_retina` variable or `is_mobile` (either should be fine).

Also one small nit, can you make sure the images are vertically centred with the icons?

After that this should be good for merge!

iambibhas:
@moollaza I'm already using svgs. I thought you wanted me to use PNG?

iambibhas:
This is how it looks now -

moollaza:
>I'm already using svgs. I thought you wanted me to use PNG?

Sorry for the confusion. The original Goodie was only using SVG only I believe. We should be using SVG unless the browser doesn't support SVG (e.g. IE < 8) and in those cases fallback to PNG

iambibhas:
@moollaza added the `Modernizr` check in last commit. Falls back to png. Need to test with IE.

iambibhas:
So, @MrChrisW just tested it on IE8. The images fallback to PNG. Looks like they are working. :D

lib/DDG/Spice/SunRiseSet.pm
```diff
@@ -0,0 +1,53 @@
+package DDG::Spice::SunRiseSet;
+# ABSTRACT: Returns package information from npm package manager's registry.
```
moollaza: Looks like this needs to be changed ;)

lib/DDG/Spice/SunRiseSet.pm
```diff
@@ -0,0 +1,53 @@
...
+
+use DDG::Spice;
+use YAML::XS qw( Load );
+use Text::Trim;
+
+primary_example_queries "sunset Oslo", "sunrise Oslo";
+secondary_example_queries "Oslp sunset time", "Oslo sunrise time";
```
moollaza: Typo in fist example query, "Oslp"

moollaza:
> So, @MrChrisW just tested it on IE8. The images fallback to PNG. Looks like they are working. :D 

:+1: Awesome.

moollaza:
Up on my dev machine: https://moollaza.duckduckgo.com/?q=sunrise+oslo&ia=sunrisesunset

/cc @abeyang

moollaza:
@iambibhas I spoke to the design team and they say it should look identical to the existing SunInfo Goodie:

2015-04-15_17h35_32

The location and date should be on one line at the top, with the location pin and the sunrise times should be larger (`text--primary` class).

I think you should be able to provide a blank `title` and use the subtitle to achieve the right look. If not, you can just use the `title` and re-style it with CSS to get the smaller fontsize and coloring.

moollaza:
@iambibhas much better. It's not exactly the same, but the Goodie has a fully custom template and changes the base font-size which I'd like to avoid. We'll make an issue with the final adjustments for you, or anyone else to implement after.

PR #1687, opened 2015-03-25
Fix for SeatGeek::Sports: JS Error

### LINKED ISSUE #1686: SeatGeek::Sports: JS Error causing failure
Tried triggering the SeatGeek Sports Spice today and I noticed there was a console error:

2015-03-24_22h08_32

// @MariagraziaAlastra @jagtalon

MariagraziaAlastra:
Fixes #1686
@moollaza sorry about that, I must have missed this one when I removed the function and the calls to it!

### DIFF
--- a/share/spice/seat_geek/sports/seat_geek_sports.js
+++ b/share/spice/seat_geek/sports/seat_geek_sports.js
@@ -31,7 +31,6 @@
                 var performer;
                 var both_performers = [];
                 var title;
-                var logo;
                 for(var i = 0; i < item.performers.length; i++) {
                     if(DDG.stringsRelevant(item.performers[i].name.toLowerCase(), clean_query, [], 3, true) || DDG.stringsRelevant(item.performers[i].short_name.toLowerCase(), clean_query, [], 3, true)) {
                         relevant = true;
--- a/share/spice/seat_geek/sports/seat_geek_sports.js
+++ b/share/spice/seat_geek/sports/seat_geek_sports.js
@@ -60,7 +60,6 @@
                     }
                 } else {
                     title = performer.name;
-                    logo = getLogo(title, item.taxonomies);
                 }
 
                 function getDate(date) {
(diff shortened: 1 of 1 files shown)


moollaza:
@MariagraziaAlastra thanks for the fix, unfortunately I'm now getting a different error, which is odd:

2015-03-25_10h03_11

Seems to be coming from line 61 (the line right above the `getLogo` call).

I think it's because after the next few games we'll be into the finals, but the teams aren't all determined yet. So `Performers` is an array with only 1 item.

The queries I tried were "chicago bulls tickets" and "toronto raptors tickets"

moollaza:
```js
for(var i = 0; i < item.performers.length; i++) {
    if(DDG.stringsRelevant(item.performers[i].name.toLowerCase(), clean_query, [], 3, true) || DDG.stringsRelevant(item.performers[i].short_name.toLowerCase(), clean_query, [], 3, true)) {
        relevant = true;
        performer = item.performers[i];
    }
    both_performers.push(item.performers[i].name);
}
```

Removing the `else` block seems to make things work -- does that seem right to you?

MariagraziaAlastra:
Ok so brief explanation: for generic events (queries like "nba tickets" and so forth) the title should be "[team1] vs [team2]", while for specific teams/performers (queries like "toronto raptors tickets") the title should instead display just the name of the opponent, so we're using the relevancy check to guess which is the case, and if it's the latter, determine the name of the opponent.

In this case it fails because we didn't foresee cases in which the performer would be just one, but the last commit should fix this.

Let me know if there are any issues :)

moollaza:
@MariagraziaAlastra thanks a lot! This is working so I'll merge. There's still one oddity that arrives from the way the API responses come. 

In some cases the only list 1 performer, but in other they list 2 where the second isn't actually a team its a perform with a name of "NBA Semifinals". It produces an odd looking result:

2015-03-25_17h42_15

Is there any reason we aren't using the `event.title`? In most cases it seems to suffice (e.g it gives us "Team 1 vs Team2" when there's two teams and gives us things like "title: "Eastern Conf Semifinals: Toronto Raptors vs TBD - Home Game 2 (Date TBA)" -- granted that's not ideal but it's perhaps better for now?

MariagraziaAlastra:
@moollaza It's just that event names aren't really predictable, even though a lot of them is some variation of "[team1] vs [team2]".

Yeah I know they list event names as performers too, which is why is difficult to nail the titles, even though the relevancy check does a bit of the job!

PR #1688, opened 2015-03-25
SimilarSites: Removed helper

### LINKED ISSUE #1672: SimilarSites: Simplify Handlebars Helper
This Spice uses it's own Handlebars helper to iterate over items and build up an HTML string. The code should be altered so the data is setup before hand to simplify or remove the Handlebars helper and move more HTML into the template rather than the JS.

If an array of objects is built beforehand (using a `normalize` function should work), we should be able to iterate over the array with our built-in `{{each}}` helper and get rid of the `similarsites_list` helper entirely.

iambibhas:
Fixes #1672. Removed the helper function and moved the HTML generation to the template.

### DIFF
--- a/share/spice/similar_sites/similar_sites.js
+++ b/share/spice/similar_sites/similar_sites.js
@@ -6,7 +6,7 @@
                 source = $(script).attr("src"),
                 query = source.match(/similar_sites\/([^\/]+)/)[1];
 
-        if(!api_result || api_result.num === 0) {
+        if(!api_result || api_result.num === 0 || api_result.status !== 'ok') {
             return Spice.failed('similar_sites');
         }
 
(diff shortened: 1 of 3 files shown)


share/spice/similar_sites/similar_sites.js
```diff
@@ -26,6 +26,25 @@
...
             },
 
+            normalize: function (data) {
+                var sites = [];
+
+                for (var i = 0; i < data.num; i++) {
+                    var url = data.results['r' + i];
+                    sites.push({
+                        url: url,
+                        name: url.replace("http://", "")
```
moollaza: Could probably simplify this with a single regex?

```js
url.replace( /https?:\/\/(www\.)?/, "");
```

share/spice/similar_sites/similar_sites.js
```diff
@@ -45,10 +64,10 @@
                 var toggle = false;
                 var $icon = $(".zci--similar_sites .chomp--link__icn");
```
moollaza: Let's make this jQuery code a little more efficient by caching our DOM lookup:

```js
var $zci = $(".zci--similar_sites"),
    $icon = $zci.find(".chomp--link__icn"),
    $more = $zci.find( ... ),
    $less = $zci.find( ... );
```

jagtalon:
@iambibhas looks good! :+1:

PR #1689, opened 2015-03-26
Dictionary: Only add subtitle if word is plural

moollaza:
Fixes small bug introduced in https://github.com/duckduckgo/zeroclickinfo-spice/pull/1682#issuecomment-85680571

/cc @jagtalon @chrismorast 

2015-03-25_21h03_42

2015-03-25_21h02_04

### DIFF
--- a/share/spice/dictionary/definition/content.handlebars
+++ b/share/spice/dictionary/definition/content.handlebars
@@ -1,14 +1,16 @@
 <h3>
-    {{#if plural_form}}
-        <span class="zci__def__word text--primary">{{plural_form}}</span>
-    {{else}}
-        <span class="zci__def__word text--primary">{{word}}</span>
-    {{/if}}
+{{#if plural_form}}
+    <span class="zci__def__word text--primary">{{plural_form}}</span>
     <span class="zci__def__pronunciation"></span>
 </h3>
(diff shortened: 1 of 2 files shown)


jagtalon:
@moollaza nice find! Although I notice that it no longer works on "cats". Could you remove `api_result.length === 1` in this line? https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/dictionary/definition/dictionary_definition.js#L119

moollaza:
@chrismorast noticed we were still putting the phonetic spelling and sound beside the plural form of the word, with the subtitle indicating the singular form. This looks wierd when the plural and singular have very different spellings (geese vs goose) so I've move the phonetics and sound to the subtitle when the queried word is in plural form. Right now we also show the definition for the singular form because WordNik doesn't have good definitions for plurals.


2015-03-26_11h55_43
2015-03-26_11h55_59

jagtalon:
@moollaza I like this idea although I thought it was a bit annoying that the play button is now a little smaller

moollaza:
Spoke with @chrismorast he's happy with this for now. If we find the button is too small we can adjust it.

jagtalon:
@moollaza sweet!

PR #1693, opened 2015-03-26
[Pokedex IA] Amended the IA trigger and the link to next evolution

AkA84:
Original IA PR: #1536 

* The IA is now triggered only by a one-word remainders, such as "pokemon pikachu". If the user searches for "how to train my pokemon", the IA is not triggered
* The IA provides a link to the next evolution of the pokemon, but it wasn't specifically redirecting to the IA tab


jagtalon:
Looks sick! :+1:

PR #1695, opened 2015-03-26
Currency: The image and image border are not properly aligned

jagtalon:
Before:
screen shot 2015-03-26 at 3 39 28 pm

After:
screen shot 2015-03-26 at 3 40 00 pm


moollaza:
LGTM :+1:

PR #1701, opened 2015-03-28
spice/time_and_date_holiday - using text and tile layout based on results

### LINKED ISSUE #1679: TimeAndDate - Holidays: Finish Integration
The awesome team at TimeAndDate had started a Spice Instant Answer that allows for us to search for upcoming Holidays.

I've pushed the code to a branch, [timeanddate/holidays](https://github.com/duckduckgo/zeroclickinfo-spice/tree/timeanddate/holidays) and we now **need someone awesome to finish up the integration**. They've done a good job but we'd like to simplify the IA and use more features of our built-in templates.

More specifically, we'd like this to be done:

- First, switch to the `text` template group
- remove the images (API will no longer provide them anyway)
- Provide a `title` (the date of the Holiday), `subtitle` (name of the holiday) and `description`
- For the better know/international holidays (like Thanksgiving, Christmas, etc) we should show a detail view (no tiles) that prominently displays the date and name of the holiday:
    2015-02-09_14h04_31

- For Holidays that span multiple days, we can use the same view, and update the title to show the date range:

    2015-02-09_14h08_33

- For Holidays/searches with multiple days (like "when is Eid") let's show a tile view (we may need to increase the tile width to display the title, subtitle and description better):

    2015-02-09_14h15_12
 
----

FWIW, this Instant Answer will eventually evolve to use a special Calendar View that we're working on. Then we'll be able to show multiple holidays in a much easier to use interface. Ideally clicking on a holiday would expand to provide more information about it. For now a simplified text view should be good enough!

iambibhas:
Fixes #1679. Haven't worked on the multiple day one as the api result is messy. Also, don't have enough examples to test it on, if I choose to parse the end date from title. Any more examples for multi day event please?

when_is_pi_day_at_duckduckgo_-_2015-03-29_02_02_12
If a future day is available, showing it in the end.

when_is_eid_at_duckduckgo_-_2015-03-29_02_02_34
Need help making the tile wider. For now overriding CSS using `!important`.

when_is_kwanzaa_at_duckduckgo_-_2015-03-29_02_02_57

------

> IA Page: https://duck.co/ia/view/holiday


iambibhas:
Looks like they dont have any concept of start/end date even on their own website http://www.timeanddate.com/holidays/us/kwanzaa The holiday name as the end date embedded. Can't find any other holiday like this though.

jagtalon:
Oh man this is good. Checking it out.

lib/DDG/Spice/TimeAndDateHoliday.pm
```diff
@@ -0,0 +1,50 @@
...
+name "Time and Date holiday search";
+source "timeanddate.com";
+icon_url "/i/www.timeanddate.com.ico";
+description "Search for holidays";
+primary_example_queries "when is halloween", "when is constitution day in kazakhstan";
+category "dates";
+topics "everyday";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/TimeAndDateHoliday.pm";
+attribution email => 'webmaster@timeanddate.com',
+            web => ['http://www.timeanddate.com', 'timeanddate.com'];
```
jagtalon: @iambibhas you can add yourself in here. :)

iambibhas: ^_^ I'll do that.

jagtalon:
@iambibhas 

Looks like the tiles don't have the same size here:

screen shot 2015-03-30 at 3 29 46 pm

jagtalon:
The `basic1` stuff in https://github.com/duckduckgo/zeroclickinfo-spice/pull/1702 should fix that :)

jagtalon:
@iambibhas we should probably move away from the CSS stuff as much as possible. We could possibly put a variant for an expanded text group template instead. :)

iambibhas:
> we should probably move away from the CSS stuff as much as possible.

I didn't want to use CSS. But found no other way to make the tiles wider.

> We could possibly put a variant for an expanded text group template instead.

Can you explain a little? The template doc is a bit confusing to me. The tiles variants(narrow, wide etc) don't seem to work with the `text` group.

jagtalon:
@iambibhas just talked to @chrismorast and @sdougbrown. We don't want to expand the text anymore than it is so the current one (without the CSS) is good. :)

iambibhas:
@jagtalon ok. but the date strings get cut sometimes then

iambibhas:
It looks fine for any month other than Sep, Oct, Nov and Dec. I'm worried that the date might get cut if it's like `Wednesday, September 23, 2015`.

iambibhas:
Can I add `triggers any => 'day';` to this?

moollaza:
> Can I add triggers any => 'day'; to this?

That's likely way too broad a trigger and will result in over triggering. I'd rather we compile a list of holiday names and put them into a text file so we can slurp them in as triggers.

iambibhas:
> I'd rather we compile a list of holiday names and put them into a text file so we can slurp them in as triggers.

That might take some time. There are thousands of them. You want to wait for that before merging this? I could work on that and push it as improvement later. For now "when is foo" works.

moollaza:
>That might take some time. There are thousands of them. 

@iambibhas we don't need to cover *every* holiday, starting with the most popular US/International holidays would be good.

> You want to wait for that before merging this? I could work on that and push it as improvement later. For now "when is foo" works.

We can stick with the simple trigger for now and make a task to build a list of holidays. We could potentially take the list from Wikipedia or timeanddate.com or if there's another resource that lists out the most popular ones that'd be great.

moollaza:
>It looks fine for any month other than Sep, Oct, Nov and Dec. I'm worried that the date might get cut if it's like Wednesday, September 23, 2015.

@iambibhas let's abbreviate the month names, that should fix the width problems :)

iambibhas:
@moollaza how does it look?

moollaza:
@iambibhas looking good!

/cc @abeyang I think this one is about ready for design pass, I'll put it on a machine soon

share/spice/time_and_date_holiday/time_and_date_holiday.js
```diff
@@ -0,0 +1,87 @@
...
+                templates: {
+                    group: 'text',
+                    options: {
+                        content: Spice.time_and_date_holiday.singleday,
+                        moreAt: true
+                    }
+                }
+            });
+        } else {
+            DDG.require('moment.js', function() {
```
moollaza: I feel like we should try to remove the duplication of most of this data and unify it with a single `Spice.add()` call.

```js
var normalize,
    meta,
    ...;

if (... ){
    normalize = function() { ... };
} else {
    normalize = function() { ... };
}

Spice.add({
    id: time_and_date_holiday,
    name: 'Holidays',
    normalize: normalize_fn,
    meta: meta_obj,
    ...
});
```

moollaza: Or alternatively start with an obj and add properties to it accordingly, then pass it to `Spice.add()`

iambibhas: Got it. I'll fix that. Just curious, is there a way to require moment.js as not a block but globally for a frontend? Like a flag at the top that says moment.js is required. And then we can use moment() anywhere in the main function. Creating blocks seem messy at times. The Flights spice was using variables set in `this` object and using that inside `DDG.require()` wont just work. `this` would point to something else inside that require block. So was just curious.

moollaza: Not yet but it's on our radar. We're running into situations where we need both `moment` and `maps`, for example, so we're investigating the best path forward. We're also investigating making much of this open-source, but everything, as you may have noticed so far, is constantly evolving and improving.

iambibhas: Cool. I've made the changes.

moollaza:
@iambibhas I just realized how long the package/callback names are. There's no need to have "TimeAndDate" in them -- lets simplify it: `DDG::Spice::Holiday` should suffuce :)

moollaza:
@iambibhas TimeandDate updated the API for us (as you noticed!):

Example API call: `m=whenis&c=uk&q=summer+bank&callback=func`

Result:

```
func({
     "c": "United Kingdom",
     "q": "summer bank",
     "h": [{
         "n": "Summer Bank Holiday",
         "u": "/holidays/uk/summer-bank-holiday",
         "a": "In England, Wales and Northern Ireland, the summer bank 
holiday is on the last Monday of August. In Scotland it is on the first 
Monday of August. This day marks the end of the summer holidays for many 
people who return to work or school in the autumn.",
         "o": [{
             "d": "Monday, August 3, 2015",
             "s": "Scotland"
         },{
             "d": "Monday, August 31, 2015",
             "s": "<span title=England>ENG</span>, <span 
title=\"Northern Ireland\">NIR</span>, <span title=Wales>WAL</span>"
         }]
     }]
});
``
- c: Name of identified country
- q: Query (without country)
- h: Array containing all matching holidays. Holidays are searched starting 30 days in the past from today and up to 1 year into the future.
	- n: Holiday name
	- u: URL of holiday article, to be prefixed with http://www.timeanddate.com
	- a: Abstract of the article, NULL if no data available.
	- o: Array with all found instances/observances of this holiday
		- d: Holiday instance date
		- s: States/regions observing this holiday, NULL if no states. This text contains HTML.

---

We should strip the HTML from `s` and can ignore it when it's not defined.

File: lib/DDG/Spice/Holiday.pm [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+    # Fallback to US if no country can be determined, that's the
+    # country that has best holiday coverage.
+    } else {
+        ($q, $c) = ($_, 'us');
+    }
+
+    # Kill eventual slashes to avoid misbehaviour of the `spice from'
+    # regular expression.
+    $q =~ s/\///g;
+    $q =~ s/\s*[\d]{4}\s*//g;
```
iambibhas: removing any year mentioned in the query because timeanddate doesn't seem to consider it while searching.

moollaza: Hmm I'm not sure we should simply ignore it -- might be best to ignore the query unless it matches the current year. 

If someone searches for "when is Easter 2017" and we show them a result for "Easter 2015", we're essentially giving them an irrelevant result. It looks like TimeAndDate has data for different years, I'll ask if they can allow us to pass along a year in the API call.

moollaza: In the interested of moving this PR along -- lets just ignore queries with a year for now.

iambibhas: Okk. I'll do that.

iambibhas:
@moollaza ignoring year mentioning queries. Anything else needed to get this going?

moollaza:
>Anything else needed to get this going?

Nope, looks like it's ready for a design pass! I'll put it up on a testing machine.

/cc @abeyang

moollaza:
Live on DDH8: https://ddh8.duckduckgo.com/?q=when+is+easter&ia=holidays

/cc @abeyang

moollaza:
@iambibhas found a bug -- results for easter are mixing 2015 and 2016 dates -- if the date of the event has passed we should obviously show the date for the next occurrence. The problem is your fault though -- the API only gives us 2016 dates for a couple holidays.

I don't think we have any logic in place to check if the date has already passed though?

iambibhas:
@moollaza no. that logic is not in place. I'll put it in place so that the dates that have already passed are ignored.

moollaza:
@iambibhas T&D got back to me with some API updates:

>No year specified -> will query from today until today + 1 year.
Year specified (/scripts/ddg.php?y=2013&....) -> will query from Jan 1 
to Dec 31 in that year.

This means we won't end up mixing results across years and can now support queries that specify a year.

iambibhas:
@moollaza That's awesome! I'll add that to the api call.

iambibhas:
@moollaza 
selection_033

share/spice/holiday/holiday.js
```diff
@@ -0,0 +1,117 @@
+(function (env) {
+	"use strict";
+	env.ddg_spice_holiday = function(api_result) {
+	    if(!api_result || !api_result.h || api_result.h.length === 0) {
+			return Spice.failed('holiday');
```
moollaza: looks like we have some tabs mixed in here. Sadly, I'll need you to remove them ;)

moollaza:
@iambibhas this is looking great! The only bug I noticed is that it doesn't seem to work for past years? Though the API does give us results. If we can get those to work, we should adjust the triggering to allow for "when *was* <holiday> <year>"

moollaza:
@abeyang this one is ready for a final design pass

iambibhas:
@moollaza I think we decided to Not show events before today. I can turn that back on.

share/spice/holiday/holiday.js
iambibhas: Damn. How did this get past my radar?!

abeyang:
Hi @iambibhas , dunno if this is because of your recent change, but "when is easter" no longer triggers the IA, whereas "when is easter 2016" does. I would expect both to have the same results...

iambibhas:
@abeyang that's because the easter has passed. Damn, I think we made a decision to not show dates before today. But that seems to be causing confusions. I'll turn that off.

abeyang:
@iambibhas right, so I'm saying we should do the logical thing, which is to show the **next** Easter.

abeyang:
In other words, if I search for easter, and I get the 2015 result back (which has passed), that's **wrong**.

iambibhas:
@abeyang made some changes - 

selection_036
selection_035

For now is/was are interchangeable. So `when is easter 2015` would work. If we want `when was easter` to return dates from 2015, which are past, might have to work a little on that.

iambibhas:
Made these changes - 

 - for the query `when is easter`, it'll search from today till one year from today
 - for the query `when was easter`, it'll search only in current year, so from 1st jan 15 ro 31st dec 15
 - for any query with a year suffixed, irrespective of is/was, it'll search in that year.
 - also, `what day is easter` will also work. Somebody asked for it - https://twitter.com/WorldsEndless/status/608388266279243778

For some reason, T&D isn't returning anything for father's day in USA. :-/ 
```
ddg_spice_holiday(
{
c: "U.S.A.",
q: "fathers day",
h: [ ]
}
)
```

moollaza:
@iambibhas sorry for the delay here! Thanks a ton for the updates :+1: 

>For some reason, T&D isn't returning anything for father's day in USA. :-/ 

I think it's a spelling thing? I believe when you have `'s` it works?

iambibhas:
> I think it's a spelling thing? I believe when you have 's it works?

I think it was some bug on their end, because now both the queries work.

moollaza:
:+1:

moollaza:
@iambibhas okay I think this is just about done! Only thing is that he design team prefers that in the case of Easter and Christmas (and potentially other well known holidays), they don't want to show the tiles. So I think we'll need to setup a little logic in the Perl (or JS) so that if the query is "easter" it implies "Easter Sunday" and so we should only show the single detail pane for Easter Sunday, instead of the tiles with Easter Sunday, Monday and Good Friday. Likewise Christmas should be Christmas Day, December 25th always.

iambibhas:
@moollaza there is no easy way to do this for all events. So for now handling it in frontend based on a mapping of `query: name of holiday`. E.g.
```
var single_days = {
                'christmas': 'Christmas Day',
                'easter': 'Easter Sunday'
            };
```

More holidays can be added to the list if needed.

jagtalon:
@iambibhas Thanks! Doing a final check then I think it's just about done!

jagtalon:
@iambibhas is this affected by https://github.com/duckduckgo/zeroclickinfo-spice/issues/1927?

moollaza:
> there is no easy way to do this for all events. So for now handling it in frontend based on a mapping of query: name of holiday. E.g.

@iambibhas that's totally fine with me :+1:

iambibhas:
@jagtalon We're just showing what timeanddate returns in this case, and just showing the dates. Probably not affected by that.

moollaza:
@iambibhas one bug: "when is day" returns results. I guess we should make sure the remainder isn't just "day" :)

jagtalon:
Sweet I'd love this to get merged!

iambibhas:
> "when is father's day 2015 in the usa", "when is father's day 2015 in the us", etc need to have the location rewritten as "united states" for the API call to return successfully

The API call works for either `us` or `united states`. The problem is the `the` in the query. `Locale::Country` doesn't recognize `usa`, I can make an alias, but then `code2country('us')` returns `The United States`. I then again have to mess with `the`. Just wondering if it's worth it. Or should I just replace `us` or `usa` with `united states`.

moollaza:
> I then again have to mess with the. Just wondering if it's worth it. Or should I just replace us or usa with united states.

@iambibhas hmm I think ultimately using `Local::Country` will be beneficial because we can also then aliases for other country names.

For now I think it's OK to just alias the US names because they will likely be the most popular

iambibhas:
@moollaza using `Locale::Country` and added tests for those queries.

iambibhas:
@moollaza any idea why I'm getting this error on travis?

```
[12:03:37] t/Holiday.t ................. Undefined subroutine &Locale::Country::alias_code called at
/home/travis/build/duckduckgo/zeroclickinfo-spice/.build/7xEmO1tpLl/blib/lib/DDG/Spice/Holiday.pm line 8.
```

I have `use Locale::Country;` in the code - 

```
use Locale::Country;

Locale::Country::alias_code('usa' => 'us');
```

moollaza:
@iambibhas I think it's because of your alias. We've moved all the aliases to the `duckduckgo/duckduckgo` repo now so they apply everywhere.

Checkout https://github.com/duckduckgo/duckduckgo/blob/27ccc3a86187b27ddab03905bcff56b6802f9dd9/lib/DDG/Meta/CountryCodes.pm#L17. I think you'll need to make a PR to that file to add the alias.

Also, it looks like `alias_code` is no longer valid: https://metacpan.org/pod/distribution/Locale-Codes/lib/Locale/Codes/Changes.pod#Removed-alias_code-function

I think you need to use `add_country_alias`?

iambibhas:
@moollaza I was just following http://perldoc.perl.org/Locale/Country.html

Let me try `add_country_alias()`.

Also, I've been getting these warnings when I run duckpan server for a while - 
```
add_country_alias(): alias already in use: Laos
 at /usr/share/perl/5.18/Locale/Codes/Country.pm line 72.
add_country_alias(): alias already in use: Russia
 at /usr/share/perl/5.18/Locale/Codes/Country.pm line 72.
```

Related to the aliases in duckduckgo repo?

moollaza:
>Also, I've been getting these warnings when I run duckpan server for a while...Related to the aliases in duckduckgo repo?

@iambibhas yes we were getting those in production because we were defining aliases in several places (Spice and Goodies) and loading all the packages (causing duplication), so we decided to declare it all in one place to avoid those warning. It worked in production and we're no longer getting those messages on our dev machines.

I'm still seeing those message on my personal computer though and I just haven't been bugged enough by it to investigate the cause. It's getting pretty annoying though...

moollaza:
@iambibhas any luck here?

iambibhas:
@moollaza removed locale for now. Just replacing us/usa. This alias business is tricky. need to spend more time with it. For now this will work. Added more tests.

moollaza:
@iambibhas thanks!

Will get it on beta soon.

@abeyang I think this one is finally done :smiley:

iambibhas:
Sorry for taking so long. :disappointed:

moollaza:
>Sorry for taking so long. :disappointed:

@iambibhas no need to apologize! It's done now, and looks great. Thanks for seeing it through to the end!

abeyang:
@moollaza can you put this on a dev server for me to check? Not working on mine...

moollaza:
@iambibhas @abeyang up on Beta: https://beta.duckduckgo.com/?q=when+is+easter+in+usa&ia=holidays

abeyang:
Um, when I search for `when is christmas 2016`, this last line on the bottom isn't correct ("the next christmas day...")

screen shot 2015-09-16 at 10 00 53 am

share/spice/holiday/holiday.js
```diff
@@ -0,0 +1,123 @@
...
+                template_obj = {
+                    group: 'text',
+                    detail: false,
+                    variants: {
+                        tile: 'basic1',
+                        tileTitle: '2line'
+                    }
+                };
+            }
+
```
moollaza: @iambibhas I think we need to check here if we have a result to show, and in the case that there's nothing use `$.getScript()` to make a call to the Spice endpoint for holidays, but force USA as the country (because it has the best coverage).

moollaza: Right now, being in Canada, when I search [when is christmas](https://beta.duckduckgo.com/?q=when+is+christmas&ia=iatesting) I'm not getting any results because the only result is "christmas" instead of "christmas day".

moollaza:
@abeyang @iambibhas the Time and Date team hasn't gotten back to me. In the interest of moving this along, I think we should drop the part that says when the next occurrence is? I don't think it's necessary anymore because we now default to showing the next occurrence anyways.

share/spice/holiday/holiday.js
iambibhas: Hmm. lemme check if I can do that.

moollaza:
Deploying updates to Beta. @iambibhas please let us know if it's ready for review!

marcantonio:
I have some concerns about this api:

screen shot 2016-01-30 at 4 50 13 pm

Also, we should just bail on the query in this case.

Sorry if this is noise, I'm just trying to figure out what needs to be done to get it across the finish line.

moollaza:
> Sorry if this is noise, I'm just trying to figure out what needs to be done to get it across the finish line.

@marcantonio not at all! It looks like Time&Date has a funny way of handling Mardi Gras they seem to call it shrove Tuesday? http://www.timeanddate.com/holidays/us/shrove-tuesday

We could probably add an exception to the code for that. I'm in contact with their API team (they've worked with us on a few other IAs) so they could probably improve the APIs handling of that query. I think most other major holidays work?

marcantonio:
@moollaza sounds good, never heard of Shrove Tuesday  :).  I'll put an exception in for now.

I think it looks pretty good.  I honestly don't like displaying tiles ever, re: this comment: https://github.com/duckduckgo/zeroclickinfo-spice/pull/1701#issuecomment-129551643.  I don't think there's an easy way to filter out the right one though.

Are you aware of any bugs?  Looking through the comments it looks like all were addressed.  I'll keep poking at it.

sekhavati:
@moollaza @marcantonio In the UK Shrove Tuesday is also commonly known as Pancake day, you might want to add that to your list of exceptions as well.

PR #1702, opened 2015-03-29
github_jobs - using moment.js and text tiles

### LINKED ISSUE #1633: GitHubJobs: Convert to Text template, use Moment.js for dates
https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/github_jobs/github_jobs.js

No need for this one to be using a Base template. It should use the text template and a footer similar to the Reddit and HackerNews Spices.

Also, we could use Moment.js (like Reddit & HN) to handle the dates!

https://duck.co/ia/view/github_jobs

iambibhas:
Fixes #1633.


share/spice/github_jobs/github_jobs.js
```diff
@@ -3,6 +3,8 @@
 
     env.ddg_spice_github_jobs = function(api_result) {
 
+        console.log(api_result);
```
jagtalon: Please remove the logs. :)

share/spice/github_jobs/github_jobs.js
```diff
@@ -16,24 +18,40 @@
...
+                    sourceName: 'GitHub',
+                    sourceIcon: true,
+                    itemType: 'Jobs'
+                },
+                normalize: function(item) {
+                    return {
+                        title: item.title,
+                        subtitle: item.company,
+                        location: item.location,
+                        description: $('<div/>').html(DDG.strip_html(item.description)).text(),
```
jagtalon: Why the wrapping and unwrapping of the text in a `div` element?

iambibhas: `item.description` has a lot of encoded html entities like `&quot;`. Passing only that text in `description` doesn't seem to be decoded when displaying the tiles. I also don't like this line. Any saner way of doing it?

share/spice/github_jobs/github_jobs.js
iambibhas: Yikes. Sorry. Removing it. :)

share/spice/github_jobs/github_jobs.js
```diff
@@ -16,24 +18,40 @@
...
-                detail: false,
-                item_detail: false
-            }
+                templates: {
+                    group: 'text',
+                    options: {
+                        footer: Spice.github_jobs.footer
+                    },
+                    variants: {
+                        tile: 'basic',
```
jagtalon: Change this to `basic1` so that the tiles will have the same sizes:

Before:
screen shot 2015-03-30 at 11 29 17 am

After:
screen shot 2015-03-30 at 11 58 29 am

If you want it bigger, you could add `tileSnippet: 'large'` in the `variants` object:

screen shot 2015-03-30 at 12 00 48 pm

Thanks to @sdougbrown for pointing these things out to me.

jagtalon: @abeyang which would you prefer? The big tile or the small one?

jagtalon: I personally like the big one.

iambibhas: @jagtalon seriously need to document these things. :(

iambibhas: `tileSnippet: 'large'` puts a bit more space above the footer than needed I think.

jagtalon: @iambibhas I agree. I wasn't aware of it either so we're def working on that.

jagtalon:
@iambibhas thanks--only thing we need is a design signoff from @abeyang

iambibhas:
@jagtalon okk. Here is the final UI for reference

screenshot from 2015-03-30 22 13 52

share/spice/github_jobs/github_jobs.js
jagtalon: Ah I see what you mean. Prob should be handled by DDG.strip_html in the future

jagtalon:
@iambibhas 

- @abeyang said that `tileSnippet: 'large'` would be better.
- Since we're using a built-in template, no need to have the CSS anymore, I think.
- Would put the type of job in the footer:
screen shot 2015-03-30 at 2 22 37 pm
- No need for the `GitHubJobs_formatDate` helper so we can delete that.
- Footer elements can overflow so best to add the `one-line` class.
screen shot 2015-03-30 at 2 25 34 pm

iambibhas:
@jagtalon made the changes.

jagtalon:
@iambibhas could you make it a hyphen `-` instead? Sorry, I didn't actually say what character that is. :)

iambibhas:
@jagtalon sure. I just saw that character being used at some other spices (e.g. github), hence used that.

abeyang:
@iambibhas I'm looking at this right now with different positions and cities -- it's coming along nicely! A few comments/questions:

1. Can we swap **Position/Location** with **Recency**? Eg, "9 hours ago" should be the last line.
2. You know, all the examples I'm seeing is full time. So can we do this: by default, just show **Location**, eg:

> San Francisco, CA

If it's part time / contract / otherwise, we can append it to the end. So, only for those examples, it would say:

> San Francisco, CA - Part Time

Hope that makes sense!

iambibhas:
@abeyang I think we should mention even if it's full time. Mentioning nothing might cause confusion if, let's say, there are 10 jobs, all of them are full time and so none of them mentions that. The user might think that we dont show that information at all and start clicking everything to know which one is part time etc.

iambibhas:
Maybe we should mention the job type beside the company instead of the location.

abeyang:
@iambibhas good point about full-time. That got me thinking about an overall better design -- one you might appreciate. :smile: 

In the meantime, do you think you can pull me a list of all possible job statuses? Full-time, part-time, ..?

abeyang:
It seems like they only have three job statuses. What do you think about this design? Should be doable, yeah? Moreover, rather than using "hours ago" or "days ago", can we have it display **NEW** for postings that are < 3 days old, and display the actual date ("Apr 2") for all other ones?

FYI, we're thinking of using this pattern for the other job IAs as well.

jobs-v1 1

iambibhas:
That looks good! Lemme try and apply that, without those colored bullets, which you can implement with CSS, right? Also, which template to use to wrap the title like that? 

Can you guys publish the new doc? Even if it's half baked? Put it on Readthedocs or somewhere maybe and keep working on it?

iambibhas:
I think you'll have to push a new template for it. Otherwise I'll have to use CSS for the footer.

moollaza:
@iambibhas You can get the title to wrap using the new variants. https://github.com/duckduckgo/zeroclickinfo-spice/blob/9ffde54d9d666f1d3b0bb4e9338605c07390cfad/share/spice/product_hunt/product_hunt.js#L50 There's a few more live examples in the repo. You can use `titleTitle: '2line'` here. Likewise, `tileFooter: '2line'`.

>Otherwise I'll have to use CSS for the footer.

That's fine, please use a `<span>` and CSS to fill the color. The Transit IA's do the same thing: https://github.com/duckduckgo/zeroclickinfo-spice/search?utf8=%E2%9C%93&q=septa+badge

>Can you guys publish the new doc? Even if it's half baked? Put it on Readthedocs or somewhere maybe and keep working on it?

Docs are being written and updated right now. There's a few PR's in the duckduckgo-documentation repo right now :). Docs for the new `variants` property are coming soon as well.

iambibhas:
@moollaza `titleTitle: '2line'` works but `tileTitle: '2line'` doesn't? I thought it was a typo!

moollaza:
@iambibhas `titleTitle` was a type -- `tileTitle` definitely works -- do you have your dev tools open? When the height hits a threshold we change to mobil viewing and so `2line` forces a single-line. Pop your dev tools out into a separate window and you should be able to see that everything works as expected.

iambibhas:
@moollaza don't have the dev tool open. My laptop screen res is 1366x768. If I zoom out with `Ctrl` + `-` couple of times, then the title wraps in 2 lines. Otherwise it's one line in normal zoom level.

moollaza:
@iambibhas sounds like it's working then. We try to minimize the height of IA's when working with smaller screen sizes so I believe that's the expected behaviour.

iambibhas:
But my laptop screen is not small! It's a 15inch display. It should show wrapped title.

jagtalon:
2-lines work for sure: https://vid.me/oMe8.

@iambibhas what do you think about @abeyang's mockup? You can find a similar design in https://duckduckgo.com/?q=next+train+from+strafford+to+ardmore&ia=septa (SEPTA).

share/spice/github_jobs/github_jobs.js
moollaza: It actually used to work that way but the functionality was removed for some reason. It needs to be brought back for this exact reason though. We're getting encoded entities in Last.FM (there's an asana task for that @jag)

iambibhas:
@jagtalon even for your case, the title doesn't wrap to 2 lines unless gone to full screen mode. I'm just curious why? What's the threshold? Who is that 2lines wrapping targetting? Hugeass full-HD/4k displays? Sorry, just want to be clear about it.

Also, I already implemented @abeyang's design. Lemme push it.

iambibhas:
This is how it looks now - 



The colors of the badge will change for other job types, which I can't find any live example of.

jagtalon:
Fixed by @iambibhas

jagtalon:
Ugh, sorry made a mistake

iambibhas:
not sure why the build failed.

moollaza:
>not sure why the build failed.

We added in trailing spaces and I accidentally merged something with trailing space. Will fix

abeyang:
Hi @iambibhas , not sure what the status of this is. Is this ready for another design pass? If what I'm seeing is the latest, then we need to make sure subtitles go directly underneath titles -- esp. for 1-line titles:

web_developer_jobs_san_francisco_at_duckduckgo

Also, let's give the "dot" a bit more breathing room:

```css
.tile--github_jobs .github_jobs__badge {
  margin: 0 5px 0 0; /* modifying your existing css */
}
```

moollaza:
Hey guy -- sorry for the late news but today I deprecated the GitHub Jobs Spice. We now have the Indeed Spice merged and it will cover all our jobs queries. I had planned to keep the GitHub Jobs Spice around, with a reduced triggers space (so you don't see two jobs tabs for every job query) but after looking at it again, it will be really difficult to reduce the trigger space to "github jobs".

i.e. searching "perl github jobs in sanfrancisco" is awkward and similarly as a `startend` trigger it doesn't work well, because it would be difficult to parse, e.g. "github jobs perl toronto" or "github jobs perl in toronto"

When we later support multiple sources for IA's we'll bring this back. @iambibhas your efforts haven't been wasted though -- I'll port the new design to the Indeed Spice and then we can launch it live, looking great.

Again really sorry about that, I wasn't expecting to kill the IA while it was still being developed.

moollaza:
>then we need to make sure subtitles go directly underneath titles -- esp. for 1-line titles:

@abeyang that's an internal fix we'll need to make if that's how we want the subtitles to act. Our defaults CSS forces a `height` for the title, forcing the `subtitle` to remain in that spot.

iambibhas:
@moollaza oh. ok. cool then. Thanks for taking the UI into consideration. :) Let's close this one and the issue.

jagtalon:
@iambibhas sorry about all the trouble that you went through. :(

iambibhas:
@jagtalon no worries. I don't mind. It was a learning experience. :-)

abeyang:
Like the others, I'm sorry this happened, @iambibhas . But I ditto what @moollaza said -- this wasn't in vain. We're going to incorporate this design into the Indeed IA -- and then some! So thanks for all your hard work in this -- it's very much appreciated!

iambibhas:
Thanks guys. :smile:

PR #1703, opened 2015-03-30
Added trigger words for Japanese, Chinese and Korean

tagawa:
It's just the translation of "weather":
天気 => Japanese, Traditional Chinese
天气 => Simplified Chinese
날씨 => Korean

The IA result itself is not localized but the symbols and temperatures are easily understandable.

### DIFF
--- a/lib/DDG/Spice/Forecast.pm
+++ b/lib/DDG/Spice/Forecast.pm
@@ -3,6 +3,7 @@ package DDG::Spice::Forecast;
 use strict;
 use DDG::Spice;
 use Text::Trim;
+use utf8;
 
 name "Forecast";
 description "Weather forecast";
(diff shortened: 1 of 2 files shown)


moollaza:
@tagawa thanks a lot for this PR, we really appreciate it. As you can imagine, internationalization is a very important component of a search engine because we have users around the world. Right now almost all of our Instant Answers use English triggers and provide English results. We plan to eventually translate both our triggers and results, but for right now we're going to stick with English triggers. Our aim is to find a method that will work for all IAs rather than adding triggers sparsely to various IA's which will create an inconsistent experience. (e.g. why does forecast trigger in English, Korean and Chinese but nothing else does?)

So for now I'm going to close this, but please understand that this is something we're currently working towards.

Thanks again for taking the time to contribute!

tagawa:
Thanks for the explanation @moollaza - I fully understand. I've worked in internationalization previously and know it can be very complex so I appreciate the task ahead of you.

Another plan I have is to create Japanese-only IAs using Japanese web APIs and services, for example a local earthquake report IA. Would that be OK?

PR #1704, opened 2015-03-30
Sharedir

jdorweiler:
Adds sharedir to dist.ini so we build the `share/dist/` directory. 
@moollaza want to do a new release after this is merged?


moollaza:
LGTM :+1:

PR #1709, opened 2015-03-31
Time: force return if capital doesn't exist

moollaza:
Right now the code falls through if the capitals lookup fails which still causes the IA to `return ""` creating an API call with no param.

This fixes that and cleans up the code a bit.

/cc @jagtalon @tommytommytommy


jagtalon:
Looks good :+1:

PR #1714, opened 2015-04-01
Time - changed regex handling to trigger on time, current time etc

### LINKED ISSUE #1641: Time: "time", "local time", "current local time", etc should show user's time
This IA could definitely be improved to show the user's local time for certain queries, such as:

- "time"
- "local time"
- "current local time"
- "current time"
- "what's the time"
- etc.

https://duck.co/ia/view/time

iambibhas:
Fixes #1641. time, current time, current local time etc now returns the user's local time(which is phoenixville in the tests).

### DIFF
--- a/lib/DDG/Spice/Time.pm
+++ b/lib/DDG/Spice/Time.pm
@@ -22,21 +22,23 @@ triggers any => "time";
 
 my $capitals = Load(scalar share("capitals.yml")->slurp);
 
-my $place_connector = join '|', qw(in of for at);
-
 handle query_lc => sub {
     my $q = shift;
 
-    return unless $q =~ m/^(what'?s?|is|the|current|local|\s)*time(?:is|it|\s)*(?:\b$place_connector\b)\s+(?<loc>[^\?]+)[\?]?$/;
-    $q = $+{loc};
-    trim($q);
-    $q =~ s/,//g;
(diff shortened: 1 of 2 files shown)


lib/DDG/Spice/Time.pm
```diff
@@ -22,21 +22,23 @@ triggers any => "time";
...
-my $place_connector = join '|', qw(in of for at);
-
 handle query_lc => sub {
     my $q = shift;
 
-    return unless $q =~ m/^(what'?s?|is|the|current|local|\s)*time(?:is|it|\s)*(?:\b$place_connector\b)\s+(?<loc>[^\?]+)[\?]?$/;
-    $q = $+{loc};
-    trim($q);
-    $q =~ s/,//g;
+    $q =~ m/(?<rest>what'?s?|is|the|current|local|\s)*time(?:is|it|in|of|for|at|\s)*(?<loc>[^\?]*)[\?]*$/;
```
jagtalon: You need to keep the `$place_connector` because queries such as `what time is it in india` don't work anymore (please put that in the test as well).

iambibhas: I kept all the pieces of `place_connector`. The spice only triggers for the capitals. The test is in place for that query -

    'what time is it in kingston?'             => test_spice(@kingston_town),

I think it'll need to be triggered for any other places by the deep triggering.

jagtalon:
Oh I see :+1: 

3zfiv

Issue #248, opened 2013-10-11
RubyGems IA - Lots of useful info from API not being used


moollaza:
There's tons of great info for each list item coming through the API such as:

- # of downloads
- date last update
- link to homepage (though this will be linked from the rubygems.com link we send the user too
- Gem version

https://duck.co/ia/view/ruby_gems

GregorioMartinez:
Thanks for the link!
I added the exact match and download #'s to the footer like Packagist has.
Did you have any suggestions for displaying the additional information?

jagtalon:
Fixed by @iambibhas

PR #1715, opened 2015-04-01
ruby_gems - showing more info in the tiles

iambibhas:
Related to #248. 

Showing more info than just the name and description -


dogweather:
:+1: Helps the user decide which to investigate further.

moollaza:
@iambibhas this is great, thanks!

My only thought is that we should flip the downloads and license so that when the license isn't there we don't have blank space below the downloads.

/cc @abeyang -- this one is already live, we're just adding more info to the tiles

iambibhas:
@moollaza already tried that, didn't help, as I've set `tileFooter: '2line'`. So if one div is absent, another goes to the top.

moollaza:
@iambibhas nice fix. A slightly more graceful approach would be to set the `tileFooter` to `1line` when we have no Licenses? I think that will have the same effect, but I'm not sure.

iambibhas:
@moollaza how to do that from inside the normalize function?

moollaza:
@iambibhas my bad, I thought we could set the variant on a per-tile basis but it's not possible. We'll stick with your solution.

abeyang:
Looks like it's getting there. How are we shortening the description? Is it by number of words? It seems to cut off a bit short.. can we increase it by just a bit?

pasted_image_4_2_15__10_01_am

Moreover, can we add the following css changes? Please add the proper namespace to it as well. :smile: 

```css
.tile__title {
  /* pull title up a bit */
  margin-top: -0.2em
}

.tile__sub {
  /* push desc down a bit */
  margin-bottom: 1em;
}

.has-foot .tile__content--lg {
  /* bring footer up a bit */
  height: 6.4em;
}
```

Finally, rather than `Licenses: MIT`, can we say `MIT License`? (Unless there can be multiple licenses...)

iambibhas:
@abeyang I think the truncation is because of the use of `tileSnippet: 'large'`. I'm passing the whole text, but it's showing only that much.

The `licences` part of the API Response is an Array. So there is a chance that there might be more than one. But I'm yet to see any such example. 

I'll make the CSS changes and show you.

abeyang:
@iambibhas you're right -- we'll work on the truncation on our end. 

Style changes are looking good! Can you add the class `tx-clr--grey-light` to the **subtitle** and the **download** lines? In the end, it should look something like this:

http://cl.ly/image/2Y3u381X1S2m

iambibhas:
@abeyang I'm currently sending the subtitle from the normalize function. Not sure if I can add custom class to it. @moollaza @jagtalon can I?

moollaza:
@iambibhas no it's not currently possible to add classes to those. We'll soon be able to specify colors and font-sizes from the `variants` obj but it's not yet implemented.

@abeyang can we implement the color changes later?

abeyang:
@moollaza yup that's fine -- we'll just wait until we can apply it to all templates like this one.

jagtalon:
@moollaza @abeyang looks like there's a bug with our template. Missing subtitles can make a tile small:

screen shot 2015-04-08 at 1 37 03 am

I'll make an internal issue for it.

jagtalon:
@iambibhas `sourceUrl` gets double-encoded which leads to links such as https://rubygems.org/search?utf8=%E2%9C%93&query=coffee%2520script.

screen shot 2015-04-08 at 1 51 24 am

share/spice/ruby_gems/ruby_gems.js
```diff
@@ -26,13 +26,42 @@
...
+            sort_default: 'downloads',
             normalize : function(item){
+                var licenses = [];
+
+                if (item.licenses) {
+                    var keys = Object.keys(item.licenses);
+
+                    for (var i = 0; i < keys.length; i++) {
+                        licenses.push(item.licenses[keys[i]]);
+                    };
```
jagtalon: Looks like `licenses` is already an array in the result:

screen shot 2015-04-08 at 1 56 12 am

So all we need is:

```javascript
var licenses = item.licenses || [];
```

jagtalon:
@abeyang since this is a built-in template, I'm a bit hesitant with using CSS. Do we need the CSS for all text templates since we're applying it here?

share/spice/ruby_gems/ruby_gems.js
iambibhas: When I last checked, it was a json object and not an array. Lemme check again.

iambibhas: Huh. Looks like they changed the format since the last commit I made here. I'll update the spice.

jagtalon:
@iambibhas got contacted by @abeyang. The CSS should go internally so you can `git rm` it from this repo. Other than that, if the issues above have been resolved, this should be good to go!

moollaza:
>@moollaza @abeyang looks like there's a bug with our template. Missing subtitles can make a tile small:

For now, please set the subtitle to `&nbsp` (`''` or `' '` might work too?) when it doesn't exist. We need to make sure they're not undef so the template accounts for them. We'll fix this internally so this doesn't happen but that'll get this moving forward :)

iambibhas:
@moollaza done.

sdougbrown:
FYI, if you'd like to adjust the number of characters displayed in the snippet, you can use the new `snippetChars` property like so:
```
    meta: {
        ...
        snippetChars: 150
    }
```

Feel free to experiment with this to get some longer descriptions, if you like. :smile:

jagtalon:
Noice. :+1: 

giphy

PR #1716, opened 2015-04-02
SimilarSites - Replaced IDs with classes (fixes #1691)

### LINKED ISSUE #1691: SimilarSites: Move away from using IDs
We often use classes in IAs and rarely IDs. Reference DOM using classes instead. 

https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/similar_sites/content.handlebars#L10

tagawa:
Also removed duplicate hide/expand (+/-) icons.

### DIFF
--- a/share/spice/similar_sites/similar_sites.css
+++ b/share/spice/similar_sites/similar_sites.css
@@ -1,7 +1,3 @@
-.zci--similar_sites .hidden, .zci--similar_sites .chomp--link__ls {
-    display: none;
-}
-
 .zci--similar_sites img {
     vertical-align: middle;
 }
--- a/share/spice/similar_sites/similar_sites.css
+++ b/share/spice/similar_sites/similar_sites.css
@@ -1,20 +1,16 @@
-.zci--similar_sites #hidden {
+.zci--similar_sites .hidden, .zci--similar_sites .chomp--link__ls {
     display: none;
 }
 
-.zci--similar_sites #show_more:hover {
-    cursor: pointer;
-}
-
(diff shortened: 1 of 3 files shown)


share/spice/similar_sites/similar_sites.css
```diff
@@ -1,20 +1,16 @@
-.zci--similar_sites #hidden {
+.zci--similar_sites .hidden, .zci--similar_sites .chomp--link__ls {
```
moollaza: I think we may already have `.hidden` set to `display: none` in our base CSS?

tagawa: There's no ```.hidden``` in the current versions of <a href="https://duckduckgo.com/s956.css">s956.css</a> or <a href="https://duckduckgo.com/t956.css">t956.css</a>. Maybe in a to-be-released version?

moollaza: sorry, our class name is actually, `is-hidden` -- you should be able to use that.

tagawa:
As well as using ```.is-hidden```, was also able to remove a couple of JavaScript toggles thanks to existing behaviour with ```.is-expanded```.

jagtalon:
:+1:

tagawa:
Cool, thanks for the merge.

jagtalon:
@tagawa no problem! I found there's a small bug that I missed but I made a fix in #1741

tagawa:
Ah, well spotted. Thanks.

PR #1718, opened 2015-04-03
Timer - Replaced .hidden class with existing .is-hidden

tagawa:


share/spice/timer/timer.handlebars
```diff
@@ -1,4 +1,4 @@
-<div id='timer_container' hidden>
+<div id='timer_container is-hidden'>
```
jagtalon: looks like `is-hidden` should be in a `class` attribute?

tagawa: You're quite right &mdash; thanks for spotting that.

mattr555:
:+1: thanks!

PR #1697, opened 2015-03-26
spice/deb_version - added the spice on top of master

iambibhas:
Reviving old stale branch `ideas/deb`. Added on top of master, as mentioned in #1667. Should close that other PR.

PR #1721, opened 2015-04-03
deb_version - added metadata

iambibhas:
Related to #1697.

### DIFF
--- a/lib/DDG/Spice/DebVersion.pm
+++ b/lib/DDG/Spice/DebVersion.pm
@@ -6,15 +6,15 @@ use DDG::Spice;
 name "Debian Versions";
 source "Debian Sources";
 icon_url "https://sources.debian.net/static/favicon.ico";
-description "Get the version of a package in various debian repositories"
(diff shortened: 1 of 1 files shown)


lib/DDG/Spice/DebVersion.pm
```diff
@@ -1,9 +1,18 @@
-# Get the version of a package in various debian repositories
 package DDG::Spice::DebVersion;
```
javathunderman: If you could put line 1 under the package declaration with the ABSTRACT declaration comment, it should be ok...

iambibhas: @javathunderman I added that string to the `description` metadata. Should I add it again as ABSTRACT?

moollaza: It wouldn't hurt to set an ABSTRACT, but we actually use the `description` (or will be soon) when the IA Pages (duck.co/ia) are further along.

moollaza:
LGTM :+1:

PR #1726, opened 2015-04-03
Currency - use of k and cardinals instead of zeros

### LINKED ISSUE #1563: Conversions: Support for queries like "10 million usd to php"
https://duckduckgo.com/?q=10%2C000%2C000+usd+to+php&ia=currency works but this query doesn't. Similar ones are "10k usd to jpy"

https://duck.co/ia/view/currency

iambibhas:
Fixes #1563. Now supporting queries like `convert 4k euro to usd`.

### DIFF
--- a/lib/DDG/Spice/Currency.pm
+++ b/lib/DDG/Spice/Currency.pm
@@ -40,7 +40,7 @@ my $question_prefix = qr/(?:convert|what (?:is|are|does)|how (?:much|many) (?:is
 my $number_re = number_style_regex();
 
 # This regexp is responsible for actually processing the query and capturing the important parts.
-my $guard = qr/^$question_prefix($number_re*k?)\s?(hundred|thousand|million|billion|trillion)?\s?($currency_qr)(?:s)?(?:$into_qr|$vs_qr|\s)?($number_re*)\s?($currency_qr)?(?:s)?\??$/i;
+my $guard = qr/^$question_prefix($number_re*)\s?(hundred|thousand|k|million|m|billion|b|trillion)?\s?($currency_qr)(?:s)?(?:$into_qr|$vs_qr|\s)?($number_re*)\s?($currency_qr)?(?:s)?\??$/i;
 
 triggers query_lc => qr/$currency_qr/;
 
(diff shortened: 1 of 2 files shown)


iambibhas:
Should I replace that `Mid-Market Rates: 2015-04-03 at 16:00 UTC` string with `Mid_market Rates Updated 10 minutes ago` or similar using moment?

iambibhas:
Hmm. Seems like I can't use Switch.

moollaza:
>Should I replace that Mid-Market Rates: 2015-04-03 at 16:00 UTC string with Mid_market Rates Updated 10 minutes ago or similar using moment?

I think we should leave it for now.

lib/DDG/Spice/Currency.pm
```diff
@@ -113,13 +113,28 @@ sub checkCurrencyCode {
...
     return $normalized_number, $from, $to;
 }
 
 handle query_lc => sub {
     if(/$guard/) {
-        my ($amount, $from, $alt_amount, $to) = ($1, $2, $3, $4 || '');
+        my ($amount, $cardinal, $from, $alt_amount, $to) = ($1, $2 || '', $3, $4, $5 || '');
+
+        if ($amount =~ m/k$/) {
+            $amount =~ s/(\,|k)//g;
```
jagtalon: You should let the methods in `DDG::SpiceRole::NumberStyler` process the number for you instead of removing the commas yourself. See what happens with the query "convert 4.000.000k euro to usd":

screen shot 2015-04-08 at 1 21 15 am
It tries to remove commas, but the number doesn't have any because it's in the European format.

lib/DDG/Spice/Currency.pm
```diff
@@ -113,13 +113,28 @@ sub checkCurrencyCode {
...
+            $amount *= 1000;
+        }
+
+        if ($cardinal ne '') {
+            $amount =~ s/\,//g;
```
jagtalon: Same here with https://github.com/duckduckgo/zeroclickinfo-spice/pull/1726/files#r27943704

lib/DDG/Spice/Currency.pm
jagtalon: Support "m" and "b" as well for million and billion?

iambibhas: Ouch. Lemme fix that.

iambibhas:
@jagtalon just changed few things, also added test for your example. Seems to be working.

jagtalon:
@iambibhas nice! Thanks for the tests. :+1:

PR #1731, opened 2015-04-04
Sublime Text Package IA

MrChrisW:
# Sublime Text Package IA :package: :page_facing_up:

Search for information about Sublime Text packages

https://packagecontrol.io

+ sublimetext package code
+ sublime text package php

Developers, Sublime Text Lovers

**Live Example: http://duckpan3.mrchris.me/?q=sublime+text+package+code&ia=software**

selection_320

## Checklist

```
[X] Added metadata and attribution information (https://duck.co/duckduckhack/metadata)
[X] Wrote test file and added to t/ directory (https://duck.co/duckduckhack/goodie_tests)

    - Windows 8
        [X] Google Chrome
        [X] Firefox

    - Windows 7
        [X] Google Chrome
        [X] Firefox
        [X] IE 8

    - Windows XP
        [X] IE 8

	- Linux
        [X] Chromium
        [X] Firefox
```

------
> IA Page: https://duck.co/ia/view/sublime_packages

### DIFF
--- a/lib/DDG/Spice/SublimePackages.pm
+++ b/lib/DDG/Spice/SublimePackages.pm
@@ -18,23 +18,15 @@ attribution github => ["MrChrisW", "Chris Wilson"],
 spice to => 'https://packagecontrol.io/search/$1.json';
 spice wrap_jsonp_callback => 1;
 
-my @triggers = ("sublimetext packages", "sublimetext package", "sublime text packages", "sublime text package", "sublime text");
-my $triggers = join '|', @triggers;
-
-triggers startend => @triggers;
+triggers startend => "sublimetext package", "sublimetext packages", "sublime text package", "sublime text packages", "sublime text";
 
 my $skip = join "|", share('skipwords.txt')->slurp(chomp => 1);
 
-handle query_lc => sub {
-	my $query = $_;
-	# If the trigger doesn't contain package(s) the trigger is "sublime text" 
-	if($query !~ m/packages?/i) {
-		# Do not trigger IA if query matches any words in skipwords.txt file
-		return if ($query =~ m/$skip/i)
-	}
-	s/$triggers//g; # Remove triggers
-	s/^\s+|\s+$//g; # Trim
-    return unless $_;   
+handle remainder => sub {
+    return unless $_; 
+    # Do not trigger IA if query matches any words in skipwords.txt file
+    return if  m/$skip/i;  
     return $_;
 };
+
 1;
--- a/lib/DDG/Spice/SublimePackages.pm
+++ b/lib/DDG/Spice/SublimePackages.pm
@@ -1,0 +1,28 @@
+package DDG::Spice::SublimePackages;
+# ABSTRACT: Search for information about Sublime Text packages
+
+use strict;
+use DDG::Spice;
+
+name "SublimePackages";
+source "packagecontrol.io";
+icon_url "/i/packagecontrol.io.ico";
+description "Display Sublime Text packages";
+primary_example_queries "sublimetext package code", "sublime text package php";
+category "software";
+topics "programming", "computing";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SublimePackages.pm";
+attribution github => ["MrChrisW", "Chris Wilson"],
+            web => ["http://chrisjwilson.com", "Chris Wilson"];
+
+spice to => 'https://packagecontrol.io/search/$1.json';
+spice wrap_jsonp_callback => 1;
+
+triggers startend => "sublimetext package", "sublime text package", "sublime text packages";
+
+handle remainder => sub {
+    return unless $_;   
+    return $_;
+};
+
+1;
--- a/lib/DDG/Spice/SublimePackages.pm
+++ b/lib/DDG/Spice/SublimePackages.pm
@@ -18,7 +18,7 @@ attribution github => ["MrChrisW", "Chris Wilson"],
 spice to => 'https://packagecontrol.io/search/$1.json';
 spice wrap_jsonp_callback => 1;
 
-triggers startend => "sublimetext package", "sublime text package", "sublime text packages";
+triggers startend => "sublimetext package", "sublime text package", "sublime text packages", "sublime text";
 
 handle remainder => sub {
     return unless $_;   
--- a/lib/DDG/Spice/SublimePackages.pm
+++ b/lib/DDG/Spice/SublimePackages.pm
@@ -18,11 +18,23 @@ attribution github => ["MrChrisW", "Chris Wilson"],
 spice to => 'https://packagecontrol.io/search/$1.json';
 spice wrap_jsonp_callback => 1;
 
-triggers startend => "sublimetext package", "sublime text package", "sublime text packages", "sublime text";
+my @triggers = ("sublimetext packages", "sublimetext package", "sublime text packages", "sublime text package", "sublime text");
+my $triggers = join '|', @triggers;
 
-handle remainder => sub {
+triggers startend => @triggers;
+
+my $skip = join "|", share('skipwords.txt')->slurp(chomp => 1);
+
+handle query_lc => sub {
+	my $query = $_;
+	# If the trigger doesn't contain package(s) the trigger is "sublime text" 
+	if($query !~ m/packages?/i) {
+		# Do not trigger IA if query matches any words in skipwords.txt file
+		return if ($query =~ m/$skip/i)
+	}
+	s/$triggers//g; # Remove triggers
+	s/^\s+|\s+$//g; # Trim
     return unless $_;   
     return $_;
 };
-
 1;
--- a/share/spice/sublime_packages/footer.handlebars
+++ b/share/spice/sublime_packages/footer.handlebars
@@ -1,4 +1,4 @@
 <div class="tile__update__info one-line">
-  <i class="star starIcon"></i>
+  <i class="ddgsi ddgsi-arrow-down"></i>
   {{unique_installs}} &middot; Updated {{last_modified}}
 </div>
--- a/share/spice/sublime_packages/footer.handlebars
+++ b/share/spice/sublime_packages/footer.handlebars
@@ -1,4 +1,4 @@
 <div class="tile__update__info one-line">
-  <i class="ddgsi ddgsi-arrow-down"></i>
-  {{unique_installs}} &middot; Updated {{last_modified}}
+	<i class="ddgsi ddgsi-arrow-down"></i>
+	{{unique_installs}} &middot; Updated {{last_modified}}
 </div>
--- a/share/spice/sublime_packages/footer.handlebars
+++ b/share/spice/sublime_packages/footer.handlebars
@@ -1,0 +1,4 @@
+<div class="tile__update__info one-line">
+  <i class="star starIcon"></i>
+  {{unique_installs}} &middot; Updated {{last_modified}}
+</div>
--- a/share/spice/sublime_packages/sublime_packages.js
+++ b/share/spice/sublime_packages/sublime_packages.js
@@ -55,8 +55,7 @@
                             { required: 'highlighted_authors' },
                             { required: 'last_modified' },
                             { required: 'highlighted_description' },
-                            { required: 'unique_installs' },
-                            { key: 'name' }
+                            { required: 'unique_installs' }
                         ]
                     },
                 });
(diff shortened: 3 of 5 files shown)


MrChrisW:
I'd like to add relevancy block checking for the description however the returned value for no description is "No description provided" so I don't believe a relevancy block can be used.

MrChrisW:
Ping, @wbond do you mind if we use your awesome site (packagecontrol.io) as an endpoint for a new DuckDuckGo Instant Answer? You can see an example above, linking back to your site of course :smile:

moollaza:
> I'd like to add relevancy block checking for the description however the returned value for no description is "No description provided" so I don't believe a relevancy block can be used. 

@MrChrisW I think you can use a regex check to make sure it matches something that isn't that string?

Alternatively, you can can check if the description is equal to that string and if so `return null` from inside the `normalize()` function. It'll then skip that item.

dogweather:
A big :+1: as a sublime text user.

MrChrisW:
@dogweather :+1: :smile: 

>I think you can use a regex check to make sure it matches something that isn't that string?

@moollaza You can use the `match` key in the relevancy block to ensure the returned field matches, however you cannot use the logic, `match =! "No description provided"` 

I'll use a conditional block in the `normalize` function :+1:

wbond:
Yes, I am ok with this integration. It looks like you discovered the JSON API for the site. It isn't well documented anywhere, I think just mentioned in a blog post, but all of the data-driven URLs on the site can have `.json` appended for an API.

In terms of the pagination, if you open an issue at https://github.com/wbond/packagecontrol.io, I should be able to add an option for increasing the number. I'm not sure I'd want to do "all" since that could return thousands of items, but more than 25 should be reasonable.

MrChrisW:
@wbond That's great news, thank for taking the time to respond! :smile:

>I should be able to add an option for increasing the number. I'm not sure I'd want to do "all" since that could return thousands of items, but more than 25 should be reasonable.

Ok understood, I'll wait and see what kind of item limit we want, 25 may actually be alright. In any case I'll lodge an issue if more is required. :+1:

Dwaligon:
@MrChrisW This is awesome! :+1:

share/spice/sublime_packages/sublime_packages.js
```diff
@@ -0,0 +1,55 @@
...
+
+        DDG.require('moment.js', function() {
+                Spice.add({
+                    id: "sublime_packages",
+                    name: "Software",
+                    data: api_result.packages,
+                    meta: {
+                        itemType: "Sublime Text Packages",
+                        sourceName: "packagecontrol.io",
+                        sourceUrl: 'http://packagecontrol.io' + api_result.name
```
jagtalon: Looks like `api_result.name` can be `undefined`:

screen shot 2015-04-08 at 12 54 24 am

share/spice/sublime_packages/sublime_packages.js
```diff
@@ -0,0 +1,55 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_sublime_packages = function(api_result){
+
+        if (!api_result || api_result.error || api_result.packages.length == 0) {
```
jagtalon: Check if `api_result.packages` exists first before trying to access `length`.

jagtalon: Does `api_result.error` actually come up?

jagtalon: Nit: Use `===`, but it's more of a personal choice. :)

lib/DDG/Spice/SublimePackages.pm
```diff
@@ -0,0 +1,28 @@
...
+category "software";
+topics "programming", "computing";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SublimePackages.pm";
+attribution github => ["MrChrisW", "Chris Wilson"],
+            web => ["http://chrisjwilson.com", "Chris Wilson"];
+
+spice to => 'https://packagecontrol.io/search/$1.json';
+spice wrap_jsonp_callback => 1;
+
+triggers startend => "sublimetext package", "sublime text package", "sublime text packages";
```
jagtalon: Add "sublime text package manager"?

jagtalon: Also I see myself just typing in "sublime text perl". See if you might need to use `DDG.isRelevant` or the relevancy block for that.

share/spice/sublime_packages/footer.handlebars
```diff
@@ -0,0 +1,4 @@
+<div class="tile__update__info one-line">
+  <i class="ddgsi ddgsi-arrow-down"></i>
+  {{unique_installs}} &middot; Updated {{last_modified}}
```
jagtalon: Are `unique_installs` and `last_modified` always available?

jagtalon: Maybe best to check them just to be on the safe side--especially since this is an undocumented API.

MrChrisW: Yes.

share/spice/sublime_packages/sublime_packages.js
MrChrisW: Updated.

lib/DDG/Spice/SublimePackages.pm
jagtalon: Maybe "package control" as well.

screen shot 2015-04-08 at 1 05 44 am

MrChrisW: @jagtalon Add "sublime text package manager" as a full trigger e.g 
`sublime text package manager <keyword>`

>Also I see myself just typing in "sublime text perl". See if you might need to use DDG.isRelevant or the relevancy block for that.

Sure, I was keeping the triggers limited to start with, happy to add sublime text as a trigger and ensure relevancy.

jagtalon:
This is awesome, btw :package: :package: :package:

share/spice/sublime_packages/footer.handlebars
MrChrisW: @jagtalon 
>Maybe best to check them just to be on the safe side--especially since this is an undocumented API. 

What would you suggest as the best way of doing this? 
+ Using a relevancy block requiring those keys
+ Wrapping the whole return in a conditional statement checking these values exist

jagtalon: @MrChrisW Relevancy block might be cleaner but I'm ok with both.

MrChrisW:
@jagtalon Updated. 

I'm using a [skipwords.txt](https://github.com/MrChrisW/zeroclickinfo-spice/blob/SublimePackages/share/spice/sublime_packages/skipwords.txt) file to avoid triggering on queries like [`sublime text buy`](http://duckpan3.mrchris.me/?q=sublime+text+buy). The API does [return](http://duckpan3.mrchris.me/js/spice/sublime_packages/buy) results for some skip words.

I've tried using a `key` within the `relevancy` block **see** [ 039aa31a059fea262c065902fa876b3cb9e5bad0](https://github.com/MrChrisW/zeroclickinfo-spice/commit/039aa31a059fea262c065902fa876b3cb9e5bad0#diff-484e5be0e234625b176364a46221ca6dR59) to tighten triggering further however this resulted in queries like `php` or `code` failing most likley due to the `DDG.isRelevant()` function being too strict and not finding any "exact" match or something like that... Please review when you're ready :bow: :smile:

moollaza:
>  Pagination: The API accepts page=2 etc to display more packages, I'm not sure how to pull more then the default 25 packages as there's no page=all. Maybe multiple requests?

@MrChrisW if you set `meta.next` in your `Spice.add()` obj, we'll create a "Load more" button when you reach the end of the carousel which will call that endpoint. It's assumed that the URL given will have a callback parameter though, so I'm not sure if it will work here.

I think 25 packages is OK, but if you can get that^ to work easily, I see no harm.

>Exact matches: Use a different template to display an exact package match? 

I've been contemplating this for a while -- how to handle exact matches for these kinds of IA's -- you can set `exactMatch` to true for a given result inside `normalize` (or `boost` it for a close match) so it appears first in the list of items. I think for now this is the best way to handle things. 

Details: https://duck.co/duckduckhack/spice_displaying#exactmatch-codebooleancode-amp-boost-codebooleancode

MrChrisW:
>if you set meta.next in your Spice.add() obj, we'll create a "Load more" button when you reach the end of the carousel which will call that endpoint. It's assumed that the URL given will have a callback parameter though, so I'm not sure if it will work here.

@moollaza After a some research it appears `meta.next` is undocumented and has only ever been used on internal Spices that utilise a DDG proxy API endpoint (e.g m.js, i.js, v.js) these return a `next` value that contains the parameter `pg` for page number.

I attempted to replicate this behaviour by doing the following
+ Accept page number parameter on Perl backend - https://gist.github.com/MrChrisW/1fa650cc0c5fafc87e43
+ Set `meta.next` value - https://gist.github.com/MrChrisW/44a2b98315ef6015bd6d

This did load the "next page" via the `meta.next` value above when I reached the end of the carousel -- looking in the console. However no additional tiles were actually displayed :frowning: I'm sure it has something to do with the JSON data being returned missing a value needed for loading more? Just a guess..

>I've been contemplating this for a while -- how to handle exact matches for these kinds of IA's -- you can set exactMatch to true for a given result inside normalize (or boost it for a close match) so it appears first in the list of items. I think for now this is the best way to handle things. 

Good to hear you've been thinking about 'exact matches', TBH I dislike the design when an exact match is found using tiles, why show a tile if there's only a single match? I doesn't seem right.  I'd prefer displaying something like the Python fathead/text template. I've seen this done with the [Whois](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/whois/whois.js) Spice however it's in a hacky way. Maybe this can be improved internally in the future. I'll leave it as is for now. :+1: 

selection_324
selection_325

jagtalon:
@MrChrisW thanks! I guess we can just get back to those issues. Made an IA page for it: https://duck.co/ia/view/sublime_packages. Off to design review!

moollaza:
@abeyang @jagtalon looks like this is ready for a design signoff?

moollaza:
@MrChrisW apologies for the delay! I think this is just about done! We've added in some new icons now that should be used here: `ddgsi-download` and `ddgsi-clock` for the footer.

moollaza:
LGTM :+1: I'll update the icons :)

moollaza:
@MrChrisW as always, thanks a ton for making this!

MrChrisW:
@moollaza as always, thanks for your feedback!

moollaza:
@MrChrisW congrats! It's live: https://duckduckgo.com/?q=deal+4&t=ffab&ia=answer

MrChrisW:
@moollaza I think you need more sleep :stuck_out_tongue_winking_eye: 
~~congrats! It's live: https://duckduckgo.com/?q=deal+4&t=ffab&ia=answer~~
**FTFY** https://duckduckgo.com/?q=sublime+text+package+code&ia=software

bendtherules:
I think the trigger word is too big for most people, I generally just search with the first word st/st2/st3. When I am feeling generous, might also add the term "package" after it, but "sublime text package" is too big.

To understand whether this is a real issue, tell any ST user that this feature exists but you wont give them the trigger word. Look at how many tries they neeed to finally find this out.

bendtherules:
Also, looking at the project description would be really great, if possible later.

So that "sublime text package jsbeautifier" would return "jsformat" directly, which is the first organic search result. Usually, most people dont know the exact name but can guess a few close ones.

MrChrisW:
@bendtherules Thanks for your input. To avoid digging up the original merge request you're welcome to lodge a [new issue](https://github.com/duckduckgo/zeroclickinfo-spice/issues/new)

>generally just search with the first word st/st2/st3.

There's currently a PR to improve some aspects of the triggering #1953

>So that "sublime text package jsbeautifier" would return "jsformat" directly, which is the first organic search result. Usually, most people dont know the exact name but can guess a few close ones.

That's a little difficult to implement as the direct API returns no results for `jsbeautifier` as it doesn't search the package description, short of creating a hack to transfer the keyword `jsbeautifier` to `jsformat` there's no easy fix for this.

The query's `sublime text package javascript formatting` or `sublime text package javascript format` does return jsformat as the first result.

wbond:
Just a note of clarification, the search API does search the title and the description, but does not currently search the readme.

MrChrisW:
@wbond Ahh thanks for clarifying!

PR #1733, opened 2015-04-05
Twinword's Word Map sorceurl, SSL connection, and icon changed.

twinword:
Twinword's Word Map instant answer has been modified with SSL connection, favicon, and sourceurl changes. 

twinword_duckduckgo

### DIFF
--- a/lib/DDG/Spice/WordMap.pm
+++ b/lib/DDG/Spice/WordMap.pm
@@ -11,39 +11,40 @@ primary_example_queries "similar to sound", "words like sound";
 category 	"reference";
 topics 		"words_and_games", "everyday";
 code_url 	"https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/WordMap.pm";
+# Current icon should be change with new one at https://www.twinword.com/favicon.ico
 icon_url 	"/i/www.twinword.com.ico";
-attribution web 	=> ['http://twinword.com','Twinword Inc.'],
-            twitter => ['http://twitter.com/levelpump', 'Twinword Inc.'],
+attribution web 	=> ['https://www.twinword.com','Twinword Inc.'],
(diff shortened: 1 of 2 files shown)


lib/DDG/Spice/WordMap.pm
```diff
@@ -12,38 +12,38 @@ category 	"reference";
...
 code_url 	"https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/WordMap.pm";
 icon_url 	"/i/www.twinword.com.ico";
-attribution web 	=> ['http://twinword.com','Twinword Inc.'],
-            twitter => ['http://twitter.com/levelpump', 'Twinword Inc.'],
+attribution web 	=> ['https://www.twinword.com','Twinword Inc.'],
+            twitter => ['https://twitter.com/twinwordinc', 'Twinword Inc.'],
             email 	=> ['feedback@twinword.com','Twinword Inc.'];
 
-spice to => 'http://duckduckgo.twinword.com/api/v2/context/user/duckduckgo/?entry=$1&api_key={{ENV{DDG_SPICE_WORDMAP_APIKEY}}}';
+spice to => 'https://www.twinword.com/api/v2/context/user/duckduckgo/?entry=$1&api_key={{ENV{DDG_SPICE_WORDMAP_APIKEY}}}';
```
moollaza: If possible we prefer `http` over `https`. We're proxying the request for the client through an HTTPS connection with DuckDuckGo so using `http` is a little faster.

twinword: I know your point though, Twinword's API policy is forcing to use SSL with all partners. To my understanding DDG has good cache system. so, I'd like to stick with https. Is it possible?

moollaza: Sure it's no problem.

PR #1736, opened 2015-04-07
Flights - integrating new design

### LINKED ISSUE #1636: Flights: Implement new design
We've recently re-designed our Flights Tiles as you can see below:

screen shot 2015-02-27 at 10 43 02 am

This new design now needs to be implemented. Any takers?!

https://duck.co/ia/view/flights_route

iambibhas:
Work in progress. Fixes #1636.

Tried on both my handsets, one with 4.7inch(left) another 5.5inch(right) screen.

 

TBD:

 - wrap the whole thing up in moment.js and use that.
 - The Landed/On Time etc texts are output from a helper function. Not sure how to provide color to the upper divider based on that. Should I use JS/jQuery?
 - The column widths are a bit shaky. I seem to have found a balance. But needs reviewing by a designer.
 - Need some help with the "disabled" view. That also depends on the helper.

-----
IA Page: https://duck.co/ia/view/route


moollaza:
>The Landed/On Time etc texts are output from a helper function. Not sure how to provide color to the upper divider based on that. Should I use JS/jQuery?

Yup. I'd create a JS variable that tells us if it's delayed or not. Then use that to add a "delayed" class to the highest level div (when applicable) and then have some CSS that targets the correct elements to change their colors:

```css
.zci--flights .delayed .departure__scheduled {
    color: #de5833
}
```

Don't forget, you can also use sub-templates if it helps, or even use different templates based on the state.

iambibhas:
@moollaza is there a way to get the div element of the item from inside the `normalize()` or `Spice.add()` functions?

moollaza:
>is there a way to get the div element of the item from inside the normalize() or Spice.add() functions?

@iambibhas no, this is all running before rendering occurs so the DOM doesn't exist at the point in time. What are you trying to do?

javathunderman:
@moollaza ping!

moollaza:
@javathunderman thanks, I'm still waiting to hear from @iambibhas about what he needs help with :)

iambibhas:
Sorry for the delay. I need couple of days to settle down. I'll update this on Sunday.

moollaza:
It's no problem @iambibhas take your time :)

iambibhas:
This is what it looks like right now. 

selection_022

The only thing missing from the proposed UI is the color for delayed flights. Need to make a demo JSON and test with that.

moollaza:
@iambibhas :+1:  I would just hardcode the value to make sure that a delayed flight has the right colors. Is this ready for review then?

moollaza:
BTW, this is looking awesome!

javathunderman:
@moollaza @iambibhas I think this is ready for review- is there anything else to address?

iambibhas:
I also think this is ready for review. I can fix bugs if there is any.

iambibhas:
Oh! Just noticed that it works for flights in India too!

moollaza:
@iambibhas I'm seeing terminals for some flights on the site, http://www.flightstats.com/go/FlightStatus/flightStatusByFlight.do?id=554039153&airline=EY&flightNumber=8767&departureDate=2015-06-10

iambibhas:
@moollaza I'll check that out. Etihad is an international carrier I think. Maybe that's why.

moollaza:
@iambibhas sorry for the delay. This is looking great! One thing I noticed is that the "On Time" flights have a filled green "status" bar. I believe these should be either empty, or partially filled based on the progress of the flight

moollaza:
@iambibhas regarding:

>TBD:
>
>- wrap the whole thing up in moment.js and use that.
>- The Landed/On Time etc texts are output from a helper function. Not sure how to provide color to the upper divider based on that. Should I use JS/jQuery?
>- The column widths are a bit shaky. I seem to have found a balance. But needs reviewing by a designer.
>- Need some help with the "disabled" view. That also depends on the helper.


It seems like all of these have been addressed?

iambibhas:
@moollaza can't add `elClass` for just a single tile, can I? Not sure how to change the tile background from the frontend then. Should I just use the `.parent()` of `.tile_heading` or something?

moollaza:
>can't add elClass for just a single tile, can I? 

Crap you're right.

>Not sure how to change the tile background from the frontend then. 

The first think I can think of is setting a `data` attribute on one of the tiles, or a css class and using jQuery to find it. Then target the parent element (or search for the closest `.tile`) and add a css class to that to change the background.

Let me check with the team if there's a better way though...

moollaza:
We're going to investigate a way to specify `elClass` on a per item basis but creating it in the `normalize` function.

If you want, for now I think the method I mentioned, using some special class on an element inside the tile, to find the tile with jQuery *should* work?

iambibhas:
This is how it looks now - 

screenshot from 2015-08-11 23 04 55

share/spice/flights/route/flights_route.js
```diff
@@ -181,24 +183,50 @@
...
-                                terminal: (flight[i].airportResources && flight[i].airportResources.arrivalTerminal) || "<span class='na'>N/A</span>",
-                                gate: (flight[i].airportResources && flight[i].airportResources.arrivalGate) || "<span class='na'>N/A</span>",
+                                airport: arriving_airport,
+                                location: arriving_location,
+                                terminal: (flight[i].airportResources && flight[i].airportResources.arrivalTerminal) || "N/A",
+                                gate: (flight[i].airportResources && flight[i].airportResources.arrivalGate) || "N/A",
                                 isDeparted: false
                             };
 
+                        var is_on_time = this.onTime(flight[i], departureDate, arrivalDate);
```
moollaza: Looks like you're missing the `scheduledDeparture` and `scheduledArrival` here?

moollaza:
@iambibhas we merged in a small change to the content template today because we realized we were incorrectly handling delayed flights. There's a conflict now, but your changes are necessary for your PR so it should be fairly easy to resolved the conflict.

Can you please pull in the upstream changes so I can deploy this to a testing machine?

share/spice/flights/route/content.handlebars
```diff
@@ -1,27 +1,56 @@
+{{#if has_landed}}
+<div class="tile__landed" style="display: none"></div>
```
moollaza: what about adding this class to the div below?


```html
<div class="tile__heading {{#if has_landed}}tile__landed{{/if}}>
```

moollaza:
@iambibhas this is looking awesome!

Regarding the CSS, now that we haver classes for font and background colours, we should be using those as the classes are built to work with the current DDG theme.

Checkotu https://duckduckgo.com/styleguide#txt-n-color for all the colors and classes we have. This should also greatly reduce the amount of CSS we have :)

moollaza:
@iambibhas live here: https://beta.duckduckgo.com/?q=Jetblue+Boston+to+Los+Angeles&ia=route

moollaza:
@iambibhas I took a stab at the CSS because it looks like it's causing problems when we add the date beside the scheduled time:

```css
.tile--route  {
    width: 36em;
    padding: 1.44em;
}

...

.zci--route .tile__departure > div,
.zci--route .tile__arrival > div {
    display: inline-block;
}

.zci--route .tile__city {
    width: 12em;
    margin-right: 1em
}

.zci--route .tile__scheduled {
    min-width: 6em;
    margin-right: 1em
}

.zci--route .tile__gate {
    width: 3em;
    margin-right: 1em
}

.zci--route .tile__terminal {
    width: 5em;
}
``

I also removed the grid sizing classes (`thirty`, `twenty`, `ten`, ...) from the Handlebars.

@thom @chrismorast I think this is ready for a design review.

I think we may need to reconsider how we show the arrival date because the current approach (indicating the arrival date when it's after departure) means the tile widths changes and so we have extra whitespace:

![air_canada_toronto_to_london_at_duckduckgo_and_1____projects_zeroclickinfo-spice__perl_](https://cloud.githubusercontent.com/assets/873785/9557414/d0328bde-4da9-11e5-834c-278780dcb0d4.png)

File: share/spice/flights/route/flights_route.js [hunk 3]
```diff
@@ -209,79 +237,96 @@
...
+                        if (!item.is_on_time) {
+                            status_text = 'Scheduled';
+                            arrival_datetime = moment(item.scheduledArrivalDate);
+                        } else {
+                            if (moment(item.arrivalDate).isBefore(moment())) {
+                                status_text = 'Arrived';
+                                arrival_datetime = moment(item.arrivalDate);
+                            } else {
+                                status_text = 'Arrives';
+                                arrival_datetime = moment(item.scheduledArrivalDate);
```
moollaza: Looks like there's a bug here? I'm seeing a landed flight that says "Arrives in 9 hours"

qatar_airways_doha_to_islamabad_at_duckduckgo

thm:
@moollaza Nice! Here's some quick feedback:
1. Both rows currently say "Departure" where the second row should be "Arrival" :wink: 
2. All the columns (departure, scheduled, gate and terminal) were intended to have a fluid width rather than a fixed width. So the width of the entire tile adapts based on its content. That also addresses your concern with the city/state/country lines potentially taking up too much space
3. Can you decrease the font-size of the column labels (departure, scheduled, etc) to 7px tall instead of the 9px it is right now?
4. Can you use an `&ndash;` (–) inbetween the status and estimated arrival rather than the short `&dash;` (-) dash?
5. I'm missing the dots that indicate the flight's current position. Green is on-time, red is delayed. Note that the dot for the delayed flight is actually behind on the timeline:
screen shot 2015-08-29 at 12 14 24 am
6. The alignments are slightly off, here's the exact numbers measured from the ascender height and the baseline of the text:
screen shot 2015-08-29 at 12 13 44 am

cc @chrismorast

jagtalon:
Oh man this looks sooo good.

moollaza:
@iambibhas are you able implement Thom's feedback?

Regarding (2) I think the best way to handle this will be to ditch the grid system and switch to a `<table>` or use inline-block `<div>`'s as columns so the width of the rows for each column are the same and adjust to the wider content. Does that make senes?

iambibhas:
@moollaza I may need some help with CSS on this. Can't figure our how to visualize the dot in this case. I'll try to make the rest of the changes.

moollaza:
> I may need some help with CSS on this. Can't figure our how to visualize the dot in this case. I'll try to make the rest of the changes.

What part do you need help on? I would use CSS for the circle, or your could use this: • and increase the font size. You'll need to position the circle with JS. I think if you can get everything else figured out we can probably merge this in and we can tackle the circle if you can't

iambibhas:
@moollaza your CSS changes make it hard to implement @thm's changes. The numbers are also a bit different. The initial draft said the border padding should be 1.44em/18px, but @thm is saying that it should be 20px. Same with the divider margins that @thm says that should be 16px.



@thm 7px is too small for those column headers 

selection_008

Can you guys create an unified spec for all the numbers?

@moollaza need help in positioning the `•`. Not sure how to put it in the middle of a line.

iambibhas:
Also, not sure how to make the inner columns fluid. Some expert hint could be helpful.

thm:
My bad, I didn't know a spec sheet was already created for this! I think all details outlined there look good, except for the padding. I think the vertical padding should be equal between all dividers and text, as outlined in my previous comment (16px for all vertical padding, 20px on both sides of the tile). @chrismorast

Regarding the __7px font__, I didn't mean the `font-size`, but rather the height of the text. In the screenshot above that seems to be 5px, can you increase the font-size by 2px so that the text is 7px tall?

Regarding the __dot__, can you make that a `span` or `div` (or something similar) with a set width and height so we can ensure a pixel-perfect dot? I'm concerned the `•` may become blurry. The width and height would both be 6px.

moollaza:
>@moollaza your CSS changes make it hard to implement @thm's changes. 

Yup you'll have to scrap my changes if we're going with a fluid tile. You could use a table with 4 columns and 2-4 rows (depending how you organize the header, city name and horizontal divider) or use 4 divs to create columns.

Either way, the departure and arrival elements need to created as columns so they can stretch and maintain the same width.

@thm I just realized that fluid-width tiles will work on mobile, so how should we handle that? Wrap gate + terminal below Departure/Arrival and Schedules?

thm:
@moollaza Yeah, I think that would work. I'd be curious to know why it won't work on mobile though, because the carousel view remains right? We would have to think about properly truncating the values when the tile has a fixed width too then.

moollaza:
@iambibhas ping! Just checking in on this. It looks like there are conflicts right now. Do you need any help on this?

moollaza:
This is being continued in #2301

PR #1737, opened 2015-04-07
TransitNJT: Convert to YAML (partially fixes #1728)

### LINKED ISSUE #1728: Transit: Use a YAML file for stations
Instead of a text file, using YAML will change our long iteration into a short hash lookup. The appropriate changes can be modeled using [this commit](https://github.com/tagawa/zeroclickinfo-spice/commit/f0b7dd10c57bf5f1c1b462bc4ce85a01db985c3d).

https://duck.co/ia/view/transit_njt
https://duck.co/ia/view/transit_path

tagawa:
Also added a few stops from https://en.wikipedia.org/wiki/List_of_New_Jersey_Transit_stations

If this looks OK I'll do the same for the PATH IA.


mattr555:
Hi @tagawa, I should have mentioned that the current list of supported stops is sourced from http://njt-api.appspot.com/njt/stops (my own API). I think I also forgot to update this file; the API doesn't do the light rail stations anymore (the schedule was too regular to really be useful). Could you update the yaml file with the new stations? Thanks!

tagawa:
Thanks for the link - didn't realise it was your API. I've added one or two versions of existing names (e.g. secaucus) that seemed to return positive values. As the IA now uses a hash key lookup I thought it more efficient to have several variants in the YAML file rather than regex matching in the code.

mattr555:
Hi @tagawa, sorry for the delay! School has been pretty busy :books:

On further review, I'm hesitant to merge this because we're being a little bit strict with the matching. I now understand why I originally wrote it this way: the [API does a better job of matching the stations](https://github.com/mattr555/njt-api/blob/master/api.py#L5-29) because it knows the routes and can disambiguate a stop name based on that.

As a result, queries like [next train from metropark to secaucus] work without specifying the upper or lower level of secaucus, and queries like [next train from metropark to new york] work even without defining "new york" in the DDG list. "new york" was matched in "new york penn station," so the API could work out that "new york" equals "new york penn station."

Sorry, I was nearsighted when I created the issue and focused on the (probably negligible) speed instead of the usability and practicality of holding a file with many substitutions. Thanks a lot for your contribution anyway :grinning:

tagawa:
Hi @mattr555. No problem and thanks for the explanation. I agree with your summary that it's OK to leave this to the backend API. The good news is I don't need to look at the PATH IA!

PR #1741, opened 2015-04-08
similar_sites.css: Add additional qualifier to avoid extra spacing.

jagtalon:
Removing the container `div` in #1716 made the CSS selectors too broad which added margins everywhere.

Now:
screen shot 2015-04-07 at 11 45 10 pm

Before:
screen shot 2015-04-07 at 11 44 58 pm

Notice the increased space at the bottom.

### DIFF
--- a/share/spice/similar_sites/similar_sites.css
+++ b/share/spice/similar_sites/similar_sites.css
@@ -2,11 +2,11 @@
     vertical-align: middle;
 }
 
-.zci--similar_sites div {
+.zci--similar_sites .zci__body div {
     margin-bottom: 5px;
 }
 
-.zci--similar_sites div > a {
+.zci--similar_sites .zci__body div > a {
     font-size: 1.357em;
     font-weight: 300;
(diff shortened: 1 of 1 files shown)


moollaza:
@jagtalon LGTM but it looks like there's a much larger problem in production -- we're duplicating the + sign somehow... 

https://duckduckgo.com/?q=similar+sites+to+duckduckgo.com&ia=similarsites
2015-04-08_21h28_51

jagtalon:
@moollaza will fix that

jagtalon:
@moollaza Ok no longer seeing that. Could you check? https://vid.me/TNsj

moollaza:
@jagtalon I never saw it on DuckPAN but now that it's in production I'm seeing it. I just realized this isn't using the list template. If we switch to that, it will simplify things and take care of all the hiding/showing logic and the layout.

moollaza:
I'll try that out.

PR #1448, opened 2015-01-25
MassOnTime: Convert to use a map #708

laouji:
This is a pull request relevant to the discussion in duckduckgo/zeroclickinfo-spice#708
~~It is not yet ready to be merged.~~

After many iterations, I think we've got a final product :smile: 
### Modifications made:
- [x] Switched from a Tile View using custom templates to the Map View
- [x] Added new map markers and adjusted the style to match the GetEvents ia [#1744](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1744) 
### ~~Things to work on:~~

~~\- [x] Marker icons (since the API does not return images)~~
~~\- [x] Use the event starttime and endtime in a valuable way (since some results are limited time events)~~

07_chrome

PR #1744, opened 2015-04-10
GetEvents: Update for newest mockup

moollaza:
Final adjustments to meet mockup specs.

/cc @uriger @abeyang 

2015-04-10_00h06_21


moollaza:
Live on DDH8: https://ddh8.duckduckgo.com/?q=events+near+me

jagtalon:
@moollaza Why is the time backwards for some? I see "8pm-7pm" which is impossible unless you're a time traveller!

giphy 1

moollaza:
@jagtalon I didn't come across any events where the times were flipped, but I have checks to make sure the end `isAfter()` the start. I noticed a bug in the API that it's giving the incorrect timezones. Perhaps in that case one time had the correct TZ? I will be emailing the GetEvents team to let them know.

jagtalon:
@moollaza sorry I forgot to post the screenshot

screen shot 2015-04-10 at 9 13 58 am

Check out the first one. Will post the API call / result in a bit.

jagtalon:
@moollaza API result: https://gist.github.com/jagtalon/1c025d0ed78d7f97e0c9 for https://api.getevents.co/ddg?lat=40.1246&lng=-75.5385&timezone=America%2FNew_York

abeyang:
Hey, I'm gonna break my comments down into multiple parts here.

First, I see some unescaped html for a few tiles. Rare, but it's there...

pasted_image_4_10_15__10_49_am

abeyang:
Second, let's talk about **titles**. After looking at @moollaza 's version, let's make these changes:

On default view (no maps), let's make the title go the full-width. Right now, there's that awkward gap; let's close it.

When an event is selected, is it possible to fit the full title inside the tooltip? This is because there are enough examples where the entire title doesn't fit in the tile -- would be nice to show the full event name somewhere.

pasted_image_4_10_15__10_53_am

abeyang:
Third, **minor color / font-size adjustments**.

For the *subtitle*, can we make the font-size 13px (currently 12.5px) and same light gray as the footer (`.tx-clr--grey`, or `#999`).

For the *snippet* in the backside:

1. Let's use `.tx-clr--slate-light` (or `#666`). 
2. I also think we need to add ellipses in there, as most of the descriptions are cut short.
3. We need to shift it down a little bit--

``` css
.tile__title {
  margin-bottom: 0.85em;
}
```

abeyang:
Fourth (and hopefully, finally), a few rare cases where the address goes beyond 4 lines:

pasted_image_4_10_15__11_17_am

Not sure what the absolute best way to deal with this, but for now, maybe we can add a limit to the height of the div, and cut everything *above the fold* out -- if possible?

moollaza:
@abeyang I believe I've addressed everything except the Map Pin text -- that's out of my control. The text is condensed internally so modifying the width via CSS has no effect. We'll have to fix that internally.

moollaza:
Updated on https://ddh8.duckduckgo.com/?q=events+near+me&ia=localevents

abeyang:
Almost there, @moollaza !

For the subtitle, can we add the class to make the color `#999`? (It's currently `#888`.. so subtle, but it bothers me.) :frowning: 

Finally, just need to tweak the backside a bit:

``` css
/* raise the attribution footer higher */
.tile__foot {
  height: 2.7em;
}

/* move the backside snippet up */
.tile--get_events__alt .tile__title {
  margin-bottom: 0.6em;
}
```

You'll notice that the favicon should line up with the rest of the footers at the end:

screen shot 2015-04-22 at 9 52 53 am

abeyang:
@moollaza I just realized while looking at #1448, that the attribution link/footer may be a template thing... maybe we can try to fix that internally. But if that takes longer, then can you just make the changes here first?

moollaza:
>For the subtitle, can we add the class to make the color #999? (It's currently #888.. so subtle, but it bothers me.

It actually has `tx-clr--grey` applied to it -- our default CSS is overiding the colour though, which sucks.

For the footer -- I'm assuming you want `1em` of padding around the tile? Right now the internal CSS doesn't actually set padding/margin for the footer, it sets a height based on how many lines (which subsequently creates space below the text).

The 1.4 line height on the tile-front address pushes the text up from the bottom of the div though, so aligning it perfectly requires custom CSS if you want the line-height adjusted.

abeyang:
@sdougbrown can you address this issue raised by Zaahir:

> It actually has tx-clr--grey applied to it -- our default CSS is overiding the colour though, which sucks.

Any idea why our default css is still overriding the .tx class?

abeyang:
^ But @moollaza , it seems like this took effect for Mass on Time, based on the same [recommendations](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1448#issuecomment-93865809).. what gives?

moollaza:
@abeyang I don't believe it's working for MassOnTime -- we specify `tx-clr--grey` before `tile__sub` in our CSS which is why the default `tile__sub` colour is applied.

moollaza:
Updated: https://ddh8.duckduckgo.com/?q=events+near+me&ia=localevents

Merging. We can make and final tweaks after.

PR #1745, opened 2015-04-10
SimilarSites: convert to list group, simplify

### LINKED ISSUE #1770: Similar site search has an extra show more plus icon (+)
Searching for similar sites works pretty well, but there's an extra plus sign (+) by the show more link. See screenshot: 
screenshot from 2015-04-15 12 24 16

moollaza:
Now that we have the List template, I've switched this over and reduce a lot of the code.

Fixes #1770 

/cc @jagtalon @abeyang 

@abeyang shall I remove the borders? This is how it looks without any added styling.

2015-04-10_10h17_51


moollaza:
@abeyang I opted to bring back the original design and removed the top-margin from the `<ul>`:

2015-04-10_10h28_29

moollaza:
Merging so we can fix the production bug, can tweak design later if needed, but this should be identical to existing IA.

PR #1747, opened 2015-04-11
Maps::Places - not triggering for 'local time', added test

### LINKED ISSUE #1740: Maps::Places- Shouldn't trigger on "local time"
A check needs to be added so that it doesn't trigger on the query "local time".

https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Maps/Places.pm#L13

iambibhas:
Fixes #1740. Added test.


moollaza:
LGTM :+1:

PR #1748, opened 2015-04-11
Revert "Guard against trailing whitespaces in Perl files when building"

pfirsichbluete:
Being overly strict with trailing whitespaces was confusing for devs
according to

  https://github.com/duckduckgo/zeroclickinfo-spice/commit/5980b02618d1f96e0d9e43af1b4ad788d392447e#commitcomment-10675310

. Hence, reverting.

This reverts commit 5980b02618d1f96e0d9e43af1b4ad788d392447e.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -57,6 +57,7 @@ version_regexp = ^(.+)$
 [PodSyntaxTests]
 [GithubMeta]
 [Test::EOL]
+trailing_whitespace = 0
 [@Git]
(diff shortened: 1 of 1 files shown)


moollaza:
LGTM :+1:

moollaza:
@pfirsichbluete I think using something like HoundCI, or asking devs to run JSHint/PerlTidy may be a better alternative.

PR #1749, opened 2015-04-12
Book: Fix image display issue - #1556

### LINKED ISSUE #1556: Book Reviews: Broken in mobile layouts
screen shot 2015-02-19 at 3 59 14 pm

https://duck.co/ia/view/book

MrChrisW:
Attempt at fixing #1556

shot

menu_330
menu_331


moollaza:
@MrChrisW nicely done! I'm very happy about all the removed CSS :+1: 

This is definitely a step in the right direction, but we can make this easier for ourselves if we switch to the `text` template group (I see no reason why it's using the info group, unless the image was intended to be bigger?)

Wit the `text` group we have the ability to specify `subtitle_content: Spice.books.subtitle` template, which we can use to build the subtitle.

That way we can also simplify the `content` template and remove the container `div` that adds a redundant `zci--books` class (we already have that class above in the DOM).

share/spice/book/book.handlebars
```diff
@@ -1,10 +1,20 @@
 <div class="zci--book">
```
moollaza: No need for this anymore :)

share/spice/book/book.js
```diff
@@ -115,8 +115,9 @@
             },
 
             templates: {
-                group: 'info',
+                group: 'base',
                 options: {
+                    moreAt: true,
                     description: true,
```
moollaza: I don't think this does anything -- please remove.

share/spice/book/book.js
```diff
@@ -115,8 +115,9 @@
...
                     content: Spice.book.book
```
moollaza: This template should really be renamed to "content.handlebars" -- can fix or we can make a task, its nbd.

share/spice/book/book.js
```diff
@@ -104,7 +104,7 @@
 
                 return {
                     title: header,
-                    image: item.to_read_or_not,
+                    image_url: item.to_read_or_not,
                     description: item.critic_review.snippet.replace(/\.\s\.\s\./g, "..."),
```
moollaza: seems like a regex isn't necessary here? `". . ."` should suffice?

MrChrisW: :+1:

share/spice/book/book.js
MrChrisW: :+1:

share/spice/book/book.js
MrChrisW: :+1:

MrChrisW:
@moollaza Agree with the use of the `text` template, easy :wink: A few questions! 

Looking at the screenshots above which content will be contained in the `content` template and which content will be contained in the `subtitle_content` template? It seems we could put everything in the subtitle_content template? 

Thanks :smiley:

moollaza:
@MrChrisW we're going to find a better way of handling this, but right now it's broken so I think we should move forward with my suggestion for now.

I would put the subtitle, image, and rating text in the `subtitle_content` template. Then it'll all be in one div and remain in-line.

The review snippet and "recommended by" should remain in the `content` template.

MrChrisW:
@moollaza This should be ready for a review. 

menu_367
menu_369
selection_368

moollaza:
>I cannot seem to get the title text to wrap/push to the left of the rating image as the title div is created before the image as stipulated by the template

I think that's fine -- I wasn't expecting it to :)

MrChrisW:
> I think that's fine -- I wasn't expecting it to :)
@moollaza So is the margin-top on the subtitle tolerated?

MrChrisW:
@moollaza Updated, cleaned the code up a lot :wink: 

Found a live bug :bug:, using a spelling mistake to bypass relevency checks: https://duckduckgo.com/?q=book+review+mookingjay - Root cause is not checking the `total_results` and calling failed.

moollaza:
@MrChrisW sorry for the delay, I'll review this next week!

MrChrisW:
@moollaza :+1:

moollaza:
LGTM :+1:

moollaza:
@MrChrisW sorry for the delay!

Issue #1746, opened 2015-04-10
ProductHunt: Likely needs a relevancy block to check title/description


moollaza:
I accidentally made a typo, "product hunt rem,t" and got some really odd results. Looks like it may be doing a fuzzy match.

We should be using a relevancy block to ensure the title or description are relevant.

https://duckduckgo.com/?q=product+hunt+rem%2Ct&ia=producthunt

PR #1750, opened 2015-04-12
ProductHunt: Add relevancy block - #1746

MrChrisW:
Fixing #1746

http://duckpan.mrchris.me/?q=product+hunt+rem%2Ct&ia=producthunt


iambibhas:
I was also going through it and saw this. WIthout the relevancy block - 



with the block, with `name` as a key - 



It throws away those other answers even though they have health in their name?

MrChrisW:
@iambibhas Yes I noticed the same thing, the issue is some of the `name` values contain HTML elements (`<em>Health</em>`) that need to be stripped before checking.

I'll probably end up doing something [like this](https://github.com/duckduckgo/zeroclickinfo-spice/blob/d8fda00a343420ec131af75468990d9d8cfae40b/share/spice/news/news.js#L50-L54) and using `strip_html` to remove the HTML values.

iambibhas:
@MrChrisW not always the issue, like for Health Mate - `"name":"Health Mate"`. No HTML or anything. But the relevancy block just threw it away.

MrChrisW:
@iambibhas Sure, this is why I hate the function `DDG.isRelevant`, I'd rather something like [fuzzy.js](http://blog.bripkens.de/fuzzy.js) as we can customise a threshold :smile: 

Anyway here's the results, it looks like if there's 4> chars that do not form part of the candidate string it's not relevant. 
```
Health vs Health IQ - true
Health vs HealthKit - true
Health vs Health Mate - false
Health vs Picnic Health - false
Health vs Microsoft Health - false
Health vs Stride Health - false
Health vs Mango Health - false
Health vs Noora Health - false
Health vs Asthma Health - false
Health vs Oscar Health (iOS) - true
Health vs Oscar Health - false
Health vs Bellabeat - false
Health vs Human 3 - false
Health vs TrueVault - false
Health vs The Fabulous - false
Health vs Joyable - false
Health vs Emissary - false
Health vs CliniCloud - false
Health vs CloudMedx - false
Health vs Up by Jawbone - false
```

iambibhas:
This is an issue. Those are perfectly good results.

MrChrisW:
Updating styling of the tile to include `item.author.name` in a subtitle. Thanks to @iambibhas for the recommendation :+1: 

selection_338

Slight spacing issue on mobile. **Self note, add:**
```css
.is-mobile .tile--producthunt .tile__title {
    height: 1em;
}
```

selection_340

share/spice/product_hunt/product_hunt.js
```diff
@@ -47,7 +58,9 @@
                     footer: Spice.product_hunt.footer
                 },
                 variants: {
-                    tileTitle: '2line-small',
+                    tile: 'basic1',
```
moollaza: We can remove this if we're specifying the `tileTitle` and `tileSnippet`

share/spice/product_hunt/product_hunt.css
```diff
@@ -20,7 +20,11 @@
     margin-bottom: 10px;
 }
 
-.is-mobile .tile--producthunt .tile__content {
+.tile--producthunt .tile__title {
+    height: 1em;
+}
+
+.is-mobile .tile__content {
```
moollaza: I think we can remove all this now -- the variants and internal CSS should be taking care of the heights/spacing :)

moollaza:
@MrChrisW @iambibhas I believe I figured out the problem with the relevancy here, you guys did find a bug! The problem is that we aren't ignoring the skipped words when considering the length of the query. Then number of relevant words and the length of the query are both taken into consideration. If you change the query to `producthunt health` the rest of the results show up. I'll make a task for the team to fix this.

share/spice/product_hunt/product_hunt.js
moollaza: Sorry about the confusion on this stuff, seriously the docs are coming, we have someone writing them right now. The idea behind `tile: basic(1/2/3/4)` is that it's similar to using `group: text` -- it sets some defaults for the `tileTitle`, `tileSnippet` and `tileFooter` -- if we're explicitly setting them we don't need the `tile: 'Basic1'`

MrChrisW:
Thanks for the feedback @moollaza, glad we could assist in finding a bug :beetle:! :+1: @iambibhas 

I've updated, removed the CSS completely and used built-in classes to set colour style.  The spacing is still an issue on mobile however...

selection_347

MrChrisW:
@moollaza I've added a CSS fix for the mobile issue. Caused by the height of `.tile__title` set to `height: 2.5em;`

moollaza:
Hmm is that a bug in the CSS? I'd like to understand the problem better because the tiles probably shouldn't do that so if there's a problem, we'll need to fix it in our CSS/Templates

MrChrisW:
@moollaza Looks like it, you'd have to find out why the `height` is set like that. :smile:

moollaza:
@MrChrisW it looks like we need to set the height for the title so that 2-line titles have enough space for their second line, but also so all tiles have the same height. If we force it to `1em` we cut off 2-line titles and if we allow dynamic heights, we have tiles with different heights it seems.

moollaza:
@MrChrisW sorry for the back-and-forth. Let's just keep the CSS to this:

```css
.tile.tile--producthunt {
    width: 16em;
}

.tile--producthunt  .tile__title  {
    font-weight: 400;
}
```

Should look like: 
2015-04-22_17h26_22

------

@abeyang please take a look at this -- I'm trying to remove as much of (hopefully all!) of the custom CSS for ProductHunt so we can see how the templates look initially. If we want any changes made, I'd like to see if we can apply them using our variants. Right now I've left the CSS which sets the tile `width` to `16em` and the `font-weight` to `14em`.

It looks like tile snippets don't have any bottom-margins by default so perhaps we should add that to the templates?

MrChrisW:
>If we force it to 1em we cut off 2-line titles and if we allow dynamic heights, we have tiles with different heights it seems.

Sure understood. Setting the `width` as you suggested doesn't seem to produce the result displayed in your screenshot, this is what I get.

selection_380

```css
.zci--producthunt .tile.tile--producthunt {
    width: 16em;
}

.zci--producthunt .tile--producthunt  .tile__title  {
    font-weight: 400;
}
```

moollaza:
@MrChrisW sorry I also switched the `snippet` variant to `small`!

MrChrisW:
@moollaza This still doesn't fix the title height issue on mobile. 

menu_388

moollaza:
@MrChrisW it's not really an issue -- it's expected -- at least for now. We're re-considering the implementation and how the title/subtitle should be have.

MrChrisW:
@moollaza No problem. I'll update with your recommendations and we can get this merged :+1:

MrChrisW:
@moollaza Sorry for the delay on pushing these changes! :+1: Should be right to go!

moollaza:
@MrChrisW some stuff has changed recently -- to make sure the subtitle always stays below the title, please use the `altSubtitle` property *instead of* the `subtitle` and everything should be resolved :+1:

MrChrisW:
@moollaza Done :+1: Thanks for pointing that out, I should probably take another look at the docs!

product hunt bike at duckduckgo

altern8tif:
Ping @moollaza.  This LGTG?

moollaza:
continued in https://github.com/duckduckgo/zeroclickinfo-spice/pull/2052

PR #1751, opened 2015-04-12
Gifs Spice: small code cleanup, ready for launch

moollaza:
This one should be ready now for production. I'm wondering if we should rename the package `DDG::Spice::Giphy` or potentially, `DDG::Spice::Gifphy::Search` -- I'd like to make IA's for their other endpoints so the names would be similar, i.e. `Gihpy::Trending`, `Giphy::Random`, `Giphy::Translate`

I also noticed that our CSS isn't applying the `margin: auto` to the images in the detail pane so they're floating off to the left. Not sure if we want to fix that internally or add CSS to this IA?

2015-04-12_19h48_50

---

I noticed that on mobile, the detail view uses the fixed image instead of the Gif -- is that expected?

**Update:** Looking at the template, we use the thumbnail for the detail view on mobile (because we set `loadHighRes = false` -- I think we need better handling of that for gifs? No point it showing a static image in the detail, even on mobile...

/cc @jagtalon @bsstoner @russellholt 

---

> IA Page: https://duck.co/ia/view/gifs


share/spice/gifs/gifs.js
```diff
@@ -1,31 +1,43 @@
...
-        },
-        meta: {
-            sourceName: 'Giphy',
-            sourceUrl: 'http://giphy.com/search/' + encodeURIComponent(searchTerm),
-            sourceIcon: true,
-            count: res.pagination.count,
-            total: res.pagination.total_count,
-            itemType: 'Gifs'
-        },
+        var searchTerm = DDG.get_query().replace(/gifs?/i,'').trim();
```
MrChrisW: This won't match the trigger `giphy`

moollaza: :+1:

share/spice/gifs/gifs.js
```diff
@@ -1,31 +1,43 @@
...
-        },
-        meta: {
-            sourceName: 'Giphy',
-            sourceUrl: 'http://giphy.com/search/' + encodeURIComponent(searchTerm),
-            sourceIcon: true,
-            count: res.pagination.count,
-            total: res.pagination.total_count,
-            itemType: 'Gifs'
-        },
+        var searchTerm = DDG.get_query().replace(/gifs?|giphy/ig,'').trim();
```
jagtalon: Maybe shouldn't be `//g`? There's the case where you could search for "giphy gifs" and it would strip out both of them.

jagtalon: `trim()` looks like it's only on IE 9. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

jagtalon:
I actually like the gifs to move in the thumbnails but they eat up too much data. Got 17mb for "hello gif". We should solve the mobile problem on our end though.

share/spice/gifs/gifs.js
MrChrisW: >trim() looks like it's only on IE 9. 

You deploy a polyfill on production for the `trim` function :+1:  
https://duckduckgo.com/d1758.js

jagtalon: @MrChrisW I should know that--thanks!

moollaza:
I think this one can move ahead then?

jagtalon:
@moollaza I added `image_mobile` which shows the gif (instead of the thumbnail) on mobile. :+1:

PR #1753, opened 2015-04-13
RxInfo: Updates

medicosconsultants:
Remove duplicate "More at" text from sourceName. DuckDuckGo is already including "More at", no need to have it twice.

Remove logical operators "and", "or" and "not" from query since api doesn't support them.


share/spice/rx_info/rx_info.js
```diff
@@ -28,16 +28,14 @@
...
             triggerWordMatch = parseInt(source.match(/rx_info\/[^\/]+\/(\d)/)[1]),
-            relCheck;
-        
-
-        // meta information
-        var sourceName = "More at DailyMed",
-            sourceUrl  = "http://dailymed.nlm.nih.gov/",
+            relCheck,
+            sourceName = "DailyMed",
+            sourceUrl = "http://dailymed.nlm.nih.gov/",
```
moollaza: This link should ideally direct the user to more search results -- can we add a search parameter to this URL?

medicosconsultants:
An API update last week has resolved this issue. I have therefore removed the code patch that was put in place to limit the number of returns. The solution is ready to go.

lib/DDG/Spice/RxInfo.pm
```diff
@@ -6,24 +6,27 @@ use DDG::Spice;
...
-my $triggerWords = join "|", share('triggerWords.txt')->slurp(chomp => 1);
+spice wrap_jsonp_callback => 1;
 
 handle remainder => sub {
-    return unless $_;
-    return $_, (/$triggerWords/) ? "0" : "1"; # return 0 if match in triggerWords file found
+    # Remove logical operators from query since they're disregarded by the API.
+    $_ =~ s/and//ig;
+    $_ =~ s/or//ig;
+    $_ =~ s/not//ig;
```
MrChrisW: These can be done on a single line:

`s/and|or|not//ig` there's also no need to include `$_ =~`

MrChrisW:
@medicosconsultants  Can you describe the API changes? and if possible provide some examples. You'll also need to remove the troggerWords.txt file.

medicosconsultants:
@MrChrisW triggerWords.txt was removed in the previous version. I condensed the removal of the logical operators, thank you for the suggestion.

This instant answer works by leveraging the API's full-text parsing which has recently been tuned to greatly improve relevancy.

For example: If you search now for "pill pfizer blue" the instant answer should return only blue pills with an imprint of pfizer. Previously, if a user misspelled pfizer "pill phizer blue", the instant answer would return all blue pills ignoring the unmatchable term "phizer", the latest version returns no results.

To be clear, the improved relevancy is just not limited to misspellings, rather any term that does not match.

share/spice/rx_info/rx_info.js
```diff
@@ -19,40 +19,45 @@
...
             return Spice.failed('rx_info');
         }
         
-        var script = $('[src*="/js/spice/rx_info/"]')[0],
-            source = $(script).attr("src"),
-            triggerWordMatch = parseInt(source.match(/rx_info\/[^\/]+\/(\d)/)[1]),
-            relCheck;
-        
+        var sourceName = "DailyMed",
+            sourceUrl = "http://dailymed.nlm.nih.gov/",
```
moollaza: Can we make this dynamic so it links to something relevant to the query?

medicosconsultants: @moollaza Assuming you are referring to the `sourceName` and `sourceUrl`, unfortunately at this time there is no single page that could be linked that would reflect the users search with any degree of specificity.  It may be less confusing to the end user to just remove this text / link for now?

moollaza: @medicosconsultants doesn't need to be a single page, but rather a dynamic one, i.e. a search page where we can append the query to the url. For example, something like...

```js
var query = api_result.query,
      url = "https://dailymed.nlm.nih.gov/search?q=" + query;
```

moollaza: If nothing like that is currently possible, then we can opt to send them to another domain and add an atribution link for the data (so users are aware we're using the RxImage API) -- that will need to be done later once the functionality is added to the templates though

medicosconsultants: @moollaza It's certainly possible we will be able to provide you with such a link to a dynamic page in the near future. Could we temporarily remove `sourceName` and `sourceUrl` so the results will appear similar to the Images and Video tabs which also don't provide a “More at” link?  Alternatively could the “More at” by changed to read “Powered by RxImage API” ?

moollaza: @medicosconsultants sorry for the long delay!

This is now possible using our new `moreText` property.

You can have a "More at" link that points somewhere useful for the user, and use the `moreText` to mention the data source.

The "Is It Raining?" Spice is a great example: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/rain/rain.js#L33

medicosconsultants: Zaahir,

Thank you so much for the response. moreText is a great new feature! The
next release of our code will take advantage of it. I will keep you posted.

moollaza: >Thank you so much for the response. 

No problem :)

>moreText is a great new feature! The next release of our code will take advantage of it. 

Excellent! Thanks

>I will keep you posted.

Please do!

moollaza:
@medicosconsultants Ping! I'm just checking in to see if you're still interested in finishing this one up? We'd love to see this go live!

moollaza:
@medicosconsultants I'm closing this one for now, but if you'd like to finish this up, we'll gladly open it back up for you!

PR #1754, opened 2015-04-13
Template Group Adjustment - Movies/Media

sdougbrown:
A bunch of 'movies' IAs are using some custom styling and scripting that could be standardized.  These have been consolidated into a 'movies' template group.

We're also adjusting the 'media' template group to be more flexible, so we're moving other IAs that were using it to the 'products_simple' group for now.

cc @bsstoner @moollaza @jagtalon @abeyang

### DIFF
--- a/share/spice/bbc/bbc.js
+++ b/share/spice/bbc/bbc.js
@@ -93,9 +93,7 @@
                 };
             },
             templates: {
-                group: 'media',
-                detail: 'products_detail',
-                item_detail: 'products_item_detail',
+                group: 'products_simple',
                 options: {
                     buy: Spice.bbc.buy,
                     subtitle_content: Spice.bbc.subtitle_content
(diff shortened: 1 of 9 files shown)


jagtalon:
@sdougbrown thanks! Checking it out.

jagtalon:
Adding an Internal tag because the "movies" template group hasn't actually been released.

jagtalon:
@sdougbrown tested it and it looks good. Only thing I found was:

- InTheaters has the time cut-off on mobile (I see it on my phone as well). This happens on the live site as well so I guess it's not really this PR's fault :)

screen shot 2015-04-16 at 4 08 39 pm

jagtalon:
Will merge when the internal changes are live.

sdougbrown:
This is live. :)

PR #1755, opened 2015-04-13
Remove Locale::Country alias/renaming

zachthompson:
To clear up warnings from Locale::Country that countries are being renamed/aliased multiple times we need to centralize those calls.

This PR will need to be coordinated with https://github.com/duckduckgo/duckduckgo/pull/125


moollaza:
LGTM :+1:

PR #1757, opened 2015-04-13
IndeedJobs: Add relevancy check

moollaza:
/cc @jagtalon @tagawa 

Prevents queries like "steve jobs" from passing -- though technically they do return somewhat relevant results (i.e. "steve" exists in the description) -- we could generalize this to check the description as well if we wanted.

Couldn't use a relevancy block because I want to only compare the search term portion of the query, and ignore the location portion if it exists.

FTR, I can't add a Perl test, because technically this query is valid and will trigger an API call.

### DIFF
--- a/share/spice/indeed_jobs/indeed_jobs.js
+++ b/share/spice/indeed_jobs/indeed_jobs.js
@@ -1,6 +1,6 @@
 (function(env) {
     "use strict";
-    
+
     env.ddg_spice_indeed_jobs = function(api_result) {
         if (api_result.error || !api_result.results.length) {
             return Spice.failed('indeed_jobs');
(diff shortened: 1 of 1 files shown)


jagtalon:
@moollaza looks good! Although shouldn't we be limiting the trigger for GithubJobs? 

screen shot 2015-04-16 at 1 30 50 pm

jagtalon:
:+1:

moollaza:
>looks good! Although shouldn't we be limiting the trigger for GithubJobs? 

Yup, I'll do that now.

tagawa:
Thanks @moollaza &mdash; good idea adding this.

PR #1761, opened 2015-04-14
IPLookup - Hotfix

MrChrisW:
Live error: https://duckduckgo.com/?q=107.191.116.44

```js
TypeError: api_result.ddg.blacklists is undefined
```

We need to check blacklists exists before checking length. 

I broke this in c833f6ad901eac52a36af1c660c88eb285775634 :smiley:

### DIFF
--- a/share/spice/iplookup/iplookup.js
+++ b/share/spice/iplookup/iplookup.js
@@ -30,7 +30,7 @@
             templates: {
                 group: 'text',
                 options: {
-                    content: api_result.ddg.blacklists.length ? Spice.iplookup.content : null,
+                    content: (DDG.getProperty(api_result.ddg,'blacklists').length) ? Spice.iplookup.content : null,
                     moreAt: true
                 }
             },
--- a/share/spice/iplookup/iplookup.js
+++ b/share/spice/iplookup/iplookup.js
@@ -19,7 +19,7 @@
             title_info = api_result.ddg.famlist[0],
             subtitle_info = "Owner: " + api_result.ddg.shortwho + location;
 
-        var blacklists = DDG.getProperty(api_result.ddg,'blacklists'),
+        var blacklists = api_result.ddg.blacklists,
             contentData = null;
 
         if(typeof blacklists !== "undefined" && blacklists.length > 0) {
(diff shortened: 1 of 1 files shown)


share/spice/iplookup/iplookup.js
```diff
@@ -30,7 +37,7 @@
             templates: {
                 group: 'text',
                 options: {
-                    content: api_result.ddg.blacklists.length ? Spice.iplookup.content : null,
```
moollaza: Slightly simpler approach would be to use `DDG.getProperty()`:

```js
content: DDG.getProperty(api_result, "ddg.blacklists") ? Spice.iplookup.content : null
```

MrChrisW: @moollaza See https://github.com/duckduckgo/zeroclickinfo-spice/commit/b01cb4cf767688ac6eabc9cc6424f9f28d80cbad 

I was using that, however I found that getProperty would return an empty array `[]` if the object didn't exist. This causes problems with checking the value using the conditional statement. I'll need to test again, however that's why I choose not to use it.

MrChrisW: P.S Please put return types for these DDG Javascript functions on the documentation! :smiley:

MrChrisW: @moollaza Ok so here's the reason I'm not using `getProperty` http://dev.mrchris.me/testing-getproperty.html 

It returns True for an empty array.

moollaza: @MrChrisW it should return an empty array if the property exists and it's an empty array.

The benefit of `getProperty()` is that you can pass along a more complex/deeply nested chain and if any of them along the way don't exist it will return `null` instead of throwing an error. If the final element doesn't exist, it will return `undefined` (as would a native check). If the property does exist, it will return the value of that property:

```js
var a = {
    b: [],
    c: ["hi", "there"]
};

DDG.getProperty(a, "b");
// returns []

DDG.getProperty(a, "b.test");
// returns undefined

DDG.getProperty(a, "c.nested.property");
// returns null (because c.nested is undef)
```

I'll make a note to update the examples and yes we should definitely document the return types for these functions!

MrChrisW: Thanks for the extra examples @moollaza ! I think you just wrote half a page of documentation :laughing:

Returning the empty array in this case is not what we want as the conditional expression will evaluate as true and the blacklists template will load. 

The API will sometimes return an empty array for the blacklists value, therefore we need to check the length before using it.

We could use `getProperty` then if true check the length, or just check if its not `undefined` and check length. Either way works.

jagtalon: @MrChrisW thanks for the info. This looks good! :+1:

PR #1766, opened 2015-04-15
Bitcoin - All In One IA

MrChrisW:
Amalgamate the Bitcoin IAs into a single IA that swaps endpoints/templates depending on the query.

Remove the following IAs

+ Bitcoin Address
+ Bitcoin Balance
+ Bitcoin Block
+ Bitcoin Transaction

This does not replace the Bitcoin exchange rate IA named Bitcoin.pm, ~~I'll create a new PR to have this renamed.~~ -  #1767

Live Example: 
+ Address: http://duckpan.mrchris.me/?q=19KJZ5aqWnrrCaNAS5Pbd4uknPFBxjALRt
+ Block: http://duckpan.mrchris.me/?q=00000000000000000b5985ad74162bcab96d75c5d7974e54930df32d55fe2408
+ Transaction: http://duckpan.mrchris.me/?q=60d709d4e222b4e950fcbf6fa8923557cc28c7795cc9e468b6c1ceae083e2c4c

-----
IA Page: https://duck.co/ia/view/bitcoin

PR #1767, opened 2015-04-15
Rename Bitcoin IA => BitcoinRate

MrChrisW:
Rename Bitcoin.pm to BitcoinRate.pm along with test files, js, css, handlebars. Tested working.
This should be merged before #1766


moollaza:
I'm wondering if we should eventually deprecate this IA in favour of the XE (and coming Cryptocurrency) Spice?

MrChrisW:
>I'm wondering if we should eventually deprecate this IA in favor of the XE (and coming Cryptocurrency) Spice?

I think that's a great idea, I'd vote in favor of moving it towards using the XE Currency Spice and just displaying the [exchange rate with USD](https://duckduckgo.com/?q=1+btc+to+usd&t=ffab&ia=currency) or use $loc to determine the persons local currency and convert based on that, although I'm not to confident in the reliability of the location API so maybe not :smile:

PR #1771, opened 2015-04-16
Allow chains without apostrophes to trigger local

b2ddg:
Allow for queries with and without the apostrophe to trigger the local IA. An example is, McDonalds, vs,  McDonald's.

@moollaza 
cc @nilnilnil


jagtalon:
@b2ddg works well! Could we have a test for this in `t/`?

moollaza:
LGTM :+1: 

Will add a test after (I can't push to @b2ddg branch otherwise I'd do it here)

b2ddg:
thanks! @moollaza @jagtalon

PR #1772, opened 2015-04-16
Update PromptIfStale to use index_base_url

moollaza:
Required update, you can no longer specify `index`

/cc @zachthompson


zachthompson:
Thanks! (need to update goodies too)

PR #1774, opened 2015-04-16
GitHubJobs: Deprecate in favour of new Indeed Jobs Spice

moollaza:
We may revert this later, or bring back GitHub Jobs as an alternative source when that becomes possible.

/cc @jagtalon


jagtalon:
https://duck.co/ia/view/github_jobs has been deprecated as well. :+1:

jagtalon:
We're replacing it with IndeedJobs. 😁

PR #1776, opened 2015-04-16
Revert "Allow chains without apostrophes to trigger local"

moollaza:
Not actually working -- I was testing the wrong Spice.

Reverts duckduckgo/zeroclickinfo-spice#1771


jagtalon:
@moollaza why was it reverted? It looked like "mcdonald's nearby" was working.

jagtalon:
Oh I was testing the wrong spice too. :/

jagtalon:
So yeah, we need a `/t` file as well. :) regex is hard to read.

moollaza:
>@moollaza why was it reverted? It looked like "mcdonald's nearby" was working.

@jagtalon "mcdonalds nearby" triggers the Maps::Places IA (I made the same mistake). "Mcdonalds" was intended to trigger the Maps::Chains IA like the query "Mcdonald's" does.

Brian and I were debugging it -- and noticed a couple bugs:

1. We're slurping in UTF8 chars and not using UTF8 mode so they're broken in the regex
2. More importantly, our regex isn't working at all with the updates when it's slurped. Copying the regex directly into the `triggers querly_lc => qr/^...$/;` causes it to work perfectly though so something else is going on that Brian is investigating

PR #1778, opened 2015-04-16
Allow apostrophes and dashes to be optional in chain words

b2ddg:
Updating the chains regex so local triggering works with and without apostrophes and dashes in chain words. 

@moollaza @jagtalon


jagtalon:
@b2ddg old queries doesn't trigger for me:

screen shot 2015-04-16 at 3 57 06 pm

Could you add a test in `t/` as well?

PR #1779, opened 2015-04-16
sourceUrl shows visualization

twinword:
Changed sourceUrl. Now more at link shows visualization graph with user typed search word.

### DIFF
--- a/share/spice/word_map/word_map.js
+++ b/share/spice/word_map/word_map.js
@@ -11,7 +11,7 @@
             name: 'Answer',
             data: api_result,
             meta: {
-                sourceUrl: 'https://www.twinword.com',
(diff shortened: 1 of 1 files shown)


share/spice/word_map/word_map.js
```diff
@@ -11,7 +11,7 @@
             name: 'Answer',
             data: api_result,
             meta: {
-                sourceUrl: 'https://www.twinword.com',
+                sourceUrl: 'https://www.twinword.com/?mailLink=' + encodeURIComponent(api_result.encrypt_entry) + '&from=ddg',
```
moollaza: @twinword please remove the `&from=ddg` -- we can't allow you to track our users :)

You can use the referer header to know that users are coming from DuckDuckGo.com

moollaza:
@twinword thanks a lot! LGTM :+1:

PR #1782, opened 2015-04-17
sun_rise_set - using tx--14 class

### LINKED ISSUE #1780: SunRiseSet: Subtitle should use `tx--14` font-size class
Instead of setting the font-size in the CSS use `ems` we should apply the `tx--14` class to the element in the template:

```html
<div class='sun_rise_set--header text--secondary tx--14'>
```

https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/sun_rise_set/content.handlebars#L1

iambibhas:
Fixes #1780.


moollaza:
@iambibhas thanks!

PR #1784, opened 2015-04-17
Transit::Switzerland - using moment to show 12h time

### LINKED ISSUE #1781: Transit::Switzerland: Should show AM / PM instead of military time.
nrphydx

iambibhas:
Fixes #1781. 

I'm showing local time of the train, not the local time of the user. Is that fine? Might be easy to, let's say, instruct someone over phone or plan something.

selection_013
selection_012

### DIFF
--- a/share/spice/transit/switzerland/transit_switzerland.js
+++ b/share/spice/transit/switzerland/transit_switzerland.js
@@ -31,8 +31,8 @@
                 normalize: function(item){
                     console.log(item);
                     return {
-                        arrival_time: moment(item.to.arrival).zone(item.to.arrival).format('LT'),
-                        departure_time: moment(item.from.departure).zone(item.from.departure).format('LT'),
+                        arrival_time: moment(item.to.arrival).utcOffset(item.to.arrival).format('LT'),
+                        departure_time: moment(item.from.departure).utcOffset(item.from.departure).format('LT'),
                         name: item.sections[0].journey.name,
                         platform: item.from.platform,
                         status: (item.from.delay) ? 'Delayed' : 'On time',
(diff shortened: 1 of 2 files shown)


share/spice/transit/switzerland/transit_switzerland.js
```diff
@@ -5,54 +5,50 @@
...
+                    options: {
+                        content: Spice.transit_switzerland.train_item,
+                        moreAt: true
+                    }
+                },
+                normalize: function(item){
+                    console.log(item);
+                    return {
+                        arrival_time: moment(item.to.arrival).zone(item.to.arrival).format('LT'),
+                        departure_time: moment(item.from.departure).zone(item.from.departure).format('LT'),
```
MrChrisW: Looks like `.zone` has been [depreciated](http://momentjs.com/docs/#/manipulating/timezone-offset/) in version 2.9.0 of Moment.js - Consider [moment.fn.utcOffset](http://momentjs.com/docs/#/manipulating/utc-offset/) instead.

iambibhas: Done.

MrChrisW: :+1:

mattr555:
I'm not sure if we want to switch to 12 hour time, as [24h time seems to be the standard in Switzerland](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Europe#Austria.2C_Germany.2C_Switzerland). Would it be better to match the signage and ticketing info in the station?

tagawa:
We could take it one step further and have 12h time for North America and 24h for everywhere else.

jagtalon:
@mattr555 that's a good point, thank you. 

@tagawa I could imagine this would only be useful in Switzerland though since it's the current trains. So no one in the US would really look at this to plan their trip, I think so probably not worth doing that.

@iambibhas local time of the train is best, I think since it's only relevant in the current location. Hmm, so I guess we're not moving to 12 hr format anymore. That's my bad. Please remove that part from the PR.

mattr555:
I guess this can be closed now that #1835 is fixed too?

jagtalon:
@mattr555 I guess so. @iambibhas closing this.

PR #1791, opened 2015-04-21
similar_sites - improved triggers

### LINKED ISSUE #1788: SimilarSites: Triggering needs to be improved
I've tried to trigger this IA a few times and had no luck. The following queries should trigger this IA:

- "websites similar to X"
- "similar websites to X"
- "X similar websites"
- "similar to X"

In all cases we need to make sure that X is a domain (meaning it has a TLD) and if not, bail.

iambibhas:
Fixes #1788. Added tests for queries mentioned in that issue.


lib/DDG/Spice/SimilarSites.pm
```diff
@@ -15,17 +16,20 @@ topics "everyday", "special_interest";
...
-handle remainder => sub {
-    s/^https?:\/\///;
-
+handle remainder_lc => sub {
+    $_ =~ s/(to|like|\s)*//g;
+    $_ =~ s/^https?:\/\///;
     # remove whitespace
-    s/^\s+|\s+$//g;
+    trim($_);
+
```
moollaza: Looks like we need to make sure the candidate website is actually a website, i.e. has a TLD. Right now "similar sites to facebook" would trigger and I don't think the API assumes ".com" in those cases

moollaza: We could potentially add it for them, but may lead to some confusing results

MrChrisW: >Looks like we need to make sure the candidate website is actually a website, i.e. has a TLD. Right now "similar sites to facebook" would trigger and I don't think the API assumes ".com" in those cases

If there's a requirement to validate FQDN's please I'd suggest using `Domain::PublicSuffix` see https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Whois.pm#L31

iambibhas: I tried using `URI->new()` but that creates a `_generic` object for queries like `sites like facebook`, and `_generic` doesn't have any `host` property. I'll try this `PublicSuffic`.

iambibhas: Actually, right now, even on lie site, queries like `sites like facebook` doesn't work. Leads to this http://www.similarsitesearch.com/api/similar/facebook

moollaza: Ah, well it's best if we can avoid making redundant API calls because the assumption is that if we triggered an IA, we're expecting it to return a result.

lib/DDG/Spice/SimilarSites.pm
```diff
@@ -15,17 +16,20 @@ topics "everyday", "special_interest";
...
 
 spice to => 'http://www.similarsitesearch.com/api/similar/$1';
 
 spice wrap_jsonp_callback => 1;
 
-handle remainder => sub {
-    s/^https?:\/\///;
-
+handle remainder_lc => sub {
+    $_ =~ s/(to|like|\s)*//g;
```
jagtalon: This doesn't look reliable. Consider the query "similar sites to like.com". Please add this to the test as well.

moollaza: I think adding a boundary check (`\b`) should suffice? Or perhaps anchoring the regex

lib/DDG/Spice/SimilarSites.pm
```diff
@@ -15,18 +16,22 @@ topics "everyday", "special_interest";
 attribution github => ['https://github.com/Adman', 'Adman'],
             twitter => ['http://twitter.com/adman_X', 'adman_X'];
 
-triggers startend => "similar sites", "similar sites to", "similar web to", "sites like", "websites like", "webs like", "pages like";
+triggers start => ["sites like", "site like", "websites like", "website like", "webs like", "pages like"];
+
+triggers any => ["similar sites", "similar web", "websites similar", "website similar", "similar websites", "similar website"];
```
moollaza: I figure we might as well support "similar to" as well. Seeing as we're checking to make sure we're working with a domain, it should be OK

moollaza:
@iambibhas almost there! It looks like we're still triggering for websites that don't have a TLD. Let's just assume `.com` and append it to the domian name?

iambibhas:
@moollaza so now even this works - 

selection_040

altern8tif:
@iambibhas Geez... Is Lycos still around?? :open_mouth:

moollaza:
@iambibhas is this one ready for review?

iambibhas:
@moollaza yes. I've made all the changes and added tests etc.

moollaza:
LGTM :+1: 

Thanks a lot @iambibhas!

PR #1795, opened 2015-04-21
Time: Switch to use Moment.js, improve relevancy checks

moollaza:
Investigating if there's a better way to handle relevancy checking...


jagtalon:
@moollaza what's the status of this?

PR #1797, opened 2015-04-21
Update Stocks Spice, switch data source

moollaza:
/cc @bsstoner @abeyang @chrismorast @pswam @jagtalon 

2015-04-21_16h56_42

### DIFF
--- a/lib/DDG/Spice/Stocks.pm
+++ b/lib/DDG/Spice/Stocks.pm
@@ -4,11 +4,10 @@ package DDG::Spice::Stocks;
 use strict;
 use DDG::Spice;
 
-primary_example_queries "AAPL";
+primary_example_queries "stock quote AAPL";
 description "Shows stock quote for a ticker";
 name "Stocks";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Stocks.pm";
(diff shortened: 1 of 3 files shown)


abeyang:
@moollaza Can we have "Data by Xignite" to be the same color as "More at NASDAQ"?

moollaza:
@abeyang fixed, https://moollaza.duckduckgo.com/?q=stock+quote+nke&ia=stock

moollaza:
Just noticed I need to round the percent change to 2 DP and also add the percent sign.

moollaza:
Updated on https://moollaza.duckduckgo.com/?q=stock+quote+MSFT&ia=stock

jagtalon:
@moollaza there's a deprecation warning in the console.

screen shot 2015-04-22 at 1 51 13 pm

moollaza:
@jagtalon ya I noticed that -- I think it's because of the way I'm parsing the `time` they give us. Unfortunately they give us a time string that can't be parsed by native JS or moment.js "hh:mm:ss A" doesn't follow any standard conventions *which* really sucks. I read the thread they linked and I don't believe we're at risk using moment's parsing ability

jagtalon:
:+1:

jagtalon:
We moved because YCharts is deprecating their API.

PR #1798, opened 2015-04-21
Forecast: return city, region and country

moollaza:
/cc @tommytommytommy


tommytommytommy:
@moollaza Looks good to me! I added a few tests for different locations and deployed it on my instance to try out. I think it's ready to merge unless there's something that I'm not aware of?

moollaza:
Builds are failing for other reasons (related to `[AutoPrereqs]` I think, /cc @zachthompson)

This should be good for merge :+1:

tommytommytommy:
:+1:

PR #1800, opened 2015-04-22
Quandl: http to https

brianrisk:
Hi @moollaza, Quandl has deprecated the http access for API calls and will be shutting it down completely on May 15.  I've made the necessary changes to the code here (which amounted to adding an 's' to three files.)


moollaza:
LGTM :+1:

PR #1803, opened 2015-04-23
Fixed broken links.

sighmon:
Using {{url}} for the <a href> links instead of building it up with the broken {{id}}.


MrChrisW:
Hi @sighmon, can you explain what you mean by "broken `{{id}}`" ? 

It looks like `{{url}` is still [created the same way](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/newint/newint.js#L28) as in those handlebars files, this is cleaner solution.

sighmon:
@MrChrisW I'm not actually sure how it was working in the original version, because I hadn't declared <code>id</code> at all. It's created in the same way, but just cleaner and *working*. :-)

MrChrisW:
@sighmon Got it, thanks for clarifying :smile: LGTM :+1:

PR #1804, opened 2015-04-23
Time - fixed triggering on random query

### LINKED ISSUE #1801: Time: any query that ends with "time" triggers...
Any query that ends with "time" seems to trigger a search for the user's local time.

Examples
- https://duckduckgo.com/?q=pdt+time&ia=time
- https://duckduckgo.com/?q=testing+time&ia=time
- https://duckduckgo.com/?q=foobar+time&ia=time
- https://duckduckgo.com/?q=adhjasdhjkd+hjkasdjasd+ashjkas+time&ia=time

---

Looks like we need to `return` if the query didn't match our regex, but still has a remainder.

iambibhas:
Fixes #1801. 

I think it was caused by not putting the `^`. :|


moollaza:
LGTM :+1:

PR #1807, opened 2015-04-24
Dogo: Switch Movies & Books to movies template group

moollaza:
Need to switch these over to the new 'movies' template group, now that it's released.

Also removed `ref=` param from `sourceUrl` links -- that's not allowed ;)

/cc @jagtalon


moollaza:
Updated UI:
2015-04-24_18h46_43

We'll bring back the tile ratings and title shortly ;)

Issue #1793, opened 2015-04-21
Quixey: Skip GoogleTV apps


moollaza:
Unfortunately the API doesn't differentiate GoogleTV apps from standard Android apps, so it seems the only way is to check the name for "GoogleTV". If it matches, we should toss the result as it's not very relevant to most users.

Example: https://duckduckgo.com/?q=youtube+android+app&ia=apps&iai=0 -- the YouTube result has two Android links, one of which is for GoogleTV -- let's get rid of the GoogleTV result.

moollaza:
It may also help to check the `package_name` which contains, "googletv": `package_name":"com.google.android.youtube.googletv`

jagtalon:
Fixed in #1810

PR #1810, opened 2015-04-26
Quixey: Additional Relevancy Checking - #1793

MrChrisW:
Fixing #1793 - Remove Google TV editions of Android apps based on regexp for ` /Google\s?TV/i` on `package_name` or `name`


jagtalon:
:+1:

PR #1812, opened 2015-04-28
IndeedJobs.pm: Switch to deep trigger.

jagtalon:
We're switching to deep triggers because it's difficult to extract the location from a query. We can understand location for some queries such as "javascript jobs in philadelphia" because of the "in" preposition, but we can do nothing for queries such as "javascript nyc jobs" because we don't have the functionality to get the location part of the query. For the mean time, we switch to a Deep trigger so that we can pass the right location to the Indeed API endpoint.

**What is a Deep trigger?**

A Deep trigger is triggering based on the SERP results instead of on the query. This means we can look in the search results, see indeed.com domains (such as indeed.com/q-Javascript-l-San-Francisco,-CA-jobs.html), extract the things that we want using regexp (in this case, "Javascript" and "San Francisco, CA"), and pass that along to the endpoint. Unfortunately, it isn't open source at the moment so it's a bit of a black box at the moment.

screen shot 2015-04-27 at 9 29 30 pm

Currently depends on an internal PR. Will ping if it's ready.

### DIFF
--- a/lib/DDG/Spice/IndeedJobs.pm
+++ b/lib/DDG/Spice/IndeedJobs.pm
@@ -18,27 +18,15 @@ attribution github  => ['parker', 'Parker'],
             twitter => ['ourmaninjapan', 'Daniel Davis'],
             web     => ['http://daniemon.com/', 'Daniel Davis'];
 
-triggers any => "job", "jobs";
+triggers any => "///***never trigger***///";
 
 spice to => 'http://api.indeed.com/ads/apisearch?publisher={{ENV{DDG_SPICE_INDEED_APIKEY}}}&v=2&useragent=DuckDuckGo&userip=1.2.3.4&q=$1&l=$2&co=$3&sort=date&format=json&callback={{callback}}';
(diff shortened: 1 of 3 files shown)


moollaza:
@jagtalon I think we can/should still maintain triggering for queries that *do* specify a location. Meaning the query must match the pattern "X jobs in Y" -- not sure we need to remove *all* the triggering functionality

jagtalon:
@moollaza I thought about it although unfortunately, there are downsides to that. :(

- The biggest downside is that we can't tell if the location is for the US or not so we can't reliably change the country. Searching "javascript jobs in london" or "javascript jobs in ontario" gives bad results:

screen shot 2015-04-28 at 12 01 03 pm
screen shot 2015-04-28 at 12 01 10 pm

- Adding back some triggers would mean calling the endpoint twice for some queries. It's a few kb but it's still a call.

moollaza:
@jagtalon good points

moollaza:
LGTM :+1:

Issue #1813, opened 2015-04-28
GoWatchIt: Incorrectly shows "Available for streaming" on Netflix


moollaza:
https://duckduckgo.com/?q=Watch+Full+metal+jacket&ia=watch

Seems like it's only available for Netflix mail? Not sure why this shows streaming is available.

/cc @jagtalon

MrChrisW:
Easy fix in JS is: 

```js
if(item.provider_name === "Netflix" && item.category !== "online") {
    return null;
}
```

Do we want to remove these results completly or just change the way they're presented? Example mockup:
![selection_389](https://cloud.githubusercontent.com/assets/5282264/7363410/9ae5e474-edac-11e4-8001-a3eb64e610cc.png)

moollaza:
@abeyang @chrismorast thoughts on this? Jordan noticed we say "Available for streaming" when in fact it's only available for rent via mail -- is @MrChrisW proposed wording^ OK?

abeyang:
@moollaza @MrChrisW can we say "Available for rent", and push "rent" to the second line?

MrChrisW:
@abeyang Happy with something like this?
![selection_391](https://cloud.githubusercontent.com/assets/5282264/7388948/2630ae0c-ee9c-11e4-9020-cb8d65188af5.png)

abeyang:
@MrChrisW yep, please capitalize 'R'. Thanks!

MrChrisW:
@abeyang :+1: Done #1816 Thanks!

abeyang:
:+1: /cc @moollaza

jagtalon:
Fixed by @MrChrisW  in #1816

PR #1816, opened 2015-04-29
GoWatchIt: Fix buy_line for "Netflix Mail" result

MrChrisW:
Fixing - #1813

selection_392

//CC @jagtalon @abeyang 

I'd rather implement this by adding the provider to an array like you have with `purchase_providers` etc, however this seems to be the quickest way to deal with it.


jagtalon:
:+1:

PR #1820, opened 2015-05-01
gTLD+ccTLD Instant Answer

javathunderman:
What does your Instant Answer do?

Returns the country or purpose of a TLD (top-level domain). 

What problem does your Instant Answer solve (Why is it better than organic links)?

It allows you to see, at a glance, what TLDs exist and can be registered, and which can't. It allows you to see a short, straightforward answer with no clicks. 

What is the data source for your Instant Answer? (Provide a link if possible)

Namecheap API: http://api.namecheap.com/xml.response

Why did you choose this data source?

It was the most reliable source I could find. 

Are there any other alternative (better) data sources?

Possibly Wikipedia, but we'd have to update that static list whenever there's a new TLD. 

What are some example queries that trigger this Instant Answer?

".us tld", ".in tld", etc. 

Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)

Most likely programmers or website devs. 

Is this Instant Answer connected to a DuckDuckHack Instant Answer idea?

Yep: https://duck.co/ideas/idea/1606/country-s-domain. This also carries on from another goodie implementation of this IA. 

Which existing Instant Answers will this one supersede/overlap with?

N/A

Are you having any problems? Do you need our help with anything?

I can't exactly get this to work (This is my first time creating a Spice from start to finish). 

What are the terms of use for the API? Will DuckDuckGo need specific authorization (e.g. an API key)? Are there any costs associated with API usage?

Not really, there's already an API key in use with the Namecheap IA. 

cc @mintsoft


lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,85 @@
...
+attribution github => ["javathunderman", "Thomas Denizou"],
+            twitter => "Emposoft";
+
+
+
+
+triggers startend => "tld";
+
+my $domain_part_regex = qr|
+                          (?:http://)?       # HTTP protocol scheme part [optional]
```
mintsoft: might want to add `http`, `ftp`, `scp`, `ssh` and so forth to this

lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,85 @@
...
+                          (?<domain> [^/]* ) # domain part
```
mintsoft: Technically I think this should probably be `(?<domain>[^/\?]?)` as `http://hello.com?yes` is a valid uri but only `hello.com` is the domain

mintsoft:
@javathunderman I'm not physically in a location where I can fire up duckpan atm, what's the actual problem you're experiencing? Does the spice trigger when using duckpan query?

javathunderman:
Nope, doesn’t trigger at all. I’ll try implementing those fixes…

mintsoft:
@javathunderman not managed to check it yet, however how does this differ from the Namecheap IA? Are you planning to change the functionality around once you've got it triggering correctly?

javathunderman:
Yeah, Namecheap has a `gettldlist` function in their API, so once this starts to trigger properly, I’m going to switch some things around. I used the Namecheap IA as a base, since it's all coming from the same API. Do you think I should integrate this IA into the Namecheap one (since the data is all coming from Namecheap), or split it up as it is now?

mintsoft:
I'd say it's probably OK to keep it separate for now whilst you're working on it. Once you're done then we can look at the similarities and merge the two if it seems appropriate.

javathunderman:
Having a hard time with this one- do you know why this isn't triggering?

moollaza:
@javathunderman looks like your tests are not correct -- I would add more test cases to see what's going on.

javathunderman:
@moollaza More tests incoming...

javathunderman:
@moollaza Pushed the commit w/ more tests, but it's still giving me triggering errors. I think it something to do with the API call...

javathunderman:
Upon further testing, that's been verified: 

API key not found. Using DuckDuckGo's endpoint:                                                                                                               
"https://ddh1.duckduckgo.com/js/spice/tlds/.in"

moollaza:
Oh yes, you'll need to get your own API key and add it to the environment. DDH1 doesn't have your code deployed to it, so the `/js/spice/tlds/` endpoint doesn't exist, hence the failed API call.

Out of curiosity did you happen to look for any other API sources? Would like to avoid using the Namecheap API if this could be a Fathead -- I don't suspect the ccTLDs are going to change much?

javathunderman:
I did look for other ones, but this one seemed the best, especially since we were already using the Namecheap API…

MrChrisW:
@moollaza I think these kind of queries are already handled via the Wikipedia fathead. 

+ https://duckduckgo.com/?q=.fr&ia=about
+ https://duckduckgo.com/?q=.my&ia=about
+ https://duckduckgo.com/?q=.us&ia=about

Triggering just needs to be updated.

javathunderman:
@MrChrisW Yeah, I know about that. This was mainly for the gTLDs- things like .technology and .photography.

MrChrisW:
@javathunderman Right. The [`primary_example_queries`](https://github.com/javathunderman/zeroclickinfo-spice/blob/master/lib/DDG/Spice/TLDs.pm#L44-L45) only include `.us tld` or `.com tld` which are handled by the Wikipedia IA.

javathunderman:
@mintsoft Is there anything else to do for this? It seems to be passing tests, aside from the fact I don't have the API key.

mintsoft:
@javathunderman I think without a thorough test (including said API key) it'll be near impossible to develop it completely. Is there a trial key or similar that can be signed up for to develop or test?

javathunderman:
Glancing over the Namecheap website, I don't think they do.

javathunderman:
My apologies, they do have one. Will test this out...

javathunderman:
Namecheap's API can only be used for a $50 fee, and their trial API is quite limited- I can't make the requests externally. 

@mintsoft Any ideas on testing this?

MrChrisW:
@javathunderman They offer free access to the Namecheap [sandbox](https://www.namecheap.com/support/api/intro.aspx) environment API which should be sufficient for testing. 

It provides access to the `namecheap.domains.gettldlist` command.
image



You'll need to add an exception for the IP address used to access the API in the user interface.
image

javathunderman:
Funny, I tried to get in and their site wouldn’t let me. Will try again…

mintsoft:
@MrChrisW Thanks for the info!

javathunderman:
@mintsoft This is what I'm getting right now:

image


It is indeed triggering, but it doesn't seem to be showing anything- probably a JS problem. I'll take a lot at the console logs...

javathunderman:
This is what I'm getting: 

image


So it is triggering, but it doesn't seem to be getting that response properly. Most likely a JS problem. @mintsoft

mintsoft:
@javathunderman is there anything in the browser console, i.e. the developer console? Normally errors show up in there

javathunderman:
@mintsoft Nope, absolutely nothing. Looking at the DuckPAN output, it's still saying that I'm missing an API key: `API key not found. Using DuckDuckGo's endpoint:                                                                                                               
"https://ddh1.duckduckgo.com/js/spice/tlds/.us"`

javathunderman:
Whoops, deleted those files.

javathunderman:
@mintsoft ping. Seems like Travis builds are failing...

javathunderman:
@mintsoft Any idea why this is still failing?

javathunderman:
ping @moollaza

javathunderman:
@moollaza Hello?

javathunderman:
@mintsoft Seems like Namecheap's API doesn't let you put TLD names in the query- perhaps this should be implemented as a CheatSheet or something?

lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,79 @@
...
+);
+
+my $query_url = uri_escape("$namecheap_endpoint?") . join( uri_escape('&'), pairmap {
+        # the param must always be escaped
+        my $param = uri_escape("$a=");
+        # the value must not be escaped if it is part of the environment or argument
+        my $val = $b;
+        "$param$val";
+    } @query_list );
+spice to => "https://salute-sunday.codio.io/x.js?u=" . $query_url;
```
moollaza: This should be pointing to the Namecheap API...

moollaza: This should be pointing to DuckDuckGo.com, `https://duckduckg.com/x.js?u=`

javathunderman: Implementing and retesting, thanks.

javathunderman:
@moollaza There's an issue with DuckPAN that's making these tests fail.

share/spice/tlds/TLDs.js
```diff
@@ -0,0 +1,43 @@
+(function(env) {
```
moollaza: @javathunderman the filename needs to be lowercased here -- all JS and CSS filenames are always lowerced.

moollaza:
> @moollaza There's an issue with DuckPAN that's making these tests fail.

I just pulled cloned your PR and the test is passing locally -- it looks like a different error is being thrown on Travis (you can click through to travis to checkout the error to see what's going on). I suspect it's related to the filename. It might be easier to renamed this `Tlds.pm`

It looks like there's also some syntax errors in the JavaScript, I suggest you use JSHint so you can check your syntax :)

share/spice/tlds/TLDs.js
```diff
@@ -0,0 +1,43 @@
...
+        if(!item) {
+            return Spice.failed('namecheap');
+        }
+        /* extract data from JSON */
+        
+        var tldName = item.tldName;
+        var tldType = item.tldType;
+       
+        var data = {
+            tldType: tldType
```
moollaza: there's a missing comma at the end of this line, causing a parsing error

share/spice/tlds/TLDs.js
```diff
@@ -0,0 +1,43 @@
...
+            tldName: tldName
+        };
+        Spice.add({
+            id: 'namecheap',
+            name: 'tldlist',
+            data: data,
```
moollaza: I'm a little confused -- does the API response contain an array or TLDs? If so I would suggest you simply pass along the array of results you need from the `api_response` to your `data`:

```js
var tld_list = api_result.ApiResponse.CommandRespoinse.gettldlist;
if (!tld_list){
    return Spice.failed('tlds');
}
Spice.add({
    id: 'tlds',
    name: 'TLDs' //this should really be an IA Page topic
    data: tld_list,
    normalize: function(item){
        title: item.tldName.
        subtitle: item.tldType
    },
    ...
}
...
```

share/spice/tlds/TLDs.js
```diff
@@ -0,0 +1,43 @@
...
+            meta: {
+                sourceName: 'Namecheap',
+                sourceUrl: "http://api.namecheap.com/xml.response"
+            },
+            normalize: function(item) {
+                return {
+                    title: item.tldName
```
moollaza: Looks like you're also missing a comma here too!

lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,79 @@
...
+            twitter => "Emposoft";
+
+
+
+
+triggers startend => "tld", "real tld";
+
+my $domain_part_regex = qr|
+                          (?:http://)?       
+                          (?<domain> [^/]* )
```
moollaza: the `*` means `0 or more` matches, which means the whole regex will pass if the string is empty!

lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,79 @@
...
+
+handle remainder => sub {
+    my ($remainder) = @_;
+
+    return unless $remainder;   
+
+    
+    $remainder =~ $domain_part_regex;
+    my $domain = $+{domain};
+
```
moollaza: It's best to explicitly `return` here:

```perl
return $domain;
```

lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,79 @@
...
+triggers startend => "tld", "real tld";
+
+my $domain_part_regex = qr|
+                          (?:http://)?       
+                          (?<domain> [^/]* ) 
+                          (?:[^\s]*)        
+                          |x;
+
+handle remainder => sub {
+    my ($remainder) = @_;
```
moollaza: the remainder is actually a string, so its in `scalar` context. Thus `my $remainder = $_` should be fine.

lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,79 @@
...
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/TLDs.pm";
+attribution github => ["javathunderman", "Thomas Denizou"],
+            twitter => "Emposoft";
+
+
+
+
+triggers startend => "tld", "real tld";
+
+my $domain_part_regex = qr|
```
moollaza: I think this regex might not be doing what you want. I suggest you use debuggex.com so you can see what it's capturing.

It might be easier to just check the domain with a CPAN library, or alternatively like so:

```perl
handle remainder => sub {
    my $remainder = lc $_;
    return unless $remainder =~ m/(?:https?:\/\/)?(?<domain>[a-z]+)\.(\w+)/;  # I *think* this regex is right...
    // do other work and then return our domain
```

moollaza: Yup! It works pretty well: https://www.debuggex.com/r/rPyVKu1pLX1EbfjU

lib/DDG/Spice/TLDs.pm
```diff
@@ -0,0 +1,79 @@
...
+# DDG_SPICE_NAMECHEAP_USERNAME : Namecheap API username
+# DDG_SPICE_NAMECHEAP_APIKEY   : Namecheap API key
+# DDG_SPICE_NAMECHEAP_IP_ADDR  : Allowed end-User IP address
+my @query_list = (
+    ApiUser    => 'javathunderman', # Username required to access the API
+    ApiKey     => 'bac485f35c51466caed5ff499e585528',   # Password required used to access the API
+    UserName   => 'javathunderman', # same as ApiUser
+    Command    => 'namecheap.domains.gettldlist',               # API method
+    ClientIp   => '0.0.0.0',  # End-user IP address
+    DomainList => '$1',                                    # argument to method
```
moollaza: I can't find this documented anywhere in the docs -- are you sure this parameter is expected?

It seems like the API is meant to simply return a list of all current TLDs?

javathunderman:
@moollaza Ah, I was wondering why that was happening. Anyway, thanks. Will push this ASAP!

share/spice/tlds/TLDs.js
javathunderman: odd, that worked for a pet project sometime ago...

share/spice/tlds/TLDs.js
javathunderman: Marking this as done for my own organization...

share/spice/tlds/TLDs.js
javathunderman: Sorry, what did you mean by IA page topic?

share/spice/tlds/TLDs.js
javathunderman: Done.

lib/DDG/Spice/TLDs.pm
javathunderman: Can't let that happen...Done.

lib/DDG/Spice/TLDs.pm
javathunderman: Done.

lib/DDG/Spice/TLDs.pm
javathunderman: Done.

lib/DDG/Spice/TLDs.pm
javathunderman: I wasn't entirely sure about that. The documentation they provide treats it as if it's spitting out all the domain names at the same time, but the code examples they give are a bit different...If that's the case then I guess this could be implemented as a CheatSheet rather than this...

share/spice/tlds/TLDs.js
moollaza: https://duck.co/ia/, checkout the "Topics" dropdown on the right. Those are what should be used for the `name` of IA's. This should help to organize and standardize the Tabs on the SERP.

lib/DDG/Spice/TLDs.pm
moollaza: Oh please link me the code examples, I didn't see those!

moollaza: >If that's the case then I guess this could be implemented as a CheatSheet rather than this...

I don't think a CheatSheet is ideal for this, because it would need to be constantly updated, and there would probably be *way* too many keys. A Spice could work if the API does actually have a search capability, or if the total API response isn't too bit (though a search param is ideal).

Potentially it could be a Fathead, but I think this list is going to change pretty often, so the IA would quickly become stale.

javathunderman: I meant that the CheatSheet would be able to make API calls- is it exclusively for goodies?

moollaza: >is it exclusively for goodies
@javathunderman yup, CheatSheets must be Goodies, because they're simply JSON files added to the CheatSheets Goodie's /share/ directory, meaning the data must be static.

moollaza:
> Anyway, thanks. Will push this ASAP!

Sounds good, let me know when this is working as expected and ready for review!

javathunderman:
@moollaza Still not working as expected...Could you help me out?

javathunderman:
I also renamed those files as you pointed out; they should be tested properly by Travis...

share/spice/tlds/tlds.js
```diff
@@ -0,0 +1,43 @@
...
+        var item = api_result.ApiResponse.CommandResponse.gettldlist;
+        if(!item) {
+            return Spice.failed('namecheap');
+        }
+        /* extract data from JSON */
+        
+        var tldName = item.tldName;
+        var tldType = item.tldType;
+       
+        var data = {
```
moollaza: Looks like this still needs to be simplified: https://github.com/duckduckgo/zeroclickinfo-spice/pull/1820#discussion_r31871580

share/spice/tlds/tlds.js
```diff
@@ -0,0 +1,43 @@
...
+            normalize: function(item) {
+                return {
+                    title: item.tldName,
+                    subtitle: item.tldType
+                };
+            },
+            templates: {
+                group: 'text',
+                options: {
+                    content: Spice.namecheap.namecheap,
```
moollaza: seems like this file doesn't exist?

If it does, it should be named `content.handlebar` so you reference it like so:

```js
content: Spice.namecheap.content,
```

t/Tlds.t
```diff
@@ -0,0 +1,43 @@
...
+spice is_cached => 1;
+
+ddg_spice_test(
+    [qw( DDG::Spice::Tlds)],
+    
+    '.us tld' => test_spice(
+        '/js/spice/tlds/.us',
+        call_type => 'include',
+        caller => 'DDG::Spice::Tlds'
+
```
moollaza: Let's remove all these extra newlines.

share/spice/tlds/tlds.js
```diff
@@ -0,0 +1,43 @@
...
+            tldType: tldType,
+            tldName: tldName
+        };
+        Spice.add({
+            id: 'namecheap',
+            name: 'tldlist',
+            data: data,
+            meta: {
+                sourceName: 'Namecheap',
+                sourceUrl: "http://api.namecheap.com/xml.response"
```
moollaza: Is there some better website we can send them to? This likely won't be very useful!

javathunderman: I am terrible at JS. :smile:

javathunderman:
I just saw this package on MetaCPAN: https://metacpan.org/pod/Net::Domain::TLD

javathunderman:
@moollaza Did you see the package link I left you?

moollaza:
Sorry, just noticed. It looks perfect. Try it out and let me know how it goes!

javathunderman:
@moollaza I think it might be better to reimplement this as a goodie. I'll close this for now and try my hand at that. In case it doesn't go as planned this can always be reopened!

moollaza:
:+1:

PR #1822, opened 2015-05-03
code_search - moved to text template group

### LINKED ISSUE #1461: CodeSearch: Move to `text` template, add Header and Subheader to give more info
https://duckduckgo.com/?q=c%2B%2B+calculator+code&ia=software

Currently the CodeSearch Spice uses the `base` template group, and dumps the code snippet result into a custom template. It doesn't provide any context though, like the name of the file or what exactly you're looking at.

It should be switched to the `text` template group, and more context should be given!

Also, because we're getting multiple results, we should probably show them in a tile view, and show the code in a detail view.

https://duck.co/ia/view/code_search

iambibhas:
Fixed #1461.


share/spice/code_search/code_search.js
```diff
@@ -9,18 +9,23 @@
```
MrChrisW: Can you update the error checking on [L4](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1822/files#diff-57066020d432d64fc9b32ace072e0c3dR4). Right now it doesn't call `Spice.failed` when no results are returned.
Due to the statement **`!api_result.results.length === 0`**

This can be updated to  **`api_result.results.length === 0`** or check the **`api_result.total === 0`** either way :+1:

share/spice/code_search/subtitle.handlebars
```diff
@@ -0,0 +1 @@
+<a href="repo_url" target="_blank">{{repo_url}}</a> - {{file_location}}
```
MrChrisW: Maybe link back to the searchcode page, this includes the script and link to repository. https://searchcode.com/codesearch/view/99010157

share/spice/code_search/code_search.js
iambibhas: Oh. That's silly. How did I miss that.

share/spice/code_search/subtitle.handlebars
iambibhas: Let's have a comment from the designer team. /cc @abeyang

moollaza:
@iambibhas this is looking great! I'm wondering if we have enough aditional data to warrant moving to the info template and using an InfoBox?

iambibhas:
@moollaza you mean like have the code and title on left and other links on right in aux box? Can we get some suggestions from the designer before working on it? I've seen few times before that the designer has different way of seeing it so we end up removing a lot of changes done beforehand. This is all the data we get from the api - 

    {
      "repo": "https://bitbucket.org/incubaid/edk2.git",
      "language": "Python",
      "linescount": 663,
      "location": "/BaseTools/Source/Python/BPDG",
      "name": "edk2",
      "url": "https://searchcode.com/codesearch/view/29338708/",
      "md5hash": "1ee7378f731f8df289f7331e22217cf0",
      "lines": {
        "615": "    def GenerateVpdFile (self, MapFileName, BinFileName):",
        "616": "        #Open an VPD file to process",
        "617": "",
        "618": "        try:",
        "619": "            fVpdFile  = open (BinFileName, \"wb\", 0)               ",
        "620": "        except:",
        "621": "            # Open failed",
        "622": "            EdkLogger.error(\"BPDG\", BuildToolError.FILE_OPEN_FAILURE, \"File open failed for %s\" %self.VpdFileName,None)",
        "623": "        ",
        "624": "        try :",
        "625": "            fMapFile  = open (MapFileName, \"w\", 0)",
        "626": "        except:",
        "627": "            # Open failed",
        "628": "            EdkLogger.error(\"BPDG\", BuildToolError.FILE_OPEN_FAILURE, \"File open failed for %s\" %self.MapFileName,None)",
        "629": "        "
      },
      "id": 29338708,
      "filename": "GenVpd.py"
    },

moollaza:
@abeyang any thoughts on design here? I was thinking we should show an InfoBox to provide context, and then have the code snippet as the body.

iambibhas:
@moollaza this is how it looks with infobox and all info available -

selection_030

moollaza:
@abeyang this one should be ready for a design pass :+1:

abeyang:
Hi @iambibhas , sorry for the delay. And I hate to confirm your fear, but can we lose the infobox? Don't think this example would warrant it. Rather, can we just utilize the title / subtitle, so the general idea is:

> **filename**
> *Language* Example from *Package*

So for your latest example, it would look like:

> **jquery.ajax.queue.js**
> Javascript Example from vyperblog

And the console.log example would look like:

> **console.log.js**
> Javascript Example from trifleJS

Hope that makes sense...

iambibhas:
@abeyang here is how it looks now - 

selection_050

abeyang:
Looks good! :+1:

Issue #1175, opened 2014-10-17
Improve Currency Triggers


bsstoner:
Review and refine the currency names, things like: ```10 dollars to pounds``` or ```juan to dollars``` should work.

Also should ideally work with symbols, i.e. ```$10 to euro```

https://duck.co/ia/view/currency

mrshu:
I just tried the following query today and would assume some Currency Conversion IA to show up

https://duckduckgo.com/?q=AU+%2434.93+to+USD

@bsstoner told me to send this example here so that there are example to test any new improvements against.

MrChrisW:
@mrshu Interesting example. Are you searching for AUD to USD? 

AU isn't a recognised currency code, while the Currency Spice could be improved to trigger in this case, I'm not sure "AU $" is correct terminolog?

mrshu:
@MrChrisW Yes, that's what I was searching for and that's a very good point. I think I saw that on eBay or somewhere like that and just copied it. 

While I understand your point, that other big search engine does show some sort of currency IA which is what I would just expect to see for this query too.

That being said, feel free to ignore this example, I can see how difficult it is to account for all these strange ways of expressing currencies.

MrChrisW:
@mrshu Don't want to ignore the example! It would be great to support they types of queries, it would just mean using the `Locale::Country` package or a list of two letter country and matching the query, definitely possible :smile:

Speaking of other search engines supporting these types of queries, I've checked and some of the results can be strange! https://duckduckgo.com/?q=BR+34.93+to+USD+!g :confused:

nilnilnil:
"@__nil my example above was "£40 in aud". Google does better at AU-centric searching. I'll try to send you them as I find them if you like…"
https://twitter.com/peternlewis/status/583286866549420032

stephenjudge:
Is there any particular reason why the old currency triggers like $20=€ to get a Dollar to Euro conversion, that used to work, now no longer work? I do remember that the information displayed was coming from Wolfram Alpha and I haven't seen any Wolfram Alpha information come up again in a long time. Has DDG lost licensing rights to Wolfram Alpha information?

moollaza:
@stephenjudge we no longer use WolframAlpha, we've since switched to use other APIs for the various niches Wolfram covered. For currency, we've moved to an open-source Spice Instant Answer which uses the XE.com API.

stephenjudge:
[...]

PR #1823, opened 2015-05-03
Currency: Add support for currency symbols

MrChrisW:
Fixing #1175 

**Only adds support for currency symbols** 

Examples queries: 
- $2 to £ => 2 usd to gbp
- 500 usd to ¥ => 500 usd to jpy


javathunderman:
Is this ready to be merged?

MrChrisW:
@javathunderman Awaiting review :+1:

moollaza:
@MrChrisW please @ mention the person you'd like to review it -- I don't believe GitHub notifies you when you're assigned? At least I don't get emails...

MrChrisW:
@moollaza Sure no problem. There's no alerts for assigning. I assumed this would be picked up by you or Jag when going through PR's :smile: 

@jagtalon Are you able to review this? as the original creator of the Currency IA.

jagtalon:
@MrChrisW Nice! A few questions:

- "$500 to kr" doesn't work but "$500 to dkk" does
- NOK and DKK seem to have the same keys. Shouldn't they be unique?
- What is the source of the YML file?
- Should something like "500 $ to php" work as well where "$" maps to "USD"? Prob outside of this PR, though.

MrChrisW:
Hi @jagtalon! 

>"$500 to kr" doesn't work but "$500 to dkk" does

`kr` wasn't included in the currency symbols file as it's not really a unicode "symbol" it can be added. Or maybe to the currencyNames.txt file instead.

>NOK and DKK seem to have the same keys. Shouldn't they be unique? 

Apparently a whole bunch of currencies with `krone` in their name use the symbol `kr`

+ [Norwegian krone](https://en.wikipedia.org/wiki/Norwegian_krone)
+ [Danish krone](https://en.wikipedia.org/wiki/Danish_krone)

>What is the source of the YML file?

http://www.xe.com/symbols.php

>Should something like "500 $ to php" work as well where "$" maps to "USD"? Prob outside of this PR, though.

That's definitely an option, it would require some modification of the regexp guard again but the mapping is already done.

jagtalon:
@MrChrisW cool. I think this is a good start. We can make issues for the others instead. :+1:

MrChrisW:
@jagtalon Damn that was the quickest review I've ever had.

jagtalon:
@MrChrisW

moollaza:
@MrChrisW nicely done!

PR #1824, opened 2015-05-04
CourseBuffet: Adding more triggers

rubydubee:
These were the old triggers:
* (subject) courses online
* online course (subject)
* (subject) (provider_name)

I have updated it to more useful and user friendly(grammatically) triggers:

* online (subject) course/s
* online (subject) class/es
* (subject) online course/s
* (subject) online class/es
* (subject) course/s online
* (subject) class/es online
* (subject) MOOC/s
* (subject) (provider_name)
* (provider_name) (subject) courses

### DIFF
--- a/lib/DDG/Spice/Coursebuffet.pm
+++ b/lib/DDG/Spice/Coursebuffet.pm
@@ -39,7 +39,7 @@ my @providers = (
 );
 my $providers_str = join('|', @providers);
 
-triggers query_lc => qr/(online|$providers_str)?(?:\s+)?(.*)(?:\s+)(courses? online|courses?|classe?s?|moocs?|$providers_str)$/;
+triggers query_lc => qr/(online|$providers_str)?(?:\s+)?(.*)(?:\s+)(courses? online|courses?|class(?:es)?|moocs?|$providers_str)$/;
 
 handle query_lc => sub {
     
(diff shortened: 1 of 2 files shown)


moollaza:
@rubydubee thanks for updating this! We'll be sure to review it shortly and give you some feedback!

moollaza:
My only initial thought is that we should stick to some generic trigger words for triggering, but then use your regex in the `handle` statement to verify the query matches the specifies format

rubydubee:
ah i see, is it better for performance?
how about:
`triggers any => 'course', 'class', 'mooc', @providers;`

moollaza:
>ah i see, is it better for performance?

@rubydubee yes, much better -- with trigger words we'll only need to run that `handle` function and regex against queries that contain those words (which are more likely to be relevant to the IA). With a regex trigger, we need to run the regex against every query (most of which aren't related to the IA)

>how about: triggers any => 'course', 'class', 'mooc', @providers;

That's pretty good, more specific is always better so the triggering signal is strong

moollaza:
@rubydubee Ping! I'm just checking in to see if you're still interested in finishing this one up? We'd love to see this go live!

moollaza:
@rubydubee I'm closing this one for now, but if you'd like to finish this up, we'll gladly open it back up for you!

PR #1827, opened 2015-05-04
Equaldex: LGBT rights by region [New IA] [WIP]

MrChrisW:
New IA for http://equaldex.com

Currently WIP, looking for a **mock-up from the design team** to push this IA to the next level :wink: Any feedback/suggestions from the community are welcome!

Example API response: https://gist.github.com/MrChrisW/d6d1fb29bd7130d75bb8

*I'll update with the PR template when ready*

WIP - Example: 
[selection_403](https://cloud.githubusercontent.com/assets/5282264/7456680/f02b1678-f2b8-11e4-8bcf-f39fa9dda93a.png)

Example result website: - http://equaldex.com/region/united-states
[selection_401](https://cloud.githubusercontent.com/assets/5282264/7456682/f2eaa0c2-f2b8-11e4-8d08-4126535f954c.png)

https://duck.co/ia/view/equaldex


moollaza:
Seems like the list/table view might work here. This might be another good use for a multi-column table.

MrChrisW:
@moollaza Sure, I'm open to any design. I'd like the design to be **decided** on before moving to complete this :+1:

abeyang:
@MrChrisW I agree with @moollaza -- think the list view could work here.

MrChrisW:
Thanks @abeyang I'll mockup an example of using a list view and ping when ready.

MrChrisW:
@abeyang @moollaza Updated to use the list template. 

One little annoyance the `.record__cell--key` `width` pushes longer key names to the next line,  assuming this is expected behavior?

+ **Live preview:** http://duckpan3.mrchris.me/?q=lgbt+rights+United+States&ia=equaldex

**Screenshot**
selection_408

abeyang:
I agree with you @MrChrisW. That is the default behavior, but that's because keys are generally shorter in length. Let's get rid of the width by doing this (properly namespaced, of course):

``` css
.record__cell--key {
  width: auto;
}
```

MrChrisW:
@abeyang 
selection_411

abeyang:
@MrChrisW nice. Can we change the title/subtitle to:

> **LGBT Rights**
> United States

jagtalon:
@abeyang should be ready. Check it out: https://ddh1.duckduckgo.com/?q=lgbt+china&ia=equaldex 

https://duck.co/ia/view/equaldex

jagtalon:
@moollaza @abeyang any tab name recommendations? Equaldex probably doesn't make sense to the ordinary user. Maybe "Answer"?

moollaza:
I think "Geography" or "Answer" would be good

t/Equaldex.t
```diff
@@ -0,0 +1,41 @@
...
+        caller => 'DDG::Spice::Equaldex'
+    ),
+    
+    'gay rights australia' => test_spice(
+        '/js/spice/equaldex/australia',
+        call_type => 'include',
+        caller => 'DDG::Spice::Equaldex'
+    ),
+    
+    'transgender rights usa' => test_spice(
```
MrChrisW: *self-note* This test will fails return a result from the API as `usa` is passed it should be `United States`. Convert country codes to full country name.

moollaza: @MrChrisW I thought we added a `usa` alias? In any case, we just migrated all the Locale::Country stuff to the duckduckgo repo, so any changes should now be made there so it affects all IA's using Locale::Country :)

moollaza: Looks like that alias doesn't exist! https://github.com/duckduckgo/duckduckgo/blob/master/lib/DDG/Meta/CountryCodes.pm

abeyang:
@MrChrisW I think "Answer" would suffice. Looks good otherwise!

t/Equaldex.t
MrChrisW: @moollaza Yup I saw those updates! Props to @zachthompson :smile:

I remember what's causing this, USA is included in the Country package however the [CODESET](http://perldoc.perl.org/Locale/Country.html) needs to be specified as by default it matches only 2 letter codes. I'll push an update.

danleveille:
Hey guys, Dan from Equaldex here. I was just testing out some queries. From my own search traffic data, people usually search "lgbt rights in X" -- and that doesn't seem to be supported here.

A few other trigger words that people commonly use:
"gay laws" / "gay laws in"
"lgbt laws" / "lgbt laws in"

And adding "in" to these current trigger words would be helpful too:
"gay rights in"
"bisexual rights in"
"transgender rights in"

Would it be possible to add those? Otherwise everything is looking great! :)

MrChrisW:
Hey @danleveille ! Thanks for the suggestions.

We can definitely add those additional triggers :+1: I'll push an update shortly. Let me know of there's anything else you notice.

t/Equaldex.t
MrChrisW: @moollaza Ahhh this is annoying! So I'm using `Locate::Country's` `country2code` function (on [line 33](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1827/files#diff-97fe7be41867fb0ad4bbaf2c9095c707R33)) to check if the country name is valid,  I don't actually want the returned country alpha code. The issue is apparently in [version 2 of Locate::Country](http://sipi.usc.edu/manuals/perldoc-html/Locale/Country.html) some variant names were added meaning `country2code` will return a result for `us` or `usa`, which I don't really want as I'm checking for a valid full country name.  Any suggestions?

lib/DDG/Spice/Equaldex.pm
```diff
@@ -0,0 +1,39 @@
...
+category "location_aware";
+topics "special_interest";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Equaldex.pm";
+attribution github => ["MrChrisW", "Chris Wilson"],
+            web => ["http://chrisjwilson.com", "Chris Wilson"];
+
+spice to => 'http://equaldex.com/api/region?format=json&region=$1&callback={{callback}}';
+
+my $triggerRe = qr/^(lgbt|lesbian|gay|bisexual|transgender)\s?(rights?|laws?)\s?(in)?/;
+triggers query_lc => $triggerRe;
```
moollaza: Let's use trigger words here and then check the query against the regex in the handler

t/Equaldex.t
moollaza: @MrChrisW sorry for the super long delay. This has been at the bottom of my inbox for way too long!

I'm not sure I fully understand what's going on, but in the absolute worst case, slurping a text file with a list of countries in it wouldn't be terrible. Certainly not ideal, but the file should be relatively small...

moollaza: We do essentially the same thing for the Quandl World Bank IA and a list of capital cities for the Time IA

lib/DDG/Spice/Equaldex.pm
MrChrisW: @moollaza :+1:

t/Equaldex.t
MrChrisW: @moollaza Happy to switch to a country test file or just deploy a regex hack for the `usa` edge case. I'll try explaining the issue a little better.

````perl
return $country if defined country2code($country);
````

I'm using the `country2code` function to test full country names e.g **"Australia", "Brazil"** `country2code` just converts a full country name to the ALPHA code.

`country2code` should return a value and be "defined" if the country name is correct, however it returns a value for the quires `usa` or `us` (which are not full country names)

`country2code("usa")` returns `"usa"`
`country2code("us")` returns `"us"`

This is due to some variant names being added to the `Locate::Country` package. :cry: 

I think the best course of action, provided these are the only quires effected is to simply convert `usa`, `us` to the full country name using regex :+1:

Comment (moollaza) [inline t/Equaldex.t]: @MrChrisW thanks for explaining. I understand now. It seems like we need a way to ignore/define our own aliases which should be do-able.

/cc @zachthompson 

In the interim I would suggest you use the `all_countries()` array -- I assume that doesn't contain `us` and `usa`? If not, then you can make a hash out of the country-names array and do a lookup to see if the country exists.

---

I stil don't get why `country2code("usa")` doesn't return "US"; If it did, couldn’t you' just get the correct country name by reversing that? I think that would let you support queries like "LGBT rights usa", right?

**Comment by MrChrisW** [comment]
@moollaza Ready for another review :+1: , added a [little fix](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1827/files#diff-97fe7be41867fb0ad4bbaf2c9095c707R30) for the issue mentioned [here](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1827#discussion_r32585529) also support for alpha-3 country codes, e.g `pol` `aus`.

**Comment by MrChrisW** [comment]
@moollaza @abeyang Are we able to move this IA along? Thanks! :smile:

File: share/spice/equaldex/equaldex.js [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+        
+        // Data is stored in region object
+        api_result = api_result.regions.region;
+        
+        Spice.add({
+            id: "equaldex",
+            name: "Answer",
+            data: api_result,
+            meta: {
+                sourceName: "equaldex.com",
```
moollaza: Let's change this to "Equaldex"

moollaza:
@MrChrisW **so** sorry for the delay. This LGTM :+1:

PR #1830, opened 2015-05-05
Start to deprecate models in Spice.add()

bsstoner:
I'd like to move towards deprecating the need to specify a `model` from Spice.add(). This removes it for Audio/Product/Image and replaces it with `relevancy.dup` attribute, which is essentially the only thing those models are actually doing right now anyway.

The only models remaining are Places and Videos. I'd like to replace them with some kind of shared normalize() function or something since that's essentially all those models are doing. They take in data and transform it into standard objects for the templates. I think this logic needs to be open source so developers understand what's happening to the data between when it's passed into Spice.add() and when it shows up in the templates.

This PR also specifies the templates for Images/Videos spices instead of us magically adding them internally. Trying to get everything to follow the same pattern.

@moollaza @jagtalon @sdougbrown

### DIFF
--- a/share/spice/amazon/amazon.js
+++ b/share/spice/amazon/amazon.js
@@ -12,7 +12,6 @@
             name: 'Products',
             data: api_result.results,
             allowMultipleCalls: true,
-            model: 'Product',
             meta: {
                 itemType: 'Products',
                 sourceName: 'Amazon',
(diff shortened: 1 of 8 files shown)


moollaza:
@bsstoner my only thought is that we should jutst make Audio, Places, Images, etc into template groups so that we don't need to specify the templates each time -- the docs can handle explaining what templates are being used by each group. We can go ahead with this is we want and when the groups are ready update the code and docs together.

bsstoner:
Yea, you're right, images/videos should be template groups, I'll add them internally.

jagtalon:
@bsstoner @moollaza tested and it works well. :+1: We can update when the groups are ready.

bsstoner:
Thanks for merging, the groups are actually deployed now so I can go back and update these. Will open new PR.

PR #1831, opened 2015-05-05
SeatGeek - Standard Date Badge

sdougbrown:
Requires an internal PR.

This uses a standard date object to display a date badge via some new internal styling, rather than duplicating the same CSS and styling across all of these different folders.

cc @bsstoner @abeyang @moollaza @jagtalon @MariagraziaAlastra

### DIFF
--- a/share/spice/seat_geek/events_by_artist/footer.handlebars
+++ b/share/spice/seat_geek/events_by_artist/footer.handlebars
@@ -10,16 +10,4 @@
 	{{else}}
 		<div class="noprice tx-clr--lt">Price N/A</div>
 	{{/if}}
-	<div class="date">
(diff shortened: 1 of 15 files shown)


jagtalon:
@sdougbrown nice--lots of cruft removed. Testing it.

bsstoner:
:+1:

bsstoner:
There's an alignment issue on mobile, but seems unrelated, already happening in prod.
screen shot 2015-05-15 at 3 31 31 pm

abeyang:
Hi, for the month, I think we have to use the 3-letter abbreviation.

Also, can someone throw this on a dev server somewhere for me to test / spot-check?

PR #1840, opened 2015-05-09
Movies + InTheaters: Use an alternate image source.

### LINKED ISSUE #1647: Movie + InTheaters: Images are slowly being deprecated in the API
We've been modifying the API results to deal with the constantly changing URL for the images (see https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/in_theaters/in_theaters.js#L6...L16), and it's been a mess especially since the images in the API will be deprecated. 

A good way is to use another API for the images such as http://docs.themoviedb.apiary.io/#introduction/useful-links. It's worth investigating a way of accessing their images through Movie and InTheaters because we're starting to see broken images again.

@moollaza mentioned that we could:
- Use the IMDB ID in the RT results and match it with [MovieDB](https://www.themoviedb.org/)
- ~~Maybe some JS magic to access the MovieDB images without calling the API~~
  - Not possible, image URLs are all unique 

screen shot 2015-03-17 at 11 39 44 am

https://duck.co/ia/view/movie
https://duck.co/ia/view/in_theaters

jagtalon:
Currently using TMDb just to get the images working in the mean time. It makes a call to the MovieImage.pm endpoint for every item in the list. Fixes #1647.

### DIFF
--- a/lib/DDG/Spice/MovieImage.pm
+++ b/lib/DDG/Spice/MovieImage.pm
@@ -1,0 +1,20 @@
+package DDG::Spice::MovieImage;
+# ABSTRACT: Show movies from Rotten Tomatoes.
+
+use strict;
+use DDG::Spice;
+
+triggers any => '///***never_trigger***///';
+
+spice to => 'https://api.themoviedb.org/3/find/$1?api_key={{ENV{DDG_SPICE_MOVIEDB_APIKEY}}}&external_source=imdb_id';
+
+# Uses $loc so needs to not cache back end.
+spice is_cached => 0;
+
+spice proxy_cache_valid => "200 30d";
+
+handle query_lc => sub {
+    return $_ if $_;
+    return;
+};
+1;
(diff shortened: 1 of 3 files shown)


MrChrisW:
2 issues with this after testing: 

  1. Sometimes `movie_results.poster_path` can be `null` this needs to be checked or the image result will be `https://image.tmdb.org/t/p/w185null`

  2. Movies that don't have any data in `movie_results` do not have the `image` attribute applied as per [line 115](https://github.com/duckduckgo/zeroclickinfo-spice/blob/jag/movie_image/share/spice/movie/movie.js#L115), however the item is already created it just shows without the image.

selection_409

jagtalon:
@MrChrisW thanks for the review! Hmm, for the second point, I think it would be best to put a default image in its place. I'll see if I can find one.

MrChrisW:
No problem @jagtalon, good luck with the default image :smile:

jagtalon:
@MrChrisW we don't have a default image yet, but I think this is better than what we currently have. If you don't mind I'm gonna merge it just to fix it and I'll come back to it. :)

PR #1841, opened 2015-05-11
Modify maps workflow to resemble other IAs

tommytommytommy:
This pull request has a matching PR internally, and both need to be merged/released at the same time. The general idea is to make our maps calls act more like other IAs by directly calling `Spice.add()` from the callback, instead of relying on an internal function in the frontend code.

@b2ddg @nilnilnil


jagtalon:
@tommytommytommy thanks testing it out.

jagtalon:
This is good to go! :+1:

PR #1842, opened 2015-05-12
Transit/Switzerland: Added 'unknown' for missing platform values (fixes #1835)

### LINKED ISSUE #1835: Transit Switzerland: API not always providing platform, leads to bad result
https://duckduckgo.com/?q=next+train+to+geneva+from+paris&ia=swisstrains

The footer in the IA shows "Platform " with no number because the API response doesn't contain one. We should check to make sure we have a platform, or show that one has not been assigned yet (I'm assuming that's the case?)

https://duckduckgo.com/js/spice/transit/switzerland/paris/geneva

/cc @tagawa @mattr555

tagawa:


share/spice/transit/switzerland/transit_switzerland.js
```diff
@@ -32,7 +32,7 @@
                     arrival_time: format_time(item.to.arrival),
                     departure_time: format_time(item.from.departure),
                     name: item.sections[0].journey.name,
-                    platform: item.from.platform,
+                    platform: (item.from.platform) ? item.from.platform : 'unknown',
```
moollaza: I think an `||` would be a little simpler?

```js
platform: item.from.platform || 'unknown';
```

tagawa: What was I thinking?! Thanks - updated.

mattr555:
@tagawa Thank you! In order to be [consistent](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/transit/njt/train_item.handlebars#L17) we may want to change it to "N/A" and gray the text to `#aaa`.

BTW @iambibhas also presents a fix to this on #1784. This solution is probably better, though, as hiding the platform may cause some tiles to be taller than others.

tagawa:
Thanks @mattr555, good idea - done.

mattr555:
screenshot - 05142015 - 09 34 58 pm
:+1: Thanks!

iambibhas:
This looks better. :)

PR #1843, opened 2015-05-13
Disable airlines trigger

tommytommytommy:
We are reaching our API limits and need to reign in this trigger. In the interim, let's disable the word-based trigger and rely on our internal trigger until we can find a more general solution.


jagtalon:
@tommytommytommy I'm getting 

```
[NOTICE] These instant answers were not loaded:                                                                                      
[NOTICE] \ {                                                                                                                         
    DDG::Spice::Airlines   "You may only give compiled regexps to regexp trigger types (like qr/reverse (.*)/i) at /home/codio/perl5/
perls/perl-5.18.2/lib/site_perl/5.18.2/DDG/Meta/Block.pm line 39.                                                                    
Compilation failed in require"                                                                                                       
}    
```

Looks like it has to be `triggers any => '...'`;

jagtalon:
`t/Airlines.t` would have to be updated too

jagtalon:
:+1:

PR #1845, opened 2015-05-15
Currency: Fix #1844 - Check api_result.conversion as object

### LINKED ISSUE #1844: Currency: Spice.failed is being called but not exiting
The line should be `return Spice.failed("...");`. 

It's also unnecessarily calling Spice.failed. The check for `api_result.conversion.length` seems to be failing.

screen shot 2015-05-15 at 5 47 21 am

MrChrisW:
`conversions` is an object not an array. 

Don't worry about < IE 9 support, duckduckgo has a polyfill for the `Object.keys` method :smile:  

// @jagtalon


jagtalon:
:+1:

PR #1852, opened 2015-05-18
Rust Cargo Instant Answer

TomBebbington:
Looks up packages on Cargo, a Rust package manager

Allows quicker lookup of Rust / Cargo packages

https://crates.io

It is completely open-source and programmer-oriented, as well as being incredibly user-friendly.

Nope

+ cargo libc
+ cargo gcc
+ jit cargo

Programmers, hobbyists, sysadmins

[Yes](https://duck.co/ideas/idea/5857/cargo-rust-package)

None, though I did base it off NPM due to the similar functionality and API 

No problems so far.

There aren't any I'm aware of. The code for the webite is open-source and the contributors, being mostly Mozillians, tend to be enthusiasts of open-source projects - I'm sure they won't mind, especially if it benefits them and users of the project. 

screenshot from 2015-05-17 17 00 54

## Checklist
```
[] Tested cross-browser compatibility
This shouldn't be an issue, it uses a standard template, there is no custom CSS in this.

    - Linux Mint
        [X] Google Chrome
```


moollaza:
@TomBebbington thanks for contributing to DuckDuckHack! We'll be sure to review this shortly and give you some feedback.

share/spice/cargo/content.handlebars
```diff
@@ -0,0 +1,3 @@
+<pre>[dependencies]
```
moollaza: nit: let's nest the HTML a little better:

```html
<pre>
    [dependencies]
    {{name}} = "{{max_version}}
</pre>
```

moollaza: Actually, is there only one dependencies for all of these?
I'm wondering if we should move this to the InfoBox like we've done for other IA's e.g. [NPM](https://duckduckgo.com/?q=npm+yolo&ia=software&iax=1)

lib/DDG/Spice/Cargo.pm
```diff
@@ -0,0 +1,27 @@
...
+description "Shows a Cargo package";
+name "Cargo";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Cargo.pm";
+icon_url "/i/crates.io.ico";
+topics "sysadmin", "programming";
+category "programming";
+attribution github  => ['https://github.com/tombebbington', 'Tom Bebbington'],
+            twitter => ['https://twitter.com/tomdbebbington', '@Tom Bebbington'];
+
+triggers startend => 'cargo';
```
moollaza: I think we may want to restrict this a little more with "cargo packages" and maybe "rust packages"? as the triggers.
I'm afraid that this will really over-trigger if we assume every query that starts/ends is a valid query for this IA

moollaza: "rust cargo" could be a trigger as well. The Sublime Packages IA uses "sublime packages" as th example.

NPM and RubyGems use their names as the triggers, but those are much less likely to signal an unrelated query.

TomBebbington: Fair enough, I'll try to update to be in parallel with npm again

share/spice/cargo/content.handlebars
TomBebbington: The dependencies are automatically resolved, I don't think there's a need to list each one. The code in the pre block is meant to be in the project build file

zekiel:
https://duck.co/ia/view/rust_cargo

zekiel:
@TomBebbington  can you please update the ID to: rust_cargo  ?

moollaza:
@TomBebbington I'm closing this one for now, but if you'd like to finish this up, we'll gladly open it back up for you!

PR #1855, opened 2015-05-19
Updates for DetailSlider changes

bsstoner:
This goes with an internal PR. Need to update some spices to work with Detail changes.


nilnilnil:
:+1:

PR #1860, opened 2015-05-22
New IA: Playing Cards

kbhat95:

It enables a user to be dealt a hand of n cards. "deal 5 cards" will give them images of five random cards.

It's not really a problem, but it's fun. :)

<http://deckofcardsapi.com>

I recently came across it over at Hacker News, and thought it might be fun to do something with it.

Not that I know of.

* "deal 7 cards"

* "deal 12"

The lazy card-player community, of course! :)

Yes; you can find discussion [here](https://duck.co/ideas/idea/5235/is-there-anything-we-could-do-using-this-api-dec) and the IA page [here](https://duck.co/ia/view/playing_cards).

None.

Yes, there is one, actually. Currently, I'm using the 'movies' group template and assigning one card to one tile. The problem is that this makes the cards clickable, even though there is no useful content one might get by doing so. I need some help getting around this issue.

Nope. It's free and open source.

Nowhere in particular... I was just going through DDG docs and came across it!

playing_cards_ia

## Checklist
```

    Browsers / devices tested on:
    - Ubuntu 14.04
        [X] Google Chrome
        [X] Mozilla Firefox
```

IA Page: https://duck.co/ia/view/playing_cards

### DIFF
--- a/lib/DDG/Spice/PlayingCards.pm
+++ b/lib/DDG/Spice/PlayingCards.pm
@@ -24,6 +24,7 @@ triggers startend => 'deal';
 
 spice to => 'http://deckofcardsapi.com/api/deck/new/draw/?count=$1';
 spice wrap_jsonp_callback => 1;
+spice proxy_cache_valid => "418 1d";
 
 handle remainder => sub {
     my $query = $_;
--- a/lib/DDG/Spice/PlayingCards.pm
+++ b/lib/DDG/Spice/PlayingCards.pm
@@ -1,0 +1,38 @@
+package DDG::Spice::PlayingCards;
+# ABSTRACT: Returns playing card
+
+use strict;
+use DDG::Spice;
+
+name "PlayingCards";
+source "http://deckofcardsapi.com";
+category "random";
+topics "words_and_games";
+description "Deals the user a random set of n cards";
+
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/PlayingCards.pm";
+
+primary_example_queries "deal 7 cards";
+secondary_example_queries "deal 3", "deal 23";
+
+attribution web     => ['https://kbhat95.github.io', 'Kishor Bhat'],
+            github  => ["kbhat95", "Kishor Bhat"];
+
+
+
+triggers startend => 'deal';
+
+spice to => 'http://deckofcardsapi.com/api/deck/new/draw/?count=$1';
+spice wrap_jsonp_callback => 1;
+
+handle remainder => sub {
+    my $num = (split)[0];
+
+    if ($num && $num =~ qr/^\d+$/) {
+        return $num;
+    }
+    
+    return;
+};
+
+1;
--- a/lib/DDG/Spice/PlayingCards.pm
+++ b/lib/DDG/Spice/PlayingCards.pm
@@ -26,10 +26,11 @@ spice to => 'http://deckofcardsapi.com/api/deck/new/draw/?count=$1';
 spice wrap_jsonp_callback => 1;
 
 handle remainder => sub {
-    my $num = (split)[0];
-
-    if ($num && $num =~ qr/^\d+$/) {
-        return $num;
+    my $query = $_;
+    
+    if ($query && $query =~ m/(\d{1,2})(?:random )?(?:(?:playing )? cards)?/) {
+        my $num = (split)[0];
+        return $num if ($num >= 1 && $num <= 52);
     }
     
     return;
--- a/lib/DDG/Spice/PlayingCards.pm
+++ b/lib/DDG/Spice/PlayingCards.pm
@@ -20,7 +20,7 @@ attribution web     => ['https://kbhat95.github.io', 'Kishor Bhat'],
 
 
 
-triggers startend => 'deal';
+triggers startend => 'deal', 'draw';
 
 spice to => 'http://deckofcardsapi.com/api/deck/new/draw/?count=$1';
 spice wrap_jsonp_callback => 1;
--- a/share/spice/playing_cards/content.handlebars
+++ b/share/spice/playing_cards/content.handlebars
@@ -1,0 +1,5 @@
+<div>
+    {{#each cards}}
+    <img src="{{this.image}}">
+    {{/each}}
+</div>
--- a/share/spice/playing_cards/content.handlebars
+++ b/share/spice/playing_cards/content.handlebars
@@ -1,5 +1,0 @@
-<div>
-    {{#each cards}}
-    <img src="{{this.image}}">
-    {{/each}}
-</div>
--- a/share/spice/playing_cards/playing_cards.js
+++ b/share/spice/playing_cards/playing_cards.js
@@ -13,16 +13,15 @@
                 sourceName: "DeckOfCards",
                 sourceUrl: 'http://deckofcardsapi.com'
             },
+            normalize: function(item) {
+                return {
+                    image: item.image
+                };
+            },
             templates: {
                 group: 'movies',
-                variants: {
-                    tile: 'poster'
-                },
-                options: {
-                    content: Spice.playing_cards.content,
-                    moreAt: true,
-                    rating: false
-                }
+                detail: false,
+                item_detail: false
             }
         });
      }
--- a/share/spice/playing_cards/playing_cards.js
+++ b/share/spice/playing_cards/playing_cards.js
@@ -1,0 +1,30 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_playing_cards = function(api_result){
+        if(api_result.error || !api_result.cards) {
+            return Spice.failed('playing_cards');
+        }
+        
+        Spice.add({
+            id: "playing_cards",
+            name: "Deck of Cards",
+            data: api_result.cards,
+            meta: {
+                sourceName: "DeckOfCards",
+                sourceUrl: 'http://deckofcardsapi.com'
+            },
+            templates: {
+                group: 'movies',
+                variants: {
+                    tile: 'poster'
+                },
+                options: {
+                    content: Spice.playing_cards.content,
+                    moreAt: true,
+                    rating: false
+                }
+            }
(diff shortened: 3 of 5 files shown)


moollaza:
@kbhat95 thanks a lot for contributing, we really appreciate it! We'll be sure to review this shortly and give you some feedback!

lib/DDG/Spice/PlayingCards.pm
```diff
@@ -0,0 +1,38 @@
...
+
+
+
+triggers startend => 'deal';
+
+spice to => 'http://deckofcardsapi.com/api/deck/new/draw/?count=$1';
+spice wrap_jsonp_callback => 1;
+
+handle remainder => sub {
+    my $num = (split)[0];
```
moollaza: This is a pretty good approach, but I would feel better if we ran the query against a regex that ensures it matches our expected format which is something like:

```perl
m/deal \d{1,2}(?:random )?(?:(?:playing )? cards)?/;
```

lib/DDG/Spice/PlayingCards.pm
```diff
@@ -0,0 +1,38 @@
...
+
+    if ($num && $num =~ qr/^\d+$/) {
+        return $num;
```
moollaza: Should make sure the number is greater than 0 (and probably < 53)

share/spice/playing_cards/content.handlebars
```diff
@@ -0,0 +1,5 @@
+<div>
```
moollaza: There's a much better way of doing this :)

Our templates create all the markup for you if you disable the `detail` pane and use a `normalize` function to indicate the `image` for each tile.

Take a look at the Movies or InTheaters IA's to see how they do this. You shouldn't need any custom handlebars templates :)

kbhat95: I'm still not getting this to work. What's going wrong here? I feel like I'm missing something obvious.

```javascript
(function (env) {
    "use strict";
    env.ddg_spice_playing_cards = function(api_result){
        if(api_result.error || !api_result.cards) {
            return Spice.failed('playing_cards');
        }
        
        Spice.add({
            id: "playing_cards",
            name: "Deck of Cards",
            data: api_result.cards,
            meta: {
                sourceName: "DeckOfCards",
                sourceUrl: 'http://deckofcardsapi.com'
            },
            normalize: function(item) {
                return {
                    image: item.image
                }
            },
            templates: {
                group: 'movies',
                detail: false,
                variants: {
                    tile: 'poster'
                },
                options: {
                    content: Spice.playing_cards.content,
                    moreAt: false
                }
            }
        });
     }
  }
(this));
```

moollaza: Remove the `content: ...` you don't need to specify any content (template to be used) because we're letting the built-in `movies` template do the work for us :)

kbhat95: Oops, I forgot about that (though I did remove the extra handlebars file). But that wasn't my problem. Here's my code right now; I'm still getting clickable tiles, even after setting ```detail``` to false.

```javascript
(function (env) {
    "use strict";
    env.ddg_spice_playing_cards = function(api_result){
        if(api_result.error || !api_result.cards) {
            return Spice.failed('playing_cards');
        }
        
        Spice.add({
            id: "playing_cards",
            name: "Deck of Cards",
            data: api_result.cards,
            meta: {
                sourceName: "DeckOfCards",
                sourceUrl: 'http://deckofcardsapi.com'
            },
            normalize: function(item) {
                return {
                    image: item.image
                };
            },
            templates: {
                group: 'movies',
                detail: false
            }
        });
     }
  }
(this));
```

moollaza: Ah sorry, you also need to set `item_detail: false` because the `movies` template group also specifies an `item_detail` template, which is what it's trying to use.

kbhat95:
@moollaza Thanks for your help! I hope we can move closer to integrating it now. :)

share/spice/playing_cards/playing_cards.js
```diff
@@ -0,0 +1,29 @@
...
+            return Spice.failed('playing_cards');
+        }
+        
+        Spice.add({
+            id: "playing_cards",
+            name: "Deck of Cards",
+            data: api_result.cards,
+            meta: {
+                sourceName: "DeckOfCards",
+                sourceUrl: 'http://deckofcardsapi.com'
```
moollaza: lets also set `itemType: "cards"` so it says "Showing 5 cards" for example

share/spice/playing_cards/playing_cards.js
```diff
@@ -0,0 +1,29 @@
...
+            },
+            normalize: function(item) {
+                return {
+                    image: item.image
+                };
+            },
+            templates: {
+                group: 'movies',
```
moollaza: Let's switch this to `media`, then we can add a css file in the share dir and alter the tile and image as follows:

- `img` should have a height of 14em (this might change)
- the tile should have an `auto` width
- the img wrap should have an `auto` height

I believe that makes it so the image controls the height/width of the tile and that way we aren't stretching the image to fill the tile

moollaza:
@abeyang this one needs a design review:

I suggested some changes above so that the card image is 14em tall (it looked pretty good at that height), but I figured you may have some thoughts too?

kbhat95:
I'll wait for @abeyang's feedback before making any more changes, then. :)

abeyang:
@kbhat95 it's looking good -- just a few things:

1. Can we make the cards a tad smaller by default?
2. Is it possible to not use the tile at all? If that's too hard, then maybe lose the border and background-color for the tile, so it'll be "transparent"? I think @moollaza will be able to help you with this, if you need it..

Thanks!

moollaza:
I believe it would be simplest to set the border and background colour to transparent (or remove the background color) so that the IA renders each image separately.

Removing the tiles would require a custom template (which we had before) and I'd like to avoid going that route.

kbhat95:
I'll get on this tomorrow, sorry for the delay.

kbhat95:
I'm running into something strange. When I test out the code by running ```duckpan server``` and put in ```deal n cards``` (replacing n with any number), the css is breaking on the results page. I haven't made any changes (that I remember) since the last commit, and that worked perfectly. I've replicated this on both Firefox and Chrome with all extensions disabled. Any ideas what might be happening?

Here's a [screenshot](https://goo.gl/f4QpQ1) of the problem. I've zoomed out a bit to show that the cards are still appearing, it's just the styling of the page that's gone berserk.

share/spice/playing_cards/playing_cards.js
kbhat95: Latest snag: I've made the other changes (smaller ```img``` of 12em height, tile with auto width, transparent border and background), and now need to set the img wrap to have an ```auto``` height. Can I have some help? Whatever I *am* trying isn't working.

Here's the [screenshot](https://goo.gl/8pWCq7), and here's the css:
```css
.zci--playing_cards .tile--playing_cards {
    width: auto;
    background-color: transparent;
    border: transparent;
}

.tile--playing_cards .tile__media__img {
    height: 12em;
}
```

moollaza: I think the problem is related to the padding on the `tile__body` which is below the `` in the DOM. Take a peak and you should see it has a ton of padding. Set it to `0` and it should work.

If that doesn't help, then please push the code so I can run it to see what's going on

kbhat95:
@moollaza @abeyang : 
What do you think? I think I've addressed the issues you've raised.
Here's another [screenshot](https://goo.gl/HCJNXC).

moollaza:
Looking better, Looks like we need to adjust the css for the tiles's when they're selected though to prevent that double border

kbhat95:
Should we remove the mouse-over highlight as well, then?

kbhat95:
@moollaza @abeyang : how does it look?

moollaza:
@kbhat95 I'll be sure to take a look soon!

lib/DDG/Spice/PlayingCards.pm
```diff
@@ -0,0 +1,39 @@
...
+attribution web     => ['https://kbhat95.github.io', 'Kishor Bhat'],
+            github  => ["kbhat95", "Kishor Bhat"];
+
+
+
+triggers startend => 'deal';
+
+spice to => 'http://deckofcardsapi.com/api/deck/new/draw/?count=$1';
+spice wrap_jsonp_callback => 1;
+
```
moollaza: we need to add `spice proxy_cache_valid => "418 1d"` to make sure we don't cache any API responses!

moollaza:
@kbhat95 looks good! @abeyang any final thoughts?

abeyang:
@moollaza It looks the same to me, even though I've updated it on my server and everything. Can you put this on a dev machine for me?

moollaza:
http://moollaza.duckduckgo.com:5000/?q=deal+5+cards&ia=deckofcards

abeyang:
@moollaza D'oh! Is it down already?

moollaza:
@abeyang sorry, http://pizza-emerald.codio.io:5001/?q=deal+5+cards&ia=deckofcards

moollaza:
@kbhat95 I think we should allow "draw" to be a trigger, as in "draw 5 cards"

kbhat95:
@moollaza only "draw", or "draw" in addition?

abeyang:
@kbhat95 @moollaza Design-wise, it looks good! :+1:

moollaza:
@kbhat95 this is just about ready for merge! While testing this though, we noticed that for some reason in Chrome and Safari on OSX the tiles seems to not display initially and have a width of 0. I suspect it's related to the CSS and lack of content in the tiles but I haven't figured it out exactly.

I'm going to debug it a bit and get back to you with my findings!

moollaza:
@kbhat95 I've figured out the problem. Going to merge and submit another PR that modifies your CSS..

kbhat95:
@moollaza thanks! I've noted your changes for reference in future PRs. :)

PR #1862, opened 2015-05-24
BrainyQuote: Added Right Quote (fixes #1628)

### LINKED ISSUE #1628: BrainyQuote: Double quotes missing at end
Ending quote is missing, SS below
quote
@jagtalon @chrismorast are we still aiming for proposed design changes? in #674

killerfish:
fixes #1628 
@moollaza can you set this up, so @chrismorast can test, and recommend changes. Thanks
eg


share/spice/brainy_quote/brainy_quote.css
```diff
@@ -1,34 +1,46 @@
...
+    font-size: 1.5em;
+    font-weight: 300;
+
 }
 
-.brainy__quote {
-    quotes: "“" "”";
-    font-size: 1.5em;
+.zci--brainy_quote .left__quote
+{
```
moollaza: nit: we like to keep CSS brackets inline with the selectors

share/spice/brainy_quote/brainy_quote.css
```diff
@@ -1,34 +1,46 @@
...
+    color: #d0d0d0;
```
moollaza: Let's use our new text-color classes to apply the colors here: 
https://duckduckgo.com/styleguide#txt-n-color
In this case, `tx-clr--platinum-dark`

moollaza:
@killerfish @chrismorast up on here: https://ddh7.duckduckgo.com/?q=gandhi+quotes&ia=quotations

@killerfish looks like you need to remove the `margin-left` though because currently it's causing the left quote to be overlapped by the text and so it's hidden

killerfish:
@moollaza apparently i had a typo in there, which is why the overlapping :/ please update and see if it persists

share/spice/brainy_quote/brainy_quote.css
killerfish: ah didnt know its that color, updated though

share/spice/brainy_quote/brainy_quote.css
killerfish: hasty work :disappointed:  updated.

moollaza:
@killerfish ah good catch! I'll test it again shortly.

moollaza:
@killerfish looking good, just a couple fixes:

- Let's remove the extra body padding (we're adding 0.5em belowe the More at and it looks funny)
- Let's remove the left padding from the attribution. I think we should change the class name to "author" or something else as we have global css that targets `.attribution` and its changing the way the Quote attribution looks.

I think that's all we need and then this is good. I noticed on mobile the leading quote isn't showing though -- is that expected? I think it looks OK.

moollaza:
@killerfish I'm closing this one for now, but if you'd like to finish this up, we'll gladly open it back up for you!

PR #1863, opened 2015-05-25
Call Spice.failed if response fails relevancy

tommytommytommy:
While debugging issues with queries such as "time london" and "time in mauritania", I noticed that we don't always call Spice.failed() after making an API call and failing relevancy. I don't think that this is what is causing the problem with those queries, but just wanted to fix it for good measure...


MrChrisW:
@tommytommytommy LGTM :+1:

PR #1868, opened 2015-05-26
GetEvents - event search in cities

Uriger:
GetEvents Events search engine gives DuckDuckGo visitors the ability to search for events in any place in the world.
For example., when the user enters the search phrase “events in berlin,” the GetEvents spice will display 20 current events from the GetEvents API which take place in Berlin.

GetEvents built the biggest events database in the world. In order to get that information, the user would need to open 20-50 websites per search. We will be able to give them the information instantly.

**What is the data source for your Instant Answer? **
https://getevents.co/ 

GetEvents created the biggest events databases in the world (over 5M live events). They collect events via APIs, their own crawlers, and from their users.
Now GetEvents offer their Events API to travel companies, search engines, calendars and other discovery services.

There are lots of events data sources. All of them are included inside GetEvents. GetEvents is a worldwide solution. 

'events in chicago', 'things to do in chicago', 'chicago events', 'what to do in chicago'

Events lovers, kids, music lovers, sports lovers, travelers.
Travelers who want to discover things to do in their destination.
Essentially everybody (we are all tourists in our own city).

Yes, Songkick and SeatGeek.

GetEvents is responsible for the API key. GetEvents implemented the API for DuckDuckGo. 

Same template as the previous GetEvents answer ('events near me' is the trigger)


moollaza:
@Uriger it looks like your branch may be behind Master -- please merge master into your branch so this is in a mergable state!

- Overall this is looking pretty good, my only concern is that "events" as an end trigger may be too broad and will result in a lot of unnecessary API calls. It may be best to wait for that trigger until we have a better way to determine if the remainder of the query is actually a legitimate location. 

    i.e. "drag racing events" would trigger this, but shouldn't.

- I noticed that a few queries didn't return results from the API but seemed to exist on the webiste ("events in Johannesburg") and also some queries returned results, but didn't display ("events in tokyo"). Can you investigate what's going on in those cases?

- Also, it seems that the API doesn't give results when you try to specify the state/country for the location: "events in durham, england" or "events in duham, north carolina" -- will you be able to support those?

Uriger:
Hi Zaahir,

Regarding the broadness of the 'events' search, is there a way we can know
if it is a location and only search in that case? you give a location api,
does it have other functionalists, maybe? Ones we could use for checking if
the term is a location?

Regrading the cases you mentioned it seems like we overlap with an instant
answer from seatgeek.
I saw in their Sports.pm code that they have a startend trigger for
'events'.
What happens with 'events in tokyo' is that I get this -
Spice.failed() called by Spice with ID "seat_geek_sports".
Our response comes ok but it does not show and the other call fails.
Can it work when the other instant answer falls?

Side note: not only we have seatgeek in our database but also their
'events' trigger is only on sport events while we send all kinds of events.

Regarding the last subject - specify the state/country, we are looking into
it and when we have an answer we will update you.

Thanks,

Uri Gertzek
GetEvents


2015-05-26 17:47 GMT+03:00 Zaahir Moolla <notifications@github.com>:

moollaza:
Hey @Uriger, sorry for the late response!

>is there a way we can know if it is a location and only search in that case?

Unfortunately not yet, but this is something we're actively working towards. Entity detection as a whole is something we're aiming for, but handling locations specifically is a priority, however we don't have anything working yet.

>Regrading the cases you mentioned it seems like we overlap with an instant answer from seatgeek. I saw in their Sports.pm code that they have a startend trigger for 'events'.

Ah I'm sorry I hadn't noticed that -- we may need to reign in the triggers there, but the system is made to allow multiple IA's to trigger -- I'll investigate further, but for the purpose of further testing, you can choose to only load your IA by specifying it's name when you initialize the DuckPAN server: `duckpan server GetEvents::City`

>Can it work when the other instant answer falls?
It should! I'll see what's going on.

>Side note: not only we have seatgeek in our database but also their 'events' trigger is only on sport events while we send all kinds of events.

Yup I know, it's only meant to apply to sports events, but we still use a broad trigger. I'll speak to the team about the SeatGeek integration.

>Regarding the last subject - specify the state/country, we are looking into it and when we have an answer we will update you.

Thanks!

Uriger:
https://github.com/duckduckgo/zeroclickinfo-spice/pull/1900 - corrected for merging

PR #1870, opened 2015-05-27
switch to using images/videos template groups

bsstoner:
Switching to images/videos template groups now that they've been added internally.

@moollaza
cc @sdougbrown


moollaza:
LGTM :+1:

PR #1876, opened 2015-05-29
Convert slurp/Load to LoadFile

zachthompson:
Use LoadFile to ingest YAML from disk.  Addresses https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1114


moollaza:
LGTM :+1:

PR #1877, opened 2015-05-30
Add check for closed market, add to frontend

### LINKED ISSUE #1854: Stocks: Check when market is closed and grey out IA
We can check whether the market is open using another API call (which means we'll need another Perl package for this new endpoint):

http://globalexchanges.xignite.com/xGlobalExchanges.json/IsExchangeOpen?Exchange=XNAS

If the market is closed, we should indicate that on the IA. Specifically, we should update the subtitle like so:

aapl_at_duckduckgo

------

IA Page: http://duck.co/ia/view/stocks

moollaza:
Check if markets are closed, if so, add notice to display. Fixes #1854

New design: 

After 4PM

stock_quote_for_aapl_at_duckduckgo

Following morning, before next opening:
aapl_stock_at_duckduckgo

/cc @abeyang @jagtalon

### DIFF
--- a/share/spice/stocks/content.handlebars
+++ b/share/spice/stocks/content.handlebars
@@ -7,6 +7,10 @@
     <span class="stocks__symbol">{{Security.Symbol}}</span>
     <span>as of {{date}} {{time}}</span>
     <span title="Eastern Time Zone"> ET</span>
+    {{#if marketClosed}}
+        <span class="zcm__sep"></span>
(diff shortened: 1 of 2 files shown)


moollaza:
Live on my machine: https://moollaza.duckduckgo.com/?q=aapl+stock+&ia=stock

abeyang:
Looks good! :+1:

PR #1880, opened 2015-06-01
symbolab spice

levaly:
Provides answers to math problems as seen on symbolab.com. In addition, for general calculator requests such as "integral calculator" returns the relevant page from symbolab.com

It's an instant answer to a specific question, while organic links may or may not have the answer.

http://www.symbolab.com

I work for Symbolab.com

Not that I am aware of.

integral calculator
compute 1+2
solve integral of x^2
calculate x^2+2x-2=0

students

None

No

I was contacted by DDG.

Sent in a separate email.

```
[] Wrote test file and added to t/ directory

    - Windows 8
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 8

    - Mac OSX
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [] Google Chrome
        [] Opera
```

------
IA Page: https://duck.co/ia/view/symbolab

### DIFF
--- a/lib/DDG/Spice/Symbolab.pm
+++ b/lib/DDG/Spice/Symbolab.pm
@@ -1,0 +1,23 @@
+package DDG::Spice::Symbolab;
+# ABSTRACT: Write an abstract here
+
+use DDG::Spice;
+
+spice is_cached => 1; 
+
+name "Symbolab";
+description "Get a solution to a math problem from symbolab.com";
+primary_example_queries "integral calculator", 
+                        "compute integral of x^2",
+                        "solve x+2>3x";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Symbolab.pm";
+
+spice to => 'http://www.symbolab.com/ddg?query=$1';
+spice wrap_jsonp_callback => 1;
+
+triggers any => "calculator", "solver", "calculate", "solve", "compute";
+handle query_raw => sub {
+    return $_;
+};
+
+1;
--- a/lib/DDG/Spice/Symbolab.pm
+++ b/lib/DDG/Spice/Symbolab.pm
@@ -9,11 +9,14 @@ description "Get a solution to a math problem from symbolab.com";
 primary_example_queries "compute integral of x^2", "solve x+2>3x";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Symbolab.pm";
 
-spice to => 'http://www.symbolab.com/ddg?query=$1';
+spice to => 'http://symhost.noip.us:9000/ddg?query=$1';
 spice wrap_jsonp_callback => 1;
 
-triggers start => "calculate", "solve", "compute";
-handle query => sub {
+triggers startend => "calculate", "solve", "compute";
+handle remainder => sub {
+    if ($_ eq "") {
+        return;
+    }
     return $_;
 };
 
--- a/lib/DDG/Spice/Symbolab.pm
+++ b/lib/DDG/Spice/Symbolab.pm
@@ -1,22 +1,19 @@
 package DDG::Spice::Symbolab;
-# ABSTRACT: Write an abstract here
 
 use DDG::Spice;
 
-spice is_cached => 1; 
+spice is_cached => 1;
 
 name "Symbolab";
 description "Get a solution to a math problem from symbolab.com";
-primary_example_queries "integral calculator", 
-                        "compute integral of x^2",
-                        "solve x+2>3x";
+primary_example_queries "compute integral of x^2", "solve x+2>3x";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Symbolab.pm";
 
 spice to => 'http://www.symbolab.com/ddg?query=$1';
 spice wrap_jsonp_callback => 1;
 
-triggers any => "calculator", "solver", "calculate", "solve", "compute";
-handle query_raw => sub {
+triggers start => "calculate", "solve", "compute";
+handle query => sub {
     return $_;
 };
 
--- a/lib/DDG/Spice/Symbolab.pm
+++ b/lib/DDG/Spice/Symbolab.pm
@@ -9,7 +9,7 @@ description "Get a solution to a math problem from symbolab.com";
 primary_example_queries "compute integral of x^2", "solve x+2>3x";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Symbolab.pm";
 
-spice to => 'http://symhost.noip.us:9000/ddg?query=$1';
+spice to => 'https://www.symbolab.com/ddg?query=$1';
 spice wrap_jsonp_callback => 1;
 
 triggers startend => "calculate", "solve", "compute";
--- a/lib/DDG/Spice/Symbolab.pm
+++ b/lib/DDG/Spice/Symbolab.pm
@@ -14,7 +14,7 @@ spice wrap_jsonp_callback => 1;
 
 triggers startend => "calculate", "solve", "compute";
 handle remainder => sub {
-    if ($_ =~ /[\^\*\=\+\-\/\\]/ or $_ =~ /(sin|cos|tan|cot|csc|sec|ln|log|sqrt|integral)/i) {
+    if ($_ =~ /[\^\*\=\+\-\/\\]/ or $_ =~ /\b(sin|cos|tan|cot|csc|sec|ln|log|sqrt|integral)/i) {
         return $_;
     }
     return;
--- a/lib/DDG/Spice/Symbolab.pm
+++ b/lib/DDG/Spice/Symbolab.pm
@@ -14,10 +14,10 @@ spice wrap_jsonp_callback => 1;
 
 triggers startend => "calculate", "solve", "compute";
 handle remainder => sub {
-    if ($_ eq "") {
-        return;
+    if ($_ =~ /[\^\*\=\+\-\/\\]/ or $_ =~ /(sin|cos|tan|cot|csc|sec|ln|log|sqrt|integral)/i) {
+        return $_;
     }
-    return $_;
+    return;
 };
 
 1;
--- a/share/spice/symbolab/subtitle_content.handlebars
+++ b/share/spice/symbolab/subtitle_content.handlebars
@@ -1,0 +1,1 @@
+Solution for: <span class="mathquill-embedded-latex">{{latex}}</span>
--- a/share/spice/symbolab/subtitle_content.handlebars
+++ b/share/spice/symbolab/subtitle_content.handlebars
@@ -1,1 +1,1 @@
-Solution for: <span class="mathquill-embedded-latex">{{latex}}</span>
+Solution for: <span {{#if problemHasLatex}}class="mathquill-embedded-latex"{{/if}}>{{problem}}</span>
--- a/share/spice/symbolab/subtitle_content.handlebars
+++ b/share/spice/symbolab/subtitle_content.handlebars
@@ -1,1 +1,3 @@
-Solution for: <span {{#if problemHasLatex}}class="mathquill-embedded-latex"{{/if}}>{{problem}}</span>
+<h4 class="c-base__sub">
+    Solution for: <span {{#if problemHasLatex}}class="mathquill-embedded-latex"{{/if}}>{{problem}}</span>
+</h4>
--- a/share/spice/symbolab/subtitle_content_nolatex.handlebars
+++ b/share/spice/symbolab/subtitle_content_nolatex.handlebars
@@ -1,1 +1,0 @@
-Solution for: <span>{{latex}}</span>
--- a/share/spice/symbolab/subtitle_content_nolatex.handlebars
+++ b/share/spice/symbolab/subtitle_content_nolatex.handlebars
@@ -1,0 +1,1 @@
+Solution for: <span>{{latex}}</span>
--- a/share/spice/symbolab/symbolab.css
+++ b/share/spice/symbolab/symbolab.css
@@ -2,3 +2,7 @@
     overflow: visible;
     margin: 0.8em 0;
 }
+
+.zci--symbolab sup{
+    top: 0em;
+}
--- a/share/spice/symbolab/symbolab.css
+++ b/share/spice/symbolab/symbolab.css
@@ -1,8 +1,3 @@
-.zci--symbolab .c-base__title {
-    overflow: visible;
-    margin: 0.8em 0;
-}
-
 .zci--symbolab sup{
     top: 0em;
 }
--- a/share/spice/symbolab/symbolab.css
+++ b/share/spice/symbolab/symbolab.css
@@ -1,0 +1,4 @@
+.zci--symbolab .c-base__title {
+    overflow: visible;
+    margin: 0.8em 0;
+}
--- a/share/spice/symbolab/symbolab.js
+++ b/share/spice/symbolab/symbolab.js
@@ -1,0 +1,58 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_symbolab = function(api_result){
+
+        // Validate the response (customize for your Spice)
+        if (!api_result || (!api_result.page && !api_result.solution)) {
+            return Spice.failed('symbolab');
+        }
+
+        if (api_result.page){
+            Spice.add({
+                id: "symbolab",
+                name: "Symbolab",
+                data: api_result,
+                meta: {
+                    sourceName: "Symbolab.com",
+                    sourceUrl: api_result.page.url,
+                    sourceIconUrl: 'https://www.symbolab.com/favicon.ico'
+                },
+                normalize: function(item) {
+                    return {
+                        description: api_result.page.description,
+                        image: api_result.page.img,
+                        title: "Symbolab - " +api_result.page.title
+                    }; 
+                },
+                templates: {
+                    group: 'info'
+                }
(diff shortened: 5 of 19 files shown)


moollaza:
Hey @levaly 
!

1. Pull Request is reviewed by the DuckDuckGo staff and
community

If you have any questions along the way, feel free to ask our staff or community:

### Community Leaders:

- MrChrisW
- mattr555
- mintsoft
- killerfish
- bradcater
- javathunderman
- kablamo
- loganom
- TomBebbington

### Staff:

- @moollaza (me)
- jagtalon
- abeyang
- chrismorast
- russellholt
- zachthompson
- [and a few others](https://github.com/orgs/duckduckgo/teams/owners)

------

Just in case, you can also reach us on [IRC](http://webchat.freenode.net/?channels=duckduckgo) or ask the community through our DuckDuckHack [Developer Email list](https://www.listbox.com/subscribe/?list_id=197814).

------

One more thing, can you please fill out the [pull request template](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/pull_request_template_spice.md) and edit your initial comment. Once completed, we'll be sure to review your PR and give you some feedback.

Thanks!

moollaza:
@levaly sorry for the long delay. The DDG backend has now been updated and will allow you to load Mathquill using our `DDG.require()` function.

I've setup a working demo and the code can be seen here: https://github.com/duckduckgo/zeroclickinfo-spice/compare/zaahir/symbolab-test

This approach will allow you to forego using any custom templates (which simplifies things) and let's mathquill render the LaTeX to HTML once the page has loaded.

I've also reduce the triggerspace any simplified the IA to put the emphasis on the result.

@abeyang do you have any thoughts on what the design should look like for this?

In my text I've used the `Text` template and the result looks like this:

calculate_3x___23___1230_at_duckduckgo

lib/DDG/Spice/Symbolab.pm
```diff
@@ -0,0 +1,23 @@
+package DDG::Spice::Symbolab;
+# ABSTRACT: Write an abstract here
+
+use DDG::Spice;
+
+spice is_cached => 1; 
+
+name "Symbolab";
+description "Get a solution to a math problem from symbolab.com";
+primary_example_queries "integral calculator",
```
moollaza: Let's stick to queries that we can actually require us to calculate a solution. For generic queries like "integral calculator" this Instant Answer isn't any more useful than an organic result, as it only provides a link and so it shouldn't be shown.

lib/DDG/Spice/Symbolab.pm
```diff
@@ -0,0 +1,23 @@
...
+                        "compute integral of x^2",
+                        "solve x+2>3x";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Symbolab.pm";
+
+spice to => 'http://www.symbolab.com/ddg?query=$1';
+spice wrap_jsonp_callback => 1;
+
+triggers any => "calculator", "solver", "calculate", "solve", "compute";
+handle query_raw => sub {
+    return $_;
```
moollaza: Please make sure you use a regex here (or some other logic) to verify the query is valid for this Instant Answer and contains a useful remainder. In it's current state, this will trigger and try to produce a result when the query is anything that contains one of these words. It will even trigger when the query is just "solve" for example, which we don't want,

moollaza: Please make sure you also create an appropriate test file, t/Symbolab.t, which verifies this IA triggers for various expected queries.

We should also add some negative tests that verify broken or unsuitable queries return `undef` meaning they don't trigger is IA.

For example "calculate mortgage rate" shouldn't return a result.

abeyang:
Hi, we should highlight the answer in the title. Thus, it should be something like:

> **x = 1207/3**
> Solution for 3x + 23 = 1230

moollaza:
@levaly sorry for the late response!

>Added a test file, although wasn't quite sure what is the syntax for negative tests. For instance a query for "calculate" shouldn't trigger a request.

Negative tests should simply map to `undef`:
```perl
"calculate" => undef;
"calculate mortgage" => undef;
```

>Mathquill rendering is not quite similar to ours. Not sure why.

How is it different? Do you have a special setup for Mathquil? We're using the latest version of the library and CSS

>Commented out title from presentation. I think it redundant as our latex solution holds the input as 

As Abe mentioned, we typically show the title as the result and have the original question as the sub-title. Do you know if it's possible to separate the left and right hand side of the equation? (or the question/answer?)

>Can we change the text "More at Symbolab" to "Click here to see a step by step solution at Symbolab" ?

We can't change the "more at" text, but we can add an additional link beside the More at link. It would look like so: `More at Symbolab | Step by step solution`. That can be achieved by using the `moreText` property in `meta`.

>Regarding a regex to limit requests to Symbolab API - for instance "mortgage rate". I cannot think of a regex that will satisfy. As I mentioned before, only internally in our code I know if this request is something we can handle or not. I am open to any suggestions.

@levaly I think the regex used by calculator.pm will suffice. At the very least I imagine we should ensure the candidate "equation" is more than just a single word? It will almost certainly need to contain an operator (+, -, *, /) or keyword (sin, tan, integral) etc. Ensuring those exist will help to prevent a lot of redundant API calls, which benefits us both. We should at least try to ensure that our candidate equation actually looks like an equation. For example, if it just contains `[a-zA-Z\s]` it's almost certainly not something we can work with.

moollaza:
@levaly @abeyang it's live on https://comserv.duckduckgo.com/?q=calculate+3+%2B+sin+2&ia=calculator

My only thought is that we should show the "parsed" query, to indicate how we understood the query.

In the case of that result, when you click to see the steps you can see that `sin 12` was understood as `sin(12°)`

levaly:
Sure, where do you suggest we should show it?

moollaza:
@levaly I just realized what you have, doesn't quite implement what @abeyang had specified.

I decided to play around with the code a bit and pushed my update here: https://github.com/duckduckgo/zeroclickinfo-spice/compare/symbolab-spice

I've moved the IA to use a `title_content` template and a `subtitle_content` template which allows both the title and subtitle to be latex that get's rendered on display.

This also achieves the layout we usually aim for which looks like:

> **answer**
> Subtitle: Input

@abeyang does this look okay?
calculate_4x___35___234_at_duckduckgo

This IA is a bit different in that the title (solution) isn't usually going to be plain text. It's going to be rendered as math, which can be complex and contain multiple lines. For example:

calculate_integral__int__left_x 2_ax-3_right_ 2dx_at_duckduckgo

lib/DDG/Spice/Symbolab.pm
```diff
@@ -0,0 +1,23 @@
...
+description "Get a solution to a math problem from symbolab.com";
+primary_example_queries "compute integral of x^2", "solve x+2>3x";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Symbolab.pm";
+
+spice to => 'https://www.symbolab.com/ddg?query=$1';
+spice wrap_jsonp_callback => 1;
+
+triggers startend => "calculate", "solve", "compute";
+handle remainder => sub {
+    if ($_ =~ /[\^\*\=\+\-\/\\]/ or $_ =~ /(sin|cos|tan|cot|csc|sec|ln|log|sqrt|integral)/i) {
```
moollaza: We should probably put `\b` around these captures to prevent matching on a sub-string. For example "instant" would match here

abeyang:
@moollaza that looks much better. But for the second screenshot, can we get rid of the `=` in the beginning? (It's like our general [calculate IA](https://duckduckgo.com/?q=3%2B4&ia=answer).)

levaly:
It's easy enough to get rid of the equals sign, but the fact that it's there might suggest that that is an answer rather than maybe interpreted question. For instance in the following case https://comserv.duckduckgo.com/?q=solve+%5Cint+e%5Ex&ia=calculator it might be confusing that's the question and what's the answer, and the equals sign might help.

moollaza:
@abeyang yup, we should be able to remove the leading `=` sign.

@levaly can you update the IA with my code, and strip the leading `=`? I think a simple regex should suffice?

levaly:
@moollaza all changes are in

moollaza:
@levaly thanks a lot! I'll take a look and give and feedback. Looking forward to getting this live very soon!

share/spice/symbolab/symbolab.js
```diff
@@ -0,0 +1,48 @@
...
+                id: "symbolab",
+                name: "Calculator",
+                data: api_result.solution,
+                meta: {
+                    sourceName: "Symbolab.com",
+                    sourceUrl: api_result.solution.url
+                },
+                normalize: function (data) {
+                    return {
+                        latex: data.originalQuery.replace(/ /g, "\\text{ }"),
```
moollaza: I think I was wrong to use `originalQuery` here -- it seems like using `problemLatex` is a better, more reliable choice?

share/spice/symbolab/symbolab.js
```diff
@@ -0,0 +1,48 @@
...
+                        solution: data.solutionLatex.replace(/^=/, "").replace(/,/, ",\\text{ }"),
```
moollaza: Should we be replacing all of the `,`'s? If so, we need to add `/g` to the regex.

What exactly is this for? Perhaps you should add a comment so others will know as well

share/spice/symbolab/symbolab.js
moollaza: Is there a simple way to detect if the `solutionLatex` actually contains LaTeX content? Will a latex string always contain `\\`?

If so, it may be best  to check for the existence of latex, and if it isn't there, we don't need to add the `mathquill-embed-latex` class to the corresponding div, so we won't waste time converting plaintext.

share/spice/symbolab/symbolab.css
```diff
@@ -0,0 +1,4 @@
+.zci--symbolab .c-base__title {
+    overflow: visible;
+    margin: 0.8em 0;
+}
```
moollaza: it looks like we need to reset the `top` for `sup` elements inside our IA's div as our global CSS modified the `top` which is causing rendering problems for fractions:

```css
.zci--symbolab sup {
    top: 0;
}
```

lib/DDG/Spice/Symbolab.pm
```diff
@@ -0,0 +1,23 @@
...
+description "Get a solution to a math problem from symbolab.com";
+primary_example_queries "compute integral of x^2", "solve x+2>3x";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Symbolab.pm";
+
+spice to => 'https://www.symbolab.com/ddg?query=$1';
+spice wrap_jsonp_callback => 1;
+
+triggers startend => "calculate", "solve", "compute";
+handle remainder => sub {
+    if ($_ =~ /[\^\*\=\+\-\/\\]/ or $_ =~ /\b(sin|cos|tan|cot|csc|sec|ln|log|sqrt|integral)/i) {
```
moollaza: what about "limit" and symbols like ">", "<" etc?

I’d love for this IA to handle as many kinds of math questions as it can handle, that aren't already supported by our builtin calculator.

For example, simple equations like 1+1 or sin(1) can be handled by Calculator.pm -- there's no need for us to spend time requesting results for those from your API when they're trivial to solve.

But for complex queries involving limits, derivatives, integrals, inequalities, etc (all stuff that Calculator.pm ***can't*** handle) it makes perfect sense for use to leverage Symbolab and benefit the user by also showing them a link to Symbolab's step-by-step instructions!

share/spice/symbolab/symbolab.css
levaly: I am not sure I understand the problem. Can you please elaborate?

lib/DDG/Spice/Symbolab.pm
levaly: Anything you can input in symbolab.com input box will be handled as well by DDG. Anything beyond that will require some additional support. Specifically tangent, derivatives  and slopes should be handled right now, limits will only be handled if they are inputted via latex as the syntax is more complex because you have to state where does x tends to.

share/spice/symbolab/symbolab.js
levaly: It's not required to replace commas, but it looks slightly better when there are multiple results. like in case for x^2=25. We should add the /g to the regex.

share/spice/symbolab/symbolab.js
levaly: problemLatex is the interpreted problem in Latex format. So it might be a better choice. If the solution or problem latex do not contain ^ _ \ then I think its safe to assume we don't have to render it into latex.

levaly:
Should I make these changes, or will you?

share/spice/symbolab/symbolab.js
moollaza: It looks like when the input is LaTeX something here is breaking causing a blank output. If you strip the `\:` from the input, it works OK.

For example I copied "calculate long\:division\:\frac{x^4+6x^2+2}{x^2+5}" from Symbolab.com and it produces this output:

calculate_long__division___frac x 4_6x 2_2 x 2_5 _at_duckduckgo

With the `\:` removed (and with the `sup { top: 0; }` css) it looks much better:

calculate_long_division__frac x 4_6x 2_2 x 2_5 _at_duckduckgo

lib/DDG/Spice/Symbolab.pm
moollaza: >Anything you can input in symbolab.com input box will be handled as well by DDG.

Correct, but it seems like the regex you have in place is a little too narrow, so it's reducing what the Instant Answer can do right now? It seems like there's a lot I can do on Symbolab.com that can't be done with the Instant Answer, and I'd like to be able to use as much of your API as we can!

> Anything beyond that will require some additional support. 

Understood, I don't mean for you to change your API at all. I just think the regex we're using here needs to be improved to let more queries through. 

It would be good to have lots of example queries so I can tests this well and especially so we can add them to the Symbolab Instant Answer Page: https://duck.co/ia/view/symbolab so others will be able to see what this Instant Answer can do :)

share/spice/symbolab/subtitle_content_nolatex.handlebars
```diff
@@ -0,0 +1 @@
+Solution for: <span>{{latex}}</span>
```
moollaza: I think it would be better to use an `{{if}}` block to add the Mathquill class when needed.

It could look like this:

```js
normalize: function(data){
  return {
    hasLatex: containsLatex(data.solutionLatex),
    solution: data.solution,
    ...
  }
}
```
```handlebars
Solution for: <span {{if hasLatex}}clas="mathquill-embeded-latex"{{/if}}>{{solution}}</span>
```

This should simplify the JS and also reduce the number of templates we're using.

moollaza:
@levaly I'm see now text in the "Solution for" area here: https://beta.duckduckgo.com/?q=calculate+integral+\int+\left%28x^2%2Bax-3\right%29^2dx&ia=calculator?

share/spice/symbolab/subtitle_content.handlebars
```diff
@@ -0,0 +1 @@
+Solution for: <span {{#if problemHasLatex}}class="mathquill-embedded-latex"{{/if}}>{{problem}}</span>
```
moollaza: It looks like we need to wrap this in an `<h4>` to ensure we apply the correct subtitle styling to this:

```handlebars
<h4 class="c-base__sub">
     Solution for: <span {{#if problemHasLatex}}class="mathquill-embedded-latex"{{/if}}>{{problem}}</span> 
</h4>
```

moollaza: Once this is wrapped in an `h4` we can remove the CSS that adds a `margin` to the title :+1:

share/spice/symbolab/symbolab.css
```diff
@@ -0,0 +1,8 @@
+.zci--symbolab .c-base__title {
+    overflow: visible;
+    margin: 0.8em 0;
```
moollaza: Let's remove this :smile:

levaly:
updated Spice, and API with above changes, here are some screenshots:
screen shot 2015-09-04 at 4 06 00 pm
screen shot 2015-09-04 at 4 06 06 pm

moollaza:
@levaly thanks a lot! Updates are live on beta: https://beta.duckduckgo.com/?q=solve+X^4+%3D+978465&ia=calculator

/cc @abeyang

abeyang:
@levaly almost there! But can we implement what I [mentioned earlier](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1880#issuecomment-132698197), to get rid of the `=` when there's nothing on the left-hand side? So, `calculate integral \int \left(x^2+ax-3\right)^2dx` should return an answer without the `=` (whereas your latest screenshot still shows it).

levaly:
Sorry my bad, it's fixed now.

screen shot 2015-09-10 at 11 01 19 am

moollaza:
@levaly LGTM :+1:

levaly:
Great! When can we expect it to be available in duckduckgo.com?

moollaza:
@levaly congrats! It's live: https://duckduckgo.com/?q=solve+x%2B2%3E3x&ia=calculator

PR #1882, opened 2015-06-02
Zillow has had a major update to their codes.  Supporting YML files i…

brianrisk:
…n share needed updating as well as the route in HomeValues.pm, Quandl.t

@moollaza no real content changes; simply had to rework what went into the path.  A bonus is that now the information will be much fresher!


moollaza:
@brianrisk thanks, I'll take a look shortly

moollaza:
LGTM :+1:

PR #1884, opened 2015-06-02
Single Product Template Updates

sdougbrown:
Some of the classes for the products templates were overly convoluted due to an attempt to re-use classes across too broad a range of styling options.  We're changing the 'single item' view here to make things easier.

Any IAs that use the products templates and could potentially show one item will require an update - I think I got them all, but it's always possible that I missed something. :smile: 

This requires an internal PR.

to @bsstoner 
cc @jagtalon


bsstoner:
:+1:

PR #1890, opened 2015-06-04
Add support for German queries to the "what's the current time in" instant answer

karottenreibe:
This is a change to the Time.pm "what's the current time" instant answer. It adds support for german queries.

* was ist die aktuelle zeit in kingston
* wieviel uhr ist es jetzt in kingston
* wie spät ist es gerade?

## Checklist
[x] added test cases for the German queries
[x] only modified the backend .pm file, so no changes to the design

### DIFF
--- a/lib/DDG/Spice/Time.pm
+++ b/lib/DDG/Spice/Time.pm
@@ -1,5 +1,5 @@
 package DDG::Spice::Time;
-# ABSTRACT: Time zone converter
+# ABSTRACT: Shows the current time for cities, states and countries
 
 use strict;
 use DDG::Spice;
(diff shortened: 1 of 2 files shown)


moollaza:
@karottenreibe thanks a lot for taking the time to contribute, we really appreciate it!

Unfortunately, I'm can't accept this PR though, because we have much larger plans for Internationalization of Instant Answers. It's a non-trivial problem though so it will take some time. Our goal is to make Instant Answers trigger-able in many different languages, and hopefully it won't require much of any changes to the code (like this).

As you can imagine if we wanted to allow trigger for other languages in this manner, the code would quickly become bloated with language-specific triggering which isn't maintainable and leads to unexpected behaviour (e.g. why does this IA work in this language, while this other one doesn't?)

I hope you can understand my reasoning, and I really wish you'll stick around to contribute to DuckDuckHack in other regards (like making a new IA or fixing bugs!)

karottenreibe:
I understand. Bummer that you didn't tell me right away when you guys encouraged me to do this via Twitter. That would have saved me an afternoon :/ It would also help if you wrote this in big bold letters on one of the 20 intro/FAQ pages I read up to get this working. Just so people don't put in a lot of effort for something that will be rejected anyways.
Let me know when you have this i18n stuff up and running, I'll contribute this stuff back then.

moollaza:
@karottenreibe We've now updated the docs to warn others to please not submit PR's that add triggers in other languages.

Apologies for the confusion, the tweet was meant to suggest that if you wanted to add some functionality you *could* as the code is open source. We do ask that users email us or give us a heads up when they plan to make a PR just so we can help pre-approve their contributions to prevents issues like this.

Again, really sorry about the mixup.

PR #1897, opened 2015-06-09
Update to newest Spice API, ensure Spice.failed() call

moollaza:
Wrap everything in a closure (as we do for all Spice IA's) and ensure we call `Spice.failed()` for mobile devices because it shouldn't show.

/cc @nilnilnil @jagtalon


moollaza:
Merging as hotfix.

jagtalon:
:+1:

PR #1900, opened 2015-06-10
GetEvents - event search in cities - corrected

Uriger:
GetEvents Events search engine gives DuckDuckGo visitors the ability to search for events in any place in the world.
For example., when the user enters the search phrase “events in berlin,” the GetEvents spice will display 20 current events from the GetEvents API which take place in Berlin.

GetEvents built the biggest events database in the world. In order to get that information, the user would need to open 20-50 websites per search. We will be able to give them the information instantly.

**What is the data source for your Instant Answer? **
https://getevents.co/

GetEvents created the biggest events databases in the world (over 5M live events). They collect events via APIs, their own crawlers, and from their users.
Now GetEvents offer their Events API to travel companies, search engines, calendars and other discovery services.

There are lots of events data sources. All of them are included inside GetEvents. GetEvents is a worldwide solution.

'events in chicago', 'things to do in chicago', 'chicago events', 'what to do in chicago'

Events lovers, kids, music lovers, sports lovers, travelers.
Travelers who want to discover things to do in their destination.
Essentially everybody (we are all tourists in our own city).

Yes, Songkick and SeatGeek.

GetEvents is responsible for the API key. GetEvents implemented the API for DuckDuckGo.

Same template as the previous GetEvents answer ('events near me' is the trigger)


moollaza:
@Uriger sorry for the delay! I'll take a quick look at the code..

lib/DDG/Spice/GetEvents/Nearby.pm
```diff
@@ -1,15 +1,15 @@
...
 
 primary_example_queries "events near me";
 description "Upcoming events from GetEvents";
-code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GetEvents.pm";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GetEvents/Nearby.pm";
 category "entertainment";
 topics "computing","entertainment","food_and_drink","music","science","special_interest","travel";
 
-triggers startend => 'events near me', 'nearby events';
+triggers startend => 'events near me';
```
moollaza: Why was the other trigger removed?

lib/DDG/Spice/GetEvents/Nearby.pm
```diff
@@ -1,15 +1,15 @@
-package DDG::Spice::GetEvents;
+package DDG::Spice::GetEvents::Nearby;
```
moollaza: @Uriger I'm wondering now if we should simply remove this IA altogether, and pass along the user's city to the other API endpoint when the query is "events near me", "local events", etc?

lib/DDG/Spice/GetEvents/Nearby.pm
Uriger: Sorry - that was a mistake, I will fix this.

lib/DDG/Spice/GetEvents/Nearby.pm
Uriger: @moollaza Since the different queries have both different caching requirement and api structures it made more sense for me to implement the new functionality in a different spice (file).The code is cleaner this way on the pearl side but duplicated on the front end. 
If I change it would the program be able to handle both caching requirements on the same file?

moollaza: @Uriger yup you can have dynamic caching.

If the query is a "nearby" query, where we don't want to cache the nginx re-write, we would:

```perl
spice is_cached => 1;
spice proxy_cache_valid => "200 12h";
...
if (is_local) {
    return $result, {is_cached => 0};
} else {
        return $result;
}
```

Uriger: Hi @moollaza.
Following your advise I'm trying to unify the two IA and I'm having a few difficulties.
The basic issue is that the parameters to the api in the two case are different. In one case I have to call 'https://api.getevents.co/ddg?city=$1' and on the other the call is 'https://api.getevents.co/ddg?lat=$1&lng=$2&timezone=$3'. I tries two approaches:
(a) Change the 'spice to' inside the handle subroutine.
(b) Use spice to => 'https://api.getevents.co/ddg?$1' and return a $1 value which includes both the parameter names and their values.

Neither approach worked.
Further, I'm not sure I fully understand your last response:
spice is_cached => 1; # by default, cache nginx rewrites
spice proxy_cache_valid => "200 12h"; #cache successful API responses for 12h
...
if (is_local) {
    return $result, {is_cached => 0}; # don't cache nginx rewrite, caches API response
} else {
        return $result; # caches nginx rewrite, caches API response
}
-What is $result? What should be it's value?
-What is is_local? Is that just a boolean variable that tells in which case are we?

moollaza: @Uriger sorry, what I meant was to just use the city API endpoint, and when the query is "events near me", instead of taking the user's lat/ln, we take `$loc->city` and pass that to the API which simplifies everything.

>What is $result? What should be it's value?
>What is is_local? Is that just a boolean variable that tells in which case are we?

These were just meant to be placeholders to show how the logic would work. Meaning you'd need to determine if we're dealing with a "local events" query or not. If so, we return whatever result is necessary (i.e. the users city) and set the caching off. Otherwise, we return the city given in the query and leave caching on.

Uriger:
Hey @moollaza I have restructured the code - using only one file and the city name from the location api instead of the geo coordinates. Also caching only when needed.

lib/DDG/Spice/GetEvents.pm
```diff
@@ -1,29 +1,50 @@
...
+use Text::Trim;
+
+
+primary_example_queries "events in chicago","what to do in chicago","things to do in chicago","events near me";
+secondary_example_queries "chicago events","chicago things to do";
 
-primary_example_queries "events near me";
 description "Upcoming events from GetEvents";
-code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GetEvents.pm";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GetEvents1.pm";
```
moollaza: Whoops, this looks like a typo?

moollaza:
@Uriger this is looking pretty good. I've noticed a few things:

1. The `name` in the JS should be updated to just `Events` (not `Local Events`)
2. Similarly, `itemType` needs to be adjusted as well
3. When an event lasts "All day", we're actually not showing the date of the event, the subtitle just reads "All day" (this bug is pre-existing it seems as it's affecting the live GetEvents IA).

    E.g. 
events_in_oshawa_at_duckduckgo

Uriger:
@moollaza I have fixed the three issues that you mentioned.
Thanks.

moollaza:
@Uriger thanks! The code looks good. Will wait for the API updates I emailed you about and then I think this should be finished.

moollaza:
LGTM :+1:

@Uriger thanks!

PR #1907, opened 2015-06-10
Added pound alias to GBP in Currency Spice

domjacko:
This allows users to use the trigger 'pound' to refer to GBP.

Example triggers would be:
"10 pounds in dollars"


moollaza:
LGTM :+1:

PR #1908, opened 2015-06-11
Fix Border CSS

### LINKED ISSUE #1857: SimilarSites: Border should be re-added and set to transparent
We're about to make some internal changes to the spacing of the list elements. In order to preserve the intended spacing we need to change the CSS so that it sets `border-color: transparent` instead of removing the border.

------

IA Page: http://duck.co/ia/view/similar_sites

kbhat95:
Resolves #1857 .

### DIFF
--- a/lib/DDG/Spice/PlayingCards.pm
+++ b/lib/DDG/Spice/PlayingCards.pm
@@ -24,6 +24,7 @@ triggers startend => 'deal';
 
 spice to => 'http://deckofcardsapi.com/api/deck/new/draw/?count=$1';
 spice wrap_jsonp_callback => 1;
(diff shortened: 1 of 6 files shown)


kbhat95:
Oops, sorry about that.

moollaza:
@kbhat95 you'll want to make a new branch that's in-sync with master from the main Spice repo, and branch off of that for your PR's so they don't interfere with eachother.

I would suggest you keep your master branch up to date with ours, and then branch off of master to make PR's. That's how we work with the open repos :)

kbhat95:
@moollaza, yup, that's what I usually do.

I somehow forgot to branch and ended up PR'ing from my master. Won't happen again! :)

Issue #1865, opened 2015-05-25
GoWatchIt: Should support the trigger "movie"


moollaza:
Currently this IA only triggers on "watch", we should also add support for the trigger "movie", e.g., "Skyfall movie"

------
IA Page: http://duck.co/ia/view/go_watch_it

altern8tif:
This still doesn't work for me... It seems like "movie" is still listed as one of the `killwords` – I suppose that's the problem?

![screen shot 2015-06-22 at 9 44 44 am](https://cloud.githubusercontent.com/assets/5793364/8274382/5d7fae3e-18c3-11e5-81dc-17b0b8c35eb9.png)

PR #1909, opened 2015-06-11
Supports trigger movie

gokul1794:
I think this should solve #1865


moollaza:
@gokul1794 not quite. You're no longer using the original array of triggers. You'll want to move the word `movie` from `@killwords` to `@triggers` so that it uses it as a trigger.

Also, whenever updating triggers, please add additional tests to the associated test file in `t/` so that we know for sure it works as expected. 

Thanks!

gokul1794:
@moollaza Anything else?

moollaza:
LGTM :+1:  Thanks a lot @gokul1794 !

PR #1910, opened 2015-06-11
bbc - using moment.js, removed unnecessary helpers

iambibhas:
Saw this IA could use some moment.js love, so using that. Sadly can't remove the `duration()` function as moment.js' `humanize()` function doesn't return accurate duration.

### DIFF
--- a/share/spice/bbc/bbc.js
+++ b/share/spice/bbc/bbc.js
@@ -7,121 +7,107 @@
             return Spice.failed('bbc');
         }
 
-        var query = DDG.get_query(),
-            broadcasts = api_result.schedule.day.broadcasts,
-            programmes = [],
-            now = new Date();
+        DDG.require("moment.js", function() {
 
-        if (query.match(/night|evening/)){
-            now.setHours(18);
-        }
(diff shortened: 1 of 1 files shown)


moollaza:
>Sadly can't remove the duration() function as moment.js' humanize() function doesn't return accurate duration.

Care to elaborate on that? Is it a bug in Moment.js?

iambibhas:
> Is it a bug in Moment.js?

See here - https://github.com/moment/moment/issues/348

It's apparently a feature to not support accurate humanization.

moollaza:
@iambibhas can we do what was described in https://github.com/moment/moment/issues/348#issuecomment-6535794?

I defined a [modification](https://github.com/duckduckgo/zeroclickinfo-spice/blob/ed05efa4152f627dd3e54821bca2eb563d105192/share/spice/tor_node/tor_node.js#L77) to the language in a similar fashion for the Tor Node IA because of a similar problem.

moollaza:
I guess its about as much code either way. This is fine for now :+1:

PR #1916, opened 2015-06-11
SimilarSites: Font weight 400, font size 1.15em

moollaza:
Fixes clipping of content.

/cc @chrismorast @abeyang 

Result:
similar_sites_to_facebook_com_at_duckduckgo

Mobile:
similar_sites_to_facebook_com_at_duckduckgo

### DIFF
--- a/share/spice/similar_sites/similar_sites.css
+++ b/share/spice/similar_sites/similar_sites.css
@@ -12,8 +12,8 @@
 }
 
 .zci--similar_sites .c-list__item a {
-    font-size: 1.357em;
-    font-weight: 300;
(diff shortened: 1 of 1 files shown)


chrismorast:
Can we turn off visited coloring here? Both in the IA and in the footer?

moollaza:
@chrismorast is this something we should do in general for IA's? I'm wondering if we should expose a way for IA's to specify this from the JS (like a variant), otherwise we'll always need to add CSS to handle this

/cc @abeyang

moollaza:
Merging. We're going to handle visited links at a global template level

PR #1664, opened 2015-03-21
Animes IA

jlcarvalho:
# Spice Pull Request Template

Using this template will help us better understand your Instant Answer and assist you when necessary.  Simply copy and paste the markdown below into the description of your GitHub pull request and complete as appropriate.

**What does your Instant Answer do?**
Uses products_simple template to display anime infos from the Hummingbird API.

**What problem does your Instant Answer solve (Why is it better than organic links)?**
The IA shows a list of animes with that query with poster, rating, description and episodes.
I think that's a good amount of information that the user gets before he/she even clicks any of the search results.

**What is the data source for your Instant Answer? (Provide a link if possible)**
https://github.com/hummingbird-me/hummingbird/wiki/API-Reference

**Why did you choose this data source?**
It's a nice, RESTful, widely used API without rate limit

**Are there any other alternative (better) data sources?**
Not found any that returns JSON

**What are some example queries that trigger this Instant Answer?**
anime pokemon, anime death note, naruto anime

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**
Anime lovers

**Is this Instant Answer connected to a DuckDuckHack [Instant Answer idea](https://duck.co/ideas)?**
No

**Which existing Instant Answers will this one supersede/overlap with?**
None

**Are you having any problems? Do you need our help with anything?**
- I don't know if the template used was the best option
- I couldn't properly utilize the hummingbird icon

**What are the terms of use for the API? Will DuckDuckGo need specific authorization (e.g. an API key)? Are there any costs associated with API usage?**
I don't found the terms of the API, but it is open and an API key is not required

**Where did you hear about DuckDuckHack? (For first time contributors)**
In hiring page of DuckDuckGo

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**
IA preview

## Checklist
Please place an 'X' where appropriate.

```
[x] Added metadata and attribution information
[x] Wrote test file and added to t/ directory
[x] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[x] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[x] Tested cross-browser compatibility

    Please let us know which browsers/devices you've tested on:
    - Windows 8
        [x] Google Chrome
        [x] Firefox
        [] Opera
        [x] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
[...]

PR #1918, opened 2015-06-12
Anime - Revived PR

iambibhas:
This is a revived version of #1664. Fixed the complaints on the other PR. Removed all the custom CSS and using the `movies` template.

selection_042

------

IA Page: https://duck.co/ia/view/anime


share/spice/anime/anime.js
```diff
@@ -0,0 +1,54 @@
...
+                    cover_image = DDG.toHTTP(item.cover_image);
+                }
+
+                return {
+                    img: cover_image,
+                    image: cover_image,
+                    img_m: cover_image,
+                    heading: item.title,
+                    title: item.title,
+                    abstract: item.synopsis.split(' ').slice(0,49).join(' ') + '...',
```
moollaza: The template handles shortening the content. We can just pass along the full synopsis and let the template deal with it.

share/spice/anime/anime.js
```diff
@@ -0,0 +1,54 @@
...
+                    rating: item.community_rating,
+                    url: item.url,
+                    is_retina: (DDG.is3x || DDG.is2x) ? "is_retina" : "no_retina"
```
moollaza: doesn't look like this get's used anywhere?

iambibhas:
@moollaza Fixed those 2. But there seems to be a design issue on mobile devices 
selection_046

moollaza:
@iambibhas I would just ditch the view on MyAnimeList button. There should really only be 1 button.

You could try changing the wording to "View on Hummingbird" and "View on MyAnimeList"

iambibhas:
@moollaza MyAnimeList is kind of the go to place for anime lovers for references of an anime, hence wanted to keep it. Lemme try removing that word..

abeyang:
I agree with @moollaza -- can we just reduce this down to one button?

iambibhas:
@moollaza @abeyang Done. I forgot that the hummingbird url was already there as anchor on the title. Now there is just one button.

lib/DDG/Spice/Anime.pm
```diff
@@ -0,0 +1,28 @@
...
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Anime.pm";
+topics "entertainment", "everyday";
+category "entertainment";
+attribution github => ['https://github.com/iambibhas','Bibhas'],
+           twitter => ['https://twitter.com/bibhasdn','Bibhas D'];
+
+spice proxy_cache_valid => "200 30d";
+spice to => 'https://hummingbird.me/api/v1/search/anime?query=$1';
+
+triggers startend => 'anime', 'hummingbird';
```
moollaza: I'm a little concerned with the `anime` trigger, given how broad that term is. A quick look at the autocomplete text when you type `anime` shows a lot of queries that trigger this which aren't searches for specifc anime books. Also, most of them return really irrelevant results so I think we'll need to use a relevancy block.

`anime wallpapers` for example returns tons of bad results..

iambibhas: Relevancy block on the front end might not be a good idea as there are literally thousands of animes and each of them have both japanese and english names, where the relevancy check will actually fail. Maybe we could have a list of stop words(taken from autocomplete) in the backend?

moollaza: @iambibhas sorry just saw this! I would suggest we use both methods to ensure we don't make redundant API calls, and only show relevant results to users

iambibhas: Added some stop words and relevancy block.

jagtalon:
@iambibhas cool, checking it again!

jagtalon:
@iambibhas updated the templates to hide the "0 reviews" that we're seeing. Please add `hideReviewText: true` to the `options` object. :)

jagtalon:
Up on https://comserv.duckduckgo.com/?q=naruto+anime&ia=anime! @abeyang is currently reviewing it.  Also please change the API endpoint from HTTPS to HTTP (a little faster that way, I think)

iambibhas:
@jagtalon -

> Please add `hideReviewText: true` to the `options` object. :)

Only if there is no review?

> Also please change the API endpoint from HTTPS to HTTP

Shouldn't we prefer HTTPS for almost everything? Just curious.

jagtalon:
@iambibhas when it comes to endpoints it doesn't really matter because it's all proxied by us so if it's HTTP it will be always HTTPS for the users because it goes through us first. It's also a weee bit faster on our end. :)

jagtalon:
@iambibhas there are no reviews for everything right?

iambibhas:
@jagtalon done.

jagtalon:
Hey @iambibhas. @abeyang suggested that we get rid of the detail pane since the data is a bit underwhelming. So now clicking on the tile should take you to the site itself.

What do you think?

iambibhas:
@jagtalon @abeyang it has episode count, each episode length, rating, gist and links to MyAnimeList. It's "good to have" in my opinion.

abeyang:
@iambibhas @jagtalon it's not so much that it's underwhelming -- I do think it's good to have -- it's that it would take a lot of work to just get it to look right. Internally, we're working on the detail pane templates ("DPT" -- which is coming along quite nicely!), so my suggestion is to hold off on this for now, and revisit once the DPT is finished.

iambibhas:
@abeyang it looks good now. How about we let it roll now and revisit when new DPT is deployed and it starts looking ugly? We'll anyway have to revisit a lot more spices when the new DPT is here, right? Is the DPT coming very soon? Then it makes sense.

Just curious, what makes it hard for this one to look good on the new template? I'm sure there are other IA with details panes with even lesser amount of data.

abeyang:
@iambibhas I'm fine with releasing this as-is, and revisiting once the new DPT is released. /cc @moollaza 

The whole point is that it should be **easy** for this (as well as all detail panes) to look good when the new template comes out, not harder! We've had a tough time in the past getting some DP's to look good, which is why we're doing the whole DPT thing.

moollaza:
@abeyang alright sounds good to me! :+1:

moollaza:
@iambibhas congrats! It's live: https://duckduckgo.com/?q=naruto+anime&ia=anime

PR #1919, opened 2015-06-12
today_in_history - updated to new spice template

### LINKED ISSUE #1917: Today In History:
I can't use this instant answer i don't know if it's due of the API or not
------
IA Page:  http://duck.co/ia/view/today_in_history

iambibhas:
Fixes #1917.

selection_043


moollaza:
There's a big PR that's almost done which greatly improves this IA. The current implementation uses XML which it shouldn't as it causes problems. Closing this in favour of the existing PR.

PR #1922, opened 2015-06-13
npm - checking dependency length before showing, added trigger "npm install"

### LINKED ISSUE #1881: NPM: Make sure we have dependencies to show before adding to InfoBox
Looks like this module doesn't have any dependencies? Or potentially there's something else going on?

npm_getconfig_at_duckduckgo

---

IA Page: https://duck.co/ia/view/npm

iambibhas:
Fixes #1881.

 - Checking if there is any dependency before showing it in infobox.
 - I sometimes end up searching for `npm install foo`, thought it might be helpful to support it as trigger.
 - Added project homepage and repo url as `url` in infobox

selection_045

### DIFF
--- a/lib/DDG/Spice/Npm.pm
+++ b/lib/DDG/Spice/Npm.pm
@@ -15,6 +15,7 @@ attribution github  => ['https://github.com/remixz', 'zachbruggeman'],
             twitter => ['https://twitter.com/zachbruggeman', 'zachbruggeman'];
 
 triggers startend => 'npm';
+triggers start => 'npm install';
 
 spice to => 'http://registry.npmjs.org/$1/latest';
 spice wrap_jsonp_callback => 1;
--- a/share/spice/npm/npm.js
+++ b/share/spice/npm/npm.js
@@ -61,15 +61,17 @@
                     });
                 }
 
-                var dependencies = $.map(item.dependencies, function(val, key) {
-                    return key + " (" + val + ")";
-                }).join(", ");
-
-                if (dependencies !== "") {
-                    boxData.push({
-                        label: "Dependencies",
-                        value: dependencies
+                if (item.dependencies) {
+                    var dependencies = $.map(item.dependencies, function(val, key) {
+                        return key + " (" + val + ")";
                     });
+
+                    if (dependencies.length + 0) {
+                        boxData.push({
+                            label: "Dependencies",
+                            value: dependencies.join(", ")
+                        });
+                    }
                 }
 
                 return {
--- a/share/spice/npm/npm.js
+++ b/share/spice/npm/npm.js
@@ -61,11 +61,11 @@
                     });
                 }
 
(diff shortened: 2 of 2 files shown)


share/spice/npm/npm.js
```diff
@@ -44,22 +45,23 @@
...
                     boxData.push({
                         label: "Source",
-                        value: item.dist.tarball
+                        value: item.dist.tarball,
+                        url: item.dist.tarball
                     });
                 }
 
-                if (item.dependencies) {      
+                if (Object.keys(item.dependencies).length > 0) {
```
moollaza: Looks like `Object.keys` doesn't support Ie 8. You can use `DDG.getProperty(item, "dependencies")` which will prevent it from croaking if the prop doesn't exist.

lib/DDG/Spice/Npm.pm
```diff
@@ -15,6 +15,7 @@ attribution github  => ['https://github.com/remixz', 'zachbruggeman'],
             twitter => ['https://twitter.com/zachbruggeman', 'zachbruggeman'];
 
 triggers startend => 'npm';
+triggers start => 'npm install';
```
moollaza: :+1:

share/spice/npm/npm.js
iambibhas: The problem is, in case of no dependency, `dependencies` does exist but has no key or value. The only way to check if it's empty is to check the number of keys.

moollaza:
LGTM :+1:

jagtalon:
@moollaza good to merge?

share/spice/npm/npm.js
```diff
@@ -44,37 +45,40 @@
...
                     var dependencies = $.map(item.dependencies, function(val, key) {
                         return key + " (" + val + ")";
-                    }).join(", ");
-
-                    boxData.push({
-                        label: "Dependencies",
-                        value: dependencies
                     });
+
+                    if (dependencies.length + 0) {
```
jagtalon: Why is there a `+ 0`?

share/spice/npm/npm.js
```diff
@@ -44,37 +45,40 @@
...
                     boxData.push({
                         label: "Source",
-                        value: item.dist.tarball
+                        value: item.dist.tarball,
+                        url: item.dist.tarball
                     });
                 }
 
-                if (item.dependencies) {      
+                if (item.dependencies) {
```
jagtalon: Can we do something like:

```javascript
if(item.dependencies && item.dependencies.length > 0) {
    ...
}
```

iambibhas: @jagtalon `item.dependencies` is a JSON object, not an array.

share/spice/npm/npm.js
iambibhas: Sorry it was a typo. :| Fixed it.

moollaza:
LGTM :+1:

Issue #1864, opened 2015-05-25
Gravatar, Keybase, Zodiac, Twitter : Switch to Icon template group


moollaza:
To Be Converted:

- [x] Gravatar (assigned to @altern8tif)
- [x] Keybase
- [ ] Zodiac (images assets: http://cl.ly/19171G2D2H21)
- [x] Twitter

---

We now have the [Icon template group](https://duck.co/duckduckhack/templates_overview#icon-template-group), which allows us to easily re-create the looks of these IA's without the need for any custom templates.

The icon template can display 3 lines of text if you specify the `title`, `subtitle` and `altsubtitle` accordingly.

The `subtitle` and `altsubtitle` can be a single string, or an `array` of `string`'s or `objects`. If an array is provided, the template will automatically separate the elements. To create separated links, an array of objects can be provided like so:

```js
normalize: function(item){
    subtitle: [{
        href: "https://website.com"
        text: "My Website"
    }, {
        href: "https://other.site.com",
        text: "Other Site"
    }, {
        'This 3rd element is plain text!"
    }]
```

This would produce:

```html
<a href="https://websitecom">My Website</a>
<span class="sep"></span>
<a href="https://other.site.com">Other Site</a>
<span class="sep"></span>
This 3rd element is plain text!
```

All of this is covered in detail in the [documentation](https://duck.co/duckduckhack/templates_overview#icon-template-group)

altern8tif:
If no one is working on this, I don't mind helping out for the Gravatar spice...

moollaza:
@altern8tif awesome! Thanks a lot :+1:

altern8tif:
@moollaza while we're on this, I'm presuming the Twitter spice needs to be converted too?

altern8tif:
Just another thought: 

Since the above spices always return a single result, wouldn't it be better to use the `basic_icon_detail` template instead of using the Icon template group?  The Icon template group seems more suited for IAs which return multiple results.

Pardon the ignorance.  Still feeling my way around here.

altern8tif:
@moollaza done for Gravatar.  Let me know if this works.

![screen shot 2015-06-14 at 12 41 38 pm](https://cloud.githubusercontent.com/assets/5793364/8147163/db11454c-1292-11e5-84b8-cc1fb289039b.png)

moollaza:
> I'm presuming the Twitter spice needs to be converted too?
[...]

PR #1925, opened 2015-06-14
Gravatar - switch to Icon template group

altern8tif:
- amend normalize and templates in gravatar.js
- remove unnecessary css
- add test cases
- remove content.handlebars

- address #1864


altern8tif:
screen shot 2015-06-14 at 12 41 38 pm

moollaza:
@altern8tif this is looking great.

@abeyang this is now using the icon template. I suspect you and Chris will want the links to be grey? If so, we should change the template CSS ;)

altern8tif:
Added `variants` attributes so that custom CSS can be removed.

share/spice/gravatar/gravatar.js
```diff
@@ -20,77 +16,64 @@
...
                         return entry.name.givenName + " " + entry.name.familyName;
                     }
                     return entry.displayName;
                 }
+                // Get current location
+                function getCurrentLocation(entry) {
+                    if (entry.currentLocation) {
+                        return entry.currentLocation;
+                    } else {
+                        return "No Location Specified";
```
moollaza: Let's rather not display anything if the info isn't there. Better to `return null`

share/spice/gravatar/gravatar.js
```diff
@@ -20,77 +16,64 @@
...
+                    var accounts = [];
+                    if (entry.accounts) {
+                        $.each(entry.accounts, function() {
+                            accounts.push({
+                                text: DDG.capitalize(this.shortname),
+                                href: this.url
+                            });
+                        });
+                    } else {
+                        accounts = ["Additional Information Unavailable"];
```
moollaza: Same here, let's not show anything if it doesn't exist.

moollaza:
@abeyang this one is ready for a design pass, it's using the Icon template.

share/spice/gravatar/gravatar.js
altern8tif: OK done!

share/spice/gravatar/gravatar.js
altern8tif: OK done!

abeyang:
@altern8tif , it's looking pretty good! However, can we make a few more tweaks, such that:

- We should show the gravatar handle, not just the social networks the user is on
- Instead of showing *all* the social networks, can we just show the 3 main ones, and have a "X More" link at the end (which points to their main gravatar site)?
- Also, can we utilize a smaller avatar size -- let's use the **medium icon** and **medium-sized title**, found on our [styleguide](https://duckduckgo.com/styleguide)

In the end, it should look like this (don't worry about the link colors -- we'll fix that):

gravatar

Thanks!

altern8tif:
@abeyang how about showing the social network icon/favicon next to the handle?  This helps to identify which social network the handle belongs to... As some pple might have the same handles for different social networks.

abeyang:
@altern8tif I don't think that's necessary, because the query itself is for gravatar. It's similar when you do a [twitter](https://duckduckgo.com/?q=twitter+matt&ia=social) search. Besides, the favicon is already showing at the bottom, so I think we're good.

altern8tif:
@abeyang Ah I think I misunderstood you the first time.  Sure, will make the changes as per your attached screenshot :smiley:

altern8tif:
@abeyang Let me know if this works...

moollaza:
@altern8tif LGTM :+1:

moollaza:
@altern8tif crap, just found a bug. Try "gravatar bob" we're trying to pass an object as as string:
gravatar_bob_at_duckduckgo

moollaza:
This is one is even worse:
gravatar_tom_at_duckduckgo

Let's use a relevancy block to require that the user at least has a valid username so we have something to show!

share/spice/gravatar/gravatar.js
```diff
@@ -20,77 +16,71 @@
...
+                    }
+                }
+                // Get array of social media accounts
+                function getAccounts(entry) {
+                    var accounts = [];
+                    var maxAccounts = 4;
+                    if (entry.accounts) {
+                        $.each(entry.accounts, function() {
+                            accounts.push({
+                                text: DDG.capitalize(this.shortname),
```
moollaza: Seems like we need to make sure these are defined

altern8tif:
@moollaza Sorry my bad... I should have done further testing on other usernames.

The thing about Gravatar is that there are numerous accounts (like [tom](https://en.gravatar.com/tom.json)) where the account is totally empty... It'll be funny if everything were empty.

Should I default to showing the same info as the existing live Gravatar IA (see screenshot)?

screen shot 2015-07-04 at 2 29 36 am

PR #1929, opened 2015-06-16
Changes SeatGeek::Sports name to Tickets

moollaza:
Prevents conflict with coming Sports IA.

/cc @sdougbrown @b1ake


sdougbrown:
Tickets :+1:

PR #1932, opened 2015-06-17
Update Forecast variableTileWidth => itemsWidthVaries

bsstoner:
I've been trying to namespace all the item-specific meta options with `items` at the beginning. This depends on an internal change.

@sdougbrown


sdougbrown:
:+1:

PR #1296, opened 2014-11-29
People In Space, Improve Information

MrChrisW:
The new API endpoint `http://www.howmanypeopleareinspacerightnow.com/peopleinspace.json` offers images of each person and a bio. This means we can improve the overall look of the People In Space IA providing more information. As suggested by @moollaza 

I've experimented with different template types but settled for using the `products` group as was used for the [Congress Spice](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/congress/congress.js)

This is a work in progress. Issues are on my list to fix.

* [x] ~~Check if information is returned from the API before populating the template `Country: ____`.~~
* [x] ~~Include bio link in details model.~~
* [x] ~~For some silly reason the API returns `italy` lowercase :grimacing:~~
* [x] Improve tile styling including! can the images be reduced in size before load? One image is 422kb :grimacing: 
* [x] ~~Fix name moving over tile on mobile view~~
* [x] ~~Include link to astronauts Twitter~~

**Experimenting with the base template**
image

**Current products template**
image

image

PR #1934, opened 2015-06-18
Google Plus – Spice API, template updates, CSS amendments

### LINKED ISSUE #1896: Convert IAs to newest Spice API
There are still many IA's in the repo that use the old API, that aren't wrapped in a closure, and most importantly aren't calling `Spice.failed()` when they trigger.

We need these IA's to be updated and they need to call `Spice.failed()` when necessary so the other IA's waiting to show their answers know that the triggered Spice isn't going to produce anything.

Here's a list of the IA's that use the API:

- [x] forecast ( should be wrapped in a closure ) 
- [ ] editor
- [ ] search_code (offline, needs update)
- [ ] detect_lang
- [x] images
- [ ] game_info (offline?)
- [x] rhymes
- [x] videos
- [ ] smbc (offline. Needs reliable/official API)
- [ ] songkick_getartistid (offline)
- [ ] ~~drinks~~ (API has gone offline :disappointed: )
- [ ] maven
- [x] google_plus
- [ ] stopwatch
- [x] mass_on_time
- [ ] news
- [ ] forvo (offline)

------

Some of these are currently offline as they were never updated and re-deployed. We'd love to have them go live if they're able to!

The IA's that are currently functioning are a higher priority!

### LINKED ISSUE #1928: GooglePlus: Content height causing border to overflow
Looks like the CSS for this IA needs adjusting. The combination of set heights and padding is causing the border to overflow on highlight.

This needs to be fixed!

google_plus_doug_at_duckduckgo_and_1____projects_zeroclickinfo-spice__ssh_

------
https://duck.co/ia/view/googleplus

altern8tif:
- Adjust css to remove border overflow (fixes #1928)
- Update to latest Spice API (fixes #1896)
- Update to `product` template group

_Some comments_: 

Had an awful time debugging the `product_item` attributes.  Seemed like the `title` attribute didn't work, but managed to find one that did – `heading`.  

Someone should look into the templates to see if the DuckDuckHack documentation matches up to it.  Would definitely help the newbies like me to understand how things are set up...

### DIFF
--- a/lib/DDG/Spice/GooglePlus.pm
+++ b/lib/DDG/Spice/GooglePlus.pm
@@ -11,9 +11,9 @@ code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/
 topics "everyday", "social";
 category "ids";
 attribution github => ['https://github.com/jagtalon','jagtalon'],
-            twitter => ['http://twitter.com/juantalon','jagtalon'];
-            github => ['https://github.com/altern8tif', 'altern8tif'];
-            twitter => ['http://twitter.com/altern8tif', 'altern8tif'];
+            twitter => ['http://twitter.com/juantalon','jagtalon'],
+            github => ['https://github.com/altern8tif', 'altern8tif'],
+            twitter => ['http://twitter.com/altern8tif', 'altern8tif']
 
 spice to => 'https://www.googleapis.com/plus/v1/people/?query=$1&key={{ENV{DDG_SPICE_GOOGLE_PLUS_APIKEY}}}&callback={{callback}}&maxResults=12';
 spice proxy_ssl_session_reuse => "off";
--- a/lib/DDG/Spice/GooglePlus.pm
+++ b/lib/DDG/Spice/GooglePlus.pm
@@ -13,7 +13,7 @@ category "ids";
 attribution github => ['https://github.com/jagtalon','jagtalon'],
             twitter => ['http://twitter.com/juantalon','jagtalon'],
             github => ['https://github.com/altern8tif', 'altern8tif'],
-            twitter => ['http://twitter.com/altern8tif', 'altern8tif']
+            twitter => ['http://twitter.com/altern8tif', 'altern8tif'];
 
 spice to => 'https://www.googleapis.com/plus/v1/people/?query=$1&key={{ENV{DDG_SPICE_GOOGLE_PLUS_APIKEY}}}&callback={{callback}}&maxResults=12';
 spice proxy_ssl_session_reuse => "off";
--- a/lib/DDG/Spice/GooglePlus.pm
+++ b/lib/DDG/Spice/GooglePlus.pm
@@ -12,6 +12,8 @@ topics "everyday", "social";
 category "ids";
 attribution github => ['https://github.com/jagtalon','jagtalon'],
             twitter => ['http://twitter.com/juantalon','jagtalon'];
+            github => ['https://github.com/altern8tif', 'altern8tif'];
+            twitter => ['http://twitter.com/altern8tif', 'altern8tif'];
 
 spice to => 'https://www.googleapis.com/plus/v1/people/?query=$1&key={{ENV{DDG_SPICE_GOOGLE_PLUS_APIKEY}}}&callback={{callback}}&maxResults=12';
 spice proxy_ssl_session_reuse => "off";
--- a/share/spice/google_plus/google_plus.css
+++ b/share/spice/google_plus/google_plus.css
@@ -1,19 +1,12 @@
-.tile--googleplus .tile__media {
-    height: 12em;
+.tile--google_plus .tile__media {
+    height: 10em;
 }
 
-.tile--googleplus  .tile__media__img {
-    margin-top: 1.6em;
+.tile--google_plus  .tile__media__img {
+    margin-top: 1.4em;
     border-radius: 100%;
 }
 
-.tile--googleplus  .tile__title--pr {
+.tile--google_plus  .tile__title {
     text-align: center;
-    font-size: 1.4em;
-    font-weight: 400;
-    height: 2.5em;
-}
-
-.tile--googleplus .tile--pr__sub {
-    display: none;
 }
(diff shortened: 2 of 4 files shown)


moollaza:
@altern8tif just saw your comment about `products_simple` -- sorry about the confusion. As mentioned over there, the `products_simple` group was deprecated and recently brought back as an interim fix. The Media group will ultimately replace it (and a few others potentially) and so I would suggest you use that.

It should simplify things a bit as it takes the standard `title`, `subtitle`, and `image`. It also allows for a `footer` sub-template, but I don't think it'll be needed here.

altern8tif:
screen shot 2015-06-23 at 10 24 07 am

share/spice/google_plus/google_plus.js
```diff
@@ -1,33 +1,38 @@
-function ddg_spice_google_plus (api_result) {
+(function (env) {
     "use strict";
+    env.ddg_spice_google_plus = function(api_result){
 
-    if(!api_result || !api_result.items || api_result.items.length === 0) {
-        return Spice.failed("googleplus");
-    }
+        if(!api_result || !api_result.items || api_result.items.length === 0) {
+            return Spice.failed("google_plus");
```
moollaza: :+1:  good catch!

lib/DDG/Spice/GooglePlus.pm
```diff
@@ -12,6 +12,8 @@ topics "everyday", "social";
 category "ids";
 attribution github => ['https://github.com/jagtalon','jagtalon'],
             twitter => ['http://twitter.com/juantalon','jagtalon'];
```
moollaza: Looks like you need to remove the `;` at the end of this line. It's throwing an error.

share/spice/google_plus/google_plus.css
```diff
@@ -1,24 +1,12 @@
...
 
-.tile--googleplus .tile__body {
-    padding: .3em 1em;
-}
-
-.tile--googleplus  .tile__title {
+.tile--google_plus  .tile__title {
     text-align: center;
-    font-size: 1.4em;
-    font-weight: 400;
```
moollaza: I think we should keep the font-weight. We can address the size using the `elClass` property...

altern8tif: No problem. Will use the `elClass` property in the `js` file instead =)

share/spice/google_plus/google_plus.js
```diff
@@ -1,33 +1,38 @@
...
+            templates: {
+                group: 'media',
+                item_detail: false,
+                detail: false,
+                options: {
+                    moreAt: true
+                },
+                variants: {
+                    tile: 'narrow'
+                }
```
moollaza: Let's specify the font size:

```js
variants: {
   ...
},
elClass: {
   tileTitle: "tx--17"
}
```

lib/DDG/Spice/GooglePlus.pm
```diff
@@ -11,7 +11,9 @@ code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/
 topics "everyday", "social";
 category "ids";
 attribution github => ['https://github.com/jagtalon','jagtalon'],
-            twitter => ['http://twitter.com/juantalon','jagtalon'];
+            twitter => ['http://twitter.com/juantalon','jagtalon'],
+            github => ['https://github.com/altern8tif', 'altern8tif'],
+            twitter => ['http://twitter.com/altern8tif', 'altern8tif']
```
moollaza: hehe looks like you're missing the `;` here now ;)

altern8tif: My bad. heh.

altern8tif:
@moollaza Is this good to go?

abeyang:
@altern8tif it's getting there! Can we do a couple more fixes:

1. Can we vertically-center align the title -- similar to my comment here: https://github.com/duckduckgo/zeroclickinfo-spice/pull/1296#issuecomment-117798975
2. Can we add the handle (+douglas) as a subtitle? (But looking at G+ right now, it seems like they're trying to *hide* their handles?)

altern8tif:
@abeyang 

1. No problemo!
2. Yea it looks like there's no much info in the JSON response.  Only `displayName` and `image` are meaningful.  Upon further reading, it seems that the API requires the end-user to [authorize via OAuth](https://developers.google.com/+/web/api/rest/index#authorization) in order to get additional user profile [properties](https://developers.google.com/+/web/api/rest/latest/people#resource).  So I guess that's not gonna be of any use for DDG then.

moollaza:
> So I guess that's not gonna be of any use for DDG then.

Yup, Spices can't deal with OAuth right now so that's not an option for this IA.

abeyang:
@altern8tif @moollaza sounds good. Let's just go with the vertical-center alignment then.

altern8tif:
The title is now vertical-aligned.  :v:

But the original IA seems to be having weird intermittent issues with the display of the profile image.  Not sure if this is due to the API or something internal within DDG.  @abeyang @moollaza Are you guys seeing this issue too?

screen shot 2015-07-09 at 4 40 42 am

altern8tif:
OK looks like it might be due to the HTTPS Anywhere Chrome extension I have on my browser.

@abeyang just some fine tuning of the css.  Looks good to go?

screen shot 2015-07-10 at 9 18 38 am

abeyang:
@altern8tif looks good! :+1:

share/spice/google_plus/google_plus.js
```diff
@@ -1,33 +1,41 @@
...
-            detail: false,
-            variants: {
-                tile: 'narrow'
+        Spice.add({
+            id: 'google_plus',
+            name: 'Social',
+            data: api_result.items,
+            meta: {
+                sourceName : 'Google+',
+                sourceUrl : 'http://plus.google.com',
```
moollaza: This should really link to a search for the query. Can fix in another PR or I can make an issue.

moollaza:
:+1:

share/spice/google_plus/google_plus.js
altern8tif: @moollaza thanks for spotting... Will make a separate PR on this.

PR #1941, opened 2015-06-19
Wikinews: Switch API call to HTTPS to prevent redirect

moollaza:
/cc @jagtalon @iambibhas


share/spice/wikinews/wikinews.js
```diff
@@ -10,6 +10,7 @@
             Spice.add({
                 id: "wikinews",
                 name: "Wikinews",
+                signla: "high",
```
jagtalon: Typo with "signal"

Issue #1867, opened 2015-05-25
DogoNews: Switch to new media template group, implement new design


moollaza:
This IA should now be switched to use the the `media` template group so it can also display a picture.

Along with displaying a picture, the body of the IA should be updated to meet the new design:

![dogo-news-v1 3](https://cloud.githubusercontent.com/assets/873785/7801981/9589ccac-02fb-11e5-9bf7-6c4a5871ecf0.jpg)

/cc @dogomedia

moollaza:
We'd love to see this one get converted!

joe1chen:
@moollaza We've switch to using the media template and can get the thumbnail and the description to show.  Any tips on getting the footer to look like the screenshot that you have (with the little icons)?

gokul1794:
@moollaza this can be closed?

gokul1794:
@moollaza This is good to close too!

moollaza:
@gokul1794 thanks!

PR #1943, opened 2015-06-20
DogoNews: Switch to new media template group, implement new design

dogomedia:
This is the first pass at fixing #1867.


dogomedia:
The current footer matches the existing news tile's footer.  Not sure how to add the icons to match the screenshot @moollaza added to #1867 

screen shot 2015-06-19 at 10 39 27 pm

moollaza:
@dogomedia this is looking awesome!

>Not sure how to add the icons to match the screenshot

Sorry about that, docs are underway for icon usage. In the interim I can just tell you what to do :)

Please use `<span class="ddgsi ddgsi-clock"></span>` for the clock and `<span class="ddgsi ddgsi-user"></span>` for the person icon.

You can also use the classes `pull-left` and `pull-right` to float the author to the left and the time to the right.

share/spice/dogo_news/dogo_news.js
```diff
@@ -28,12 +28,16 @@
...
                         image: thumb,
                         relative_time: moment(item.published_at).fromNow()
                     };
                 },
                 templates:{
-                    item: 'news_item'
+                    group: 'media',
+                    options: {
+                        footer: Spice.dogo_news.footer
+                    }
```
moollaza: Please use a `variants` object to increase the site of the tile snippet:

```js
options: {
  ...
},
variants: {
    tileSnippet: "large"
}
```

share/spice/dogo_news/footer.handlebars
```diff
@@ -0,0 +1,8 @@
+<div class="dogo_news_tile_footer">
```
moollaza: We actually wrap the footer in a div with class `tile__foot`. We can probably remove this wrapping div and adjust the CSS above

joe1chen:
@moollaza Updated footer with icons and cleaned up the css.  Also added ```tileSnippet: "large"```.  This now looks much closer to the screenshot you added.

screen shot 2015-06-22 at 3 14 07 pm

One more remaining question is how you generally prefer to handle the ellipsis for text that is too long?  We used ```source: Handlebars.helpers.ellipsis(item.author, 14)``` to truncate the source to 14 characters.  Is this how you generally prefer to do ellipsis?

dogomedia:
@moollaza Updated footer with icons and cleaned up the css.  Also added ```tileSnippet: "large"```.  This now looks much closer to the screenshot you added.

screen shot 2015-06-22 at 3 14 07 pm

One more remaining question is how you generally prefer to handle the ellipsis for text that is too long?  We used ```source: Handlebars.helpers.ellipsis(item.author, 14)``` to truncate the source to 14 characters.  Is this how you generally prefer to do ellipsis?

moollaza:
>Is this how you generally prefer to do ellipsis?

I hadn't considered those being too long. That should be fine. Alternatively, you could go the CSS route because this is a single line of text, but the JS route is simple and definitely compatible for IE 8+

BTW to increase the size of the snippet further, you can set `meta: { snippetChars: 200 }` which should round out the snippet at about 4 lines.

moollaza:
@abeyang this one is ready for a design review :+1:  It's now using the media template!

dogomedia:
@moollaza Thanks for the tips. We've increased the snippet to 4 lines as you suggested.

screen shot 2015-06-22 at 4 21 12 pm

moollaza:
@dogomedia whoops, that looks like 5 lines! Let's trim that down so we have some space between the footer and snippet :+1:

dogomedia:
@moollaza Good catch.  Dropping it to around 110 chars seems to work best to keep it to 4 lines in both the standard view as well as the grid view.

moollaza:
@dogomedia thanks! Once @abeyang gives us the OK this should be good to merge :+1:

abeyang:
@dogomedia looking good! I'll go more into the visual styles in a bit, but first--

1. How come Meera Dolasia's name is truncated so early? Can we just try to do **first name / last initial**, because otherwise, almost all the authors' names would be truncated...
2. We shouldn't use the detail pane for this; clicking on the tile should just jump straight to the page (like before).

More comments coming soon!

dogomedia:
@abeyang 
1. The name was truncated so early at 14 chars because in the grid view the name would push the time into a second row.  But your suggestion to use first name /last initial is a good one.
2. You are right, the detail pane should not be used.

We will fix both issues immediately.

abeyang:
@dogomedia looking better! Now, for the visuals, let's do this:

1. Let's lose the icons for now, and see how they'll look
2. For the footer (`tile__foot`), let's add this class to it as well: `tx-clr--grey-light`. Should make the footer text become lighter.
3. And for the css, let's change this:

```css
.zci--dogo_news .tile--dogo_news .tile__foot {
  // remove the existing padding-top.. don't think it's doing anything
  // add this instead, to move the footer up a bit:
  padding-bottom: 0.2em;
}
```

Alrighty, hope that all makes sense!

dogomedia:
@abeyang We've removed the icons now.  Regarding 2) we were not sure how to add a custom class to the footer since the ```Spice.dogo_news.footer``` footer we specify in our template options is the content of the ```tile_foot```.  So instead we just added the same grey color to our custom css (```color: #aaa;```) to achieve the same effect.  

screen shot 2015-06-26 at 11 44 18 am

moollaza:
@dogomedia we actually have a new mechanism that let's you apply classes to certain elements:

```js
templates: {
    variants: {
        tileFoot: "tx-clr--grey-light"
    }
}
```

The benefit of our special colour classes is that they change colours based on the chose DDG UI theme, so that way we don't need to specify other colours for the dark theme, or others.

dogomedia:
@moollaza We're now using the ```elClass``` ```tileFoot``` to set the footer class.

moollaza:
@dogomedia excellent. Thanks :+1:

abeyang:
Looks great! :+1:

Issue #1946, opened 2015-06-22
Stocks: Convert to use new moreAtText property


moollaza:
We now have a "moreText" property which can be defined in `meta.option`. We should use is here so the template doesn't have to manually create the More at link anymore:

Example
```js
id: ...,
name: ...,
meta: {
   ...
   options: {
      moreText: {
        url: "https://website.com",
        text: "Link text"
      }
   }
}
```

------

Documentation will be live shortly. PR here: https://github.com/duckduckgo/duckduckgo-documentation/pull/269

jagtalon:
Fixed by #1949

PR #1949, opened 2015-06-23
Stocks: Use template.options.moreText property for attribution 

MrChrisW:
Fixing: #1946 

@moollaza Can you clarify if the moreText hyperlink key is `url` or `href` ? also this doesn't seem to work yet. 

Currently the spices uses `Data by Xignite`, are we able to add "Data by" or is it not needed?

````perl
meta: {
    sourceName: 'NASDAQ',
    sourceUrl: url,
    options: {
        moreText: {
            href: 'https://xignite.com',
            text: 'Xignite'
        }
    }
}
```
![selection_447](https://cloud.githubusercontent.com/assets/5282264/8305677/1b73069a-19e4-11e5-8ec0-b0a5223e033a.png)

### DIFF
--- a/share/spice/stocks/stocks.js
+++ b/share/spice/stocks/stocks.js
@@ -16,7 +16,12 @@
                 meta: {
                     sourceName: 'NASDAQ',
                     sourceUrl: url,
-                    attributionText: "Xignite"
+                    options: {
+                        moreText: { 
+                            href: 'https://xignite.com', 
+                            text: 'Xignite' 
+                        }
+                    }
                 },
                 normalize: function(data){
                     var change = data.ChangeFromPreviousClose,
@@ -52,7 +57,7 @@
                     group: 'base',
                     options: {
                         content: Spice.stocks.content,
-                        moreAt: false
+                        moreAt: true,
                     }
                 }
             });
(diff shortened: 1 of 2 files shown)


moollaza:
@MrChrisW my mistake, `href` is correct.

>Currently the spices uses Data by Xignite, are we able to add "Data by" or is it not needed?

The text for the link should remain as it is: `Data by Xignite`

MrChrisW:
@moollaza :+1:

moollaza:
@MrChrisW sorry, there was a misunderstanding on my part about how this is supposed to be used. `moreText` should actually be nested in `templates.options` and then it will work as expected.

share/spice/stocks/stocks.js
```diff
@@ -52,7 +57,7 @@
                     group: 'base',
                     options: {
                         content: Spice.stocks.content,
-                        moreAt: false
+                        moreAt: true,
```
moollaza: Please move `moreText` into this `options` block.

MrChrisW:
@moollaza :+1:

jagtalon:
Ah, looks like `moreText` isn't available for base templates! Made a PR internally and tested this on it so this is good. :+1:

PR #1951, opened 2015-06-23
Fix border css

gokul1794:
Border set to transparent #1857


moollaza:
@gokul1794 thanks! It looks like the first element still has a top border though! Let's make sure that one is transparent as well!

moollaza:
In general, when updating the UI, it's best to take a screenshot so we can quickly assess the changes :)

moollaza:
My mistake, I was looking at the wrong code!

This LGTM :+1:

gokul1794:
Is this IA not live btw?

gokul1794:
Sorry, my mistake I forgot to add the .com while searching

Issue #1944, opened 2015-06-22
SublimePackages: should strip "for" from query


moollaza:
This IA is not yet, live, but will be soon!

During some final testing I noticed that the query "sublime text packages for json" triggers and returns decent results, but the API call we make sends "for json" as the search term, when it should really be "json".

Let's strip any potential filler words from the query.

------
IA Page: https://duck.co/ia/view/sublime_packages

MrChrisW:
@moollaza Any other words that need to be filtered? 7e1d2e7

moollaza:
@MrChrisW none that I can think of right now. Actually, I see some queries come up through auto-complete that mention the version of sublime text. That might be a separate problem because I don't think we can specify that to the API?

MrChrisW:
@moollaza It looks like we can specify the version using the search term :`st2` or `:st3` also operating system `:win` `:osx` `:linux`

I'll add this functionally :+1:

jagtalon:
Fixed in #1953

PR #1953, opened 2015-06-24
SublimePackages: Remove common words, add filtering options

MrChrisW:
Fixing  #1944

Currently only removing the word `for` 

Filtering by Operating System (windows, mac os x, linux)
Filtering by Sublime version 2/3

//cc @moollaza

### DIFF
--- a/lib/DDG/Spice/SublimePackages.pm
+++ b/lib/DDG/Spice/SublimePackages.pm
@@ -24,12 +24,12 @@ my $skip = join "|", share('skipwords.txt')->slurp(chomp => 1);
 handle remainder => sub {
     return unless $_; 
     # OS search filters
-    s/linux/:linux/g; 
-    s/windows/:win/g;
-    s/mac\s?os\s?x|os\s?x/:osx/g;
+    s/\blinux\b/:linux/g; 
+    s/\bwindows\b|\bwin\b/:win/g;
+    s/\bmac\s?os\s?x\b|\bos\s?x\b/:osx/g;
     # Version search filters
-    s/version 2/:st2/g;
-    s/version 3/:st3/g;
+    s/v?(ersion )?2/:st2/g;
+    s/v?(ersion )?3/:st3/g;
     # Do not trigger IA if query matches any words in skipwords.txt file
     return if  m/$skip/i;
     s/\b(for)\s+?\b//g; #skip common words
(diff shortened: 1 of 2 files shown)


lib/DDG/Spice/SublimePackages.pm
```diff
@@ -24,8 +23,16 @@ my $skip = join "|", share('skipwords.txt')->slurp(chomp => 1);
 
 handle remainder => sub {
     return unless $_; 
+    # OS search filters
+    s/linux/:linux/g; 
+    s/windows/:win/g;
+    s/mac\s?os\s?x|os\s?x/:osx/g;
+    # Version search filters
+    s/version 2/:st2/g;
+    s/version 3/:st3/g;
```
MrChrisW: @moollaza I've set these to only match exactly `version [2-3]`, maybe this could be expanded?

lib/DDG/Spice/SublimePackages.pm
```diff
@@ -24,8 +23,16 @@ my $skip = join "|", share('skipwords.txt')->slurp(chomp => 1);
 
 handle remainder => sub {
     return unless $_; 
+    # OS search filters
+    s/linux/:linux/g; 
+    s/windows/:win/g;
+    s/mac\s?os\s?x|os\s?x/:osx/g;
```
MrChrisW: Should probably convert searches with `win` or `osx` to `:win` and `:osx`

Also add `\b\ boundary ?

lib/DDG/Spice/SublimePackages.pm
moollaza: @MrChrisW this seems fine -- maybe s/v?(ersion)?2/.../` ? e.g. "sublime text 2", "sublime text v3"

This may be premature optimization, but I think those are reasonable for now.

lib/DDG/Spice/SublimePackages.pm
moollaza: >Should probably convert searches with win or osx to :win and :osx

:+1: 

>Also add \b\ boundary ?

Definitely :)

MrChrisW:
@moollaza Any ideas of how to resolve this small issue. 

The first query result contains the word  `packages` as the [trigger word](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1953/files#diff-2c7fdc1b886aebe0d4920228fd8ea9c7R20) `sublime text packages` is skipped and `sublime text` used this means the reminder will contain `packages`. However I think it should be treated as just a search for `php` 

I don't want to strip the word `packages` someone could be searching for a `package` [sublime text plugin](https://packagecontrol.io/search/packages). Hope I've made sense :+1: 

**Query        _________________________              |               ____     Result**
 1. `sublime text v3 packages php` = `%3Ast3%20packages%20php`


 2. `sublime text packages php` = `php`

jagtalon:
@MrChrisW this is tricky since things like https://packagecontrol.io/packages/Package%20Syncing are also popular. Also what about "sublime text package control json"?

Maybe an improvement would be to just trigger with "sublime text" and try to decide based on the remainder what to do with it.

This PR is good to go though. :+1:

PR #1954, opened 2015-06-24
BrainyQuote: Prevent triggering on 'stock quote'

### LINKED ISSUE #1952: BrainyQuote: Don't trigger on queries containing "stock quote"
In testing the Stock IA, I noticed that brainy quote triggers for any query that has "quote" and so it triggers on queries containing "stock quote" as well.

We should detect this in the `handle` function and bail early to prevent redundant API calls.

------
https://duck.co/ia/view/brainy_quote

kbhat95:
Seems rather inelegant, but it fixes #1952. Is there a better way to do this?

Also, I saw that this Spice was missing a tests file, so I added that in as well.

### DIFF
--- a/lib/DDG/Spice/BrainyQuote.pm
+++ b/lib/DDG/Spice/BrainyQuote.pm
@@ -27,9 +27,10 @@ handle remainder => sub {
         s/\./\. /g;
     }
     
-    # Hacky fix to avoid triggering on 'stock quote'
-    # 'stock quote' is handled by the Stocks IA
-    s/stock//g;
+    # Avoid triggering on 'stock' quotes; these are handled by Stocks IA
+    if ($req->query_lc =~ m/stock quote/) {
+       return;
+    }
 
     return $_ if $_;
     return;
--- a/lib/DDG/Spice/BrainyQuote.pm
+++ b/lib/DDG/Spice/BrainyQuote.pm
@@ -26,6 +26,10 @@ handle remainder => sub {
     if(/^\w\.\w/) {
         s/\./\. /g;
     }
+    
+    # Hacky fix to avoid triggering on 'stock quote'
+    # 'stock quote' is handled by the Stocks IA
+    s/stock//g;
 
     return $_ if $_;
     return;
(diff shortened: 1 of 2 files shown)


moollaza:
>Also, I saw that this Spice was missing a tests file, so I added that in as well.

@kbhat95 I'm really surprised we were missing a test! Thanks a lot for making one.

lib/DDG/Spice/BrainyQuote.pm
```diff
@@ -26,6 +26,10 @@ handle remainder => sub {
     if(/^\w\.\w/) {
         s/\./\. /g;
     }
+    
+    # Hacky fix to avoid triggering on 'stock quote'
+    # 'stock quote' is handled by the Stocks IA
+    s/stock//g;
```
moollaza: I guess I wasn't very clear, but this won't catch the case where the query is "apple stock quote" or similar.

So what we really want to do is check if the original query contains the word "stock quote" and if so, bail out with a `return;`

You can use the `$req` object which contains several useful properties, including `$req->query` which provides the original query and `$req->query_lc` which provides the lower-cased query.

So I suggest you run a regex against that for `m/stock quote/` and `return;` if it's true.

kbhat95: Oh, I see.

I'll get this done. :+1:

kbhat95:
@moollaza funny story: it turns out that the 'apple stock quote' query also triggers the Recipes Spice. Should this be another issue?

lib/DDG/Spice/BrainyQuote.pm
```diff
@@ -26,6 +26,11 @@ handle remainder => sub {
     if(/^\w\.\w/) {
         s/\./\. /g;
     }
+    
+    # Avoid triggering on 'stock' quotes; these are handled by Stocks IA
+    if ($req->query_lc =~ m/stock quote/) {
+       return;
+    }
```
iambibhas: Or you could just check if `$_` equals to `stock`.

kbhat95: Hmm, yes.

@moollaza @iambibhas , is there a significant gain in using that method, or does ```$req->query_lc``` do the trick? Disclaimer: new to Perl. :P

jagtalon:
This looks good :+1:

PR #1956, opened 2015-06-25
Timer will not trigger at five-alarm

### LINKED ISSUE #1937: Timer shouldn't trigger on "five-alarm" query
https://duckduckgo.com/?q=five-alarm&ia=timer seems out of place!

gokul1794:
This should fix #1937

### DIFF
--- a/lib/DDG/Spice/Timer.pm
+++ b/lib/DDG/Spice/Timer.pm
@@ -19,7 +19,7 @@ triggers start => ['time', 'timer for'];
 spice call_type => 'self';
 
 handle remainder => sub {
-    return if lc($req->query_raw) =~ /^time($|[^r ][\S]+$)/;  #filter out queries with like "time" or "Time::Piece"
+    return if lc($req->query_raw) =~ /^time($|[^r ][\S]+$)|five-/i;  #filter out queries with like "time" or "Time::Piece"
     return unless /^( ?([\d.]+ ?(m(in((ute)?s?)?)?|s(ec((ond)?s?)?)?|h(ours?)?|hr)|online) ?)+$/ || $_ eq '';
     return '';
 };
--- a/lib/DDG/Spice/Timer.pm
+++ b/lib/DDG/Spice/Timer.pm
@@ -19,7 +19,7 @@ triggers start => ['time', 'timer for'];
 spice call_type => 'self';
 
 handle remainder => sub {
-    return if lc($req->query_raw) =~ /^time($|[^r ][\S]+$)|five-/i;  #filter out queries with like "time" or "Time::Piece"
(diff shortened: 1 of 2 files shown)


moollaza:
@gokul1794 close, but not quite the right approach. For example "six-alarm" will still trigger this. In fact "<anything>-alarm" will trigger the IA.

The bigger problem is that we're not ensuring the remainder is something we can use. If there is a remainder in the query, we need to ensure it's something useful.

gokul1794:
@moollaza `\[a-z]+-|-alarm\ig` will not trigger on  `<something>-alarm` and `-alarm`
Is that good enough?

moollaza:
@gokul1794 that fix is still a little too niche because we have other `startend` trigger words. For example, "test-countdown" will still trigger this.

We need a better regex for qualifying the query. When the trigger is "alarm", "countdown" or "timer" the trigger word needs to be the only thing in the query, or it needs a remainder.

Something like:

```perl
m/(^|\w+ \s)(?:alarm|countdown|timer)($|\s \w+)
```

should work? We really need a more robust set of tests that outline the expected triggering behaviour for all our trigger words to make sure this is correct.

gokul1794:
Sorry! I didn't take into consideration the others because the issue said
only alarm, I will take a look at it.

gokul1794:
Sorry @moollaza ! I didn't take into consideration the others because the issue mentioned alarm, I will take a look at it.

jagtalon:
@gokul1794 any updates? :)

gokul1794:
@jagtalon whoops, I'm not so good with regexs, I'll give it another shot though :+1:

mattr555:
Maybe the best course of action is to break it down like the [color codes](https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/ColorCodes.pm#L36) goodie. We might want to cut back the trigger space as well. I personally don't see an advantage for the `time` trigger at the start, especially if it is causing problems with other unrelated triggers.

moollaza:
@gokul1794 @mattr555 I agree. Let's kill the `time` trigger word for now. This IA is causing problems and its triggering needs to be reigned in. It's triggering on tons of random stuff:

`time/`, `-time`, https://duckduckgo.com/?q=http%3A%2F%2Fdennybritz.com%2Fblog%2F2013%2F08%2F05%2Fwhy-i-will-be-a-solo-founder-next-time%2F (partially our IA triggering code's fault, but the timer code should prevent this too!)

jagtalon:
@gokul1794 ping!

moollaza:
@gokul1794 I'm closing this one for now, but if you'd like to finish this up, we'll gladly open it back up for you!

Issue #1765, opened 2015-04-14
LeakDB: Template group should be switched to 'list'


moollaza:
Now that we have the `list` template group, this IA (and others!) can be switched over to use it.

https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/leak_db/leak_db.js#L27

- It's already setup to use the `record` template which is perfect. Switching to 'list' group is beneficial because it will automatically supress the output so it doesn't take up too much vertical space. The "Show more" button will allow users to expand the content to see it all.
- We should also remove the row-highlight option

It also wouldn't hurt to add a `title` and possible `subtitle` if we have something worth putting there :)

The end result will look almost identical to the Whois Spice: https://duckduckgo.com/?q=whois+duckduckgo.com&ia=whois

gokul1794:
@moollaza This can be closed!

moollaza:
@gokul1794 thanks!

PR #1961, opened 2015-06-27
Make leakdb IA use the list group

marianosimone:
- Deleted the css, as the list group seems to take care of everything needed
- Added a title
- Removed conversion to lowercase (as the hashing is different for different casings)
- Excluded the "plain text" version from the result, as it's the same as the title, and what the user searched for

## Before
before

## After
leakdb

This should take care of #1765


share/spice/leak_db/leak_db.js
```diff
@@ -23,8 +24,13 @@
                 sourceUrl: 'http://leakdb.abusix.com/?q='+ encodeURIComponent(query),
                 sourceName: 'Abusix'
             },
+            normalize: function(item) {
+                return {
+                    title: 'hashes for "' + api_result.query + '"'
```
marianosimone: I'm not sure if I like this title, but couldn't think of a better one. Any idea?

moollaza: Perhaps a title like "LeakDB Hashes" and subtitle: "Key: MySuperNicePassword"?

/cc @abeyang -- any thoughts on the title design

marianosimone:
@abeyang: taking @moollaza's idea into consideration, I updated the title to show:

leakdb_new

lib/DDG/Spice/LeakDB.pm
```diff
@@ -22,7 +22,7 @@ triggers startend => 'hashme', 'leakdb';
 spice to => 'http://api.leakdb.abusix.com/?j=$1';
 spice wrap_jsonp_callback => 1;
 
-handle remainder_lc => sub{
+handle remainder => sub{
```
moollaza: I think you'll want to keep `remainder_lc`, otherwise the remainder won't be lowercased when it's passed into your regex below, making your IA case sensitive.

moollaza: Nevermind, I see now that this IA has actually been broken for a long time because of the `_lc`! We've been sending the API the **lowercased** input all this time...

We can actually remove that regex as I believe it's not needed at all. We're dealing with the remainder of the query, which is expected to be just a single "word". We should really check that there are no spaces in the remainder and if not `return $_`.

share/spice/leak_db/leak_db.js
```diff
@@ -12,7 +12,8 @@
                         query : api_result.type + ' hash';
 
         api_result.hashes = api_result.hashes[0];
-        
+        delete api_result.hashes.plaintext;
```
moollaza: I realize now that plaintext is used for showing what the plaintext version of a password is, when the user enters a hash as the lookup.

E.g. 
- "leakdb MySuperNicePassword" => MD5: 8ea6794b65f326b437bd2d562ea10700
- "leakdb 8ea6794b65f326b437bd2d562ea10700" => Plaintext: MySuperNicePassword

marianosimone: @moollaza you are right... reverted this!

lib/DDG/Spice/LeakDB.pm
marianosimone: @moollaza I removed the regex (you were right, it's already the remainder, so doesn't make sense to try to clean it up)

However, I'm not sure about the white space thing... wouldn't by "my super password" a valid password?

marianosimone:
@moollaza @abeyang anything else to improve this one to make it ready to merge?

abeyang:
Sorry @marianosimone -- I've been swamped lately. From what I understand, main purpose of this is to show whether or not a hashed portion of the key phrase is stashed somewhere? If that's the case, wouldn't we want to highlight that?

Also, we probably want a short description of some sort. When I first saw this, I thought this was just a mass hashing IA (but nothing more than that).

Let me know if I understand the purpose of this IA correctly; I'll be working on a design in the meantime.

One thing you could do for me is to lose the zebra stripes; let's make it more similar to the [whois IA](https://duckduckgo.com/?q=whois+cnn.com&ia=whois).

marianosimone:
@abeyang actually, the IA just sends the key to [abusix](https://leakdb.abusix.com), which figures out if it's plain text or looks like a hash, and hashes it with various methods. Even if the key is NOT in their database as a "leaked" one, the hashes are still returned, so it doesn't really give you information about the password itself.

@moollaza @abeyang BTW, I've removed the zebra stripes and updated the screenshot in the description of the pull request

moollaza:
>Even if the key is NOT in their database as a "leaked" one, the hashes are still returned, so it doesn't really give you information about the password itself.

But, if you search for a password hash, it *does* tell you if it was found or not. When it was not found, the `found` property is set to false. I don't think we currently handle that case?

E.g. http://api.leakdb.abusix.com/?j=d06f9d1c9398d581003767df6f0fa35d

From: https://duckduckgo.com/?q=md5+G%232g6Kg%257VzH&t=ffab&ia=answer

marianosimone:
@moollaza we don't handle that case... Funny thing is, once someone searches for a password, it gets added to their database, so if you now follow http://api.leakdb.abusix.com/?j=d06f9d1c9398d581003767df6f0fa35d, you'll see that it IS found.

I think we might be able to improve the overall experience, but maybe deserves to have a separate ticket and PRs

moollaza:
>Funny thing is, once someone searches for a password, it gets added to their database, so if you now follow http://api.leakdb.abusix.com

Yes I realized that after playing around with it. I suspect that's a bug? Or rather we should always be calculating the SHA512 of the input (like they do) and query that so we never send any plaintext passwords.

We have plans to build a unified "Cryptography" Instant Answer that will let user's choose which has they want to calculate for a given plain text input. So I think this IA should be redesigned to accept passwords and tell the user if it's compromised or not.

>I think we might be able to improve the overall experience, but maybe deserves to have a separate ticket and PRs

I agree. I'll merge this for now. Would you like to continue working on this in a new PR?

marianosimone:
@moollaza sure.. I'd love to keep working on improving this IA. However, I'm not 100% sure of what approach to take. If you want, open a new issue for the improvement and we can discuss over there :D

moollaza:
>I'd love to keep working on improving this IA.

Excellent!

> If you want, open a new issue for the improvement and we can discuss over there :D

:+1:

PR #1966, opened 2015-06-28
fedora_project_package_db - removed extra CSS, fixed UI

### LINKED ISSUE #1958: FedoraProjectPackge: Switch subtitle, to altSubtitle, fix footer spacing
Currently the IA uses the `subTitle` property, which forces space between the title and subtitle. Let's switch this one over to use an `altSubtitle` which will force it to stick to the bottom of the title.

Also, it looks like the footer needs some breathing room, as the snippet is nearly touching it:
fedora_package_gcc_at_duckduckgo

It should look like this:

screen_shot_2015-06-22_at_1_11_06_pm_png__png_image__774 x 337_pixels_

I don't think we'll need to use any CSS to achieve the desired look. Using `variants` to set the `tileSnippet` to large should increase the size of the tile. Similarly we can set `snippetChars` if we need to allow more text.

------
IA Page: http://duck.co/ia/view/fedora_project_package_db

iambibhas:
Fixes #1958. This is how it looks now - 

selection_053


share/spice/fedora_project_package_db/fedora_project_package_db.js
```diff
@@ -25,13 +25,17 @@
...
+                    tileTitle: '1line',
+                    tileSnippet: 'large',
                 }
             },
             normalize: function (item) {
                 return {
                     url: "https://admin.fedoraproject.org/pkgdb/package/" + item.name,
                     title: item.name,
-                    subtitle: item.description ? item.summary : "",
+                    altSubtitle: item.description ? item.summary : "",
```
moollaza: Let’s return `null` instead of an empty string so the template won't bother creating a `<span>` for the subtitle

moollaza:
@iambibhas thanks a lot, this looks great. 1 minor adjustment and it's good for merge.

iambibhas:
@moollaza done.

moollaza:
LGTM :+1:

Issue #1768, opened 2015-04-15
InTheaters: "kids movies in theaters" should show g/pg movies


moollaza:
Right now the query "kids movies in theaters" doesn't trigger the InTheater's IA. It should be considered an alias for "pg movies in theaters" or "g movies in theaters" (if "g" is supported).

- https://duckduckgo.com/?q=kids+movies+in+theaters
- https://duckduckgo.com/?q=pg+movies+in+theaters

gokul1794:
@moollaza Looks like this can be closed too!

moollaza:
@gokul1794 thanks!

PR #1968, opened 2015-06-28
make "kids" an alias for "g" in InThreaters ia

marianosimone:
"kids" seems to be a good alias for G, as requested in #1768

kids-movies


marianosimone:
I don't really like how the full implementation is currently now, but I didn't want to refactor and add the new feature at the same time.

If you agree, we could create a new ticket for the refactor and tackle it to make it easier/faster to add/change aliases

moollaza:
@marianosimone thanks for another great fix!

>If you agree, we could create a new ticket for the refactor and tackle it to make it easier/faster to add/change aliases

100% agreed. I'm sure there's a better way to handle all these triggers/aliases.

mariano-simone-zocdoc:
:+1: I'll brush up my :camel: skills and see if I come up with a refactor (BTW, I created #1970, feel free to edit the name/description)

PR #1972, opened 2015-06-29
MLB Boxscore - Whitespace Fix

sdougbrown:
Some whitespace crept into the compiled template, which makes the end result look a little skewed. :smile: 

to @bsstoner 
cc @thm

### DIFF
--- a/share/spice/sports/mlb/games/mlb_score.handlebars
+++ b/share/spice/sports/mlb/games/mlb_score.handlebars
@@ -15,7 +15,7 @@
 				{{/if}}
 			{{/if}}
 		</div>
-		{{#if has_started}}
+		{{~#if has_started~}}
 			<div class="c-score__sequence  c-score__opt">
 			{{#each score.away.innings}}
 				<div class="c-score__item  c-score__item--{{type}}">{{number}}</div>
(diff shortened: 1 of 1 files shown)


bsstoner:
Do you think we can we solve this at the compile level instead of needing to add all the extra syntax to the templates? /cc @zachthompson @russellholt

russellholt:
I explored removing all leading white space from precompiled templates once - it worked (seemed to) and saved otherwise unminifiable space but it needs this large scale regression testing before I'd trust it. Is this what you mean?

zachthompson:
Are you referring to the indentation in the rendering:
white-space
I think this is [fixed in handlebars 3.x](https://github.com/wycats/handlebars.js/issues/858):
no-white-space1
We could, as Russell suggests, remove leading white space pretty easily.

bsstoner:
We're doing it with the internal templates. Seems like maybe we could compile the same way in open source to avoid confusion? Where is the code that compiles the templates for open source?

Internally we're using a grunt task in core with a step that removes new lines and spaces. `/www-release/frontend/Gruntfile.jsL54`

zachthompson:
Handlebars templates are compiled here - https://github.com/duckduckgo/p5-dist-zilla-plugin-buildshareassets/blob/master/lib/Dist/Zilla/Plugin/BuildShareAssets.pm#L281

sdougbrown:
Merging now to fix the immediate issue with MLB scores - I agree that we should address the issue globally, though.

Issue #1585, opened 2015-03-02
SeatGeek: Maybe "amanda palmer tickets" should be the same as "amanda palmer concerts"


jagtalon:
Not sure how hard this is to do especially since "tickets" could either mean concerts or sports or actually any event.

https://duck.co/ia/view/seat_geek_events_by_artist

marianosimone:
I made a small change in #1975 that takes care of this.

However, there's an alternative for this solution, that would involve merging [EventsByArtist](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SeatGeek/EventsByArtist.pm) with [Sports](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SeatGeek/Sports.pm), and deciding if the `taxonomies.name`(as documented [here](http://platform.seatgeek.com/#events)) should be included or not, depending on some keywords to be present (e.g. "game", "match" vs "concert", "gig")

russellholt:
'tour' and 'tour dates' would improve this, also.

moollaza:
@marianosimone I believe we attempted to have a unified Instant Answer that looked through the taxonomies but we didn't end up going down that route.

@jagtalon do you remember why?

PR #1975, opened 2015-06-30
use "tickets" as trigger for EventsByArtist IA

marianosimone:
As reported in #1585, "tickets" was not a trigger for concerts

### DIFF
--- a/lib/DDG/Spice/SeatGeek/EventsByArtist.pm
+++ b/lib/DDG/Spice/SeatGeek/EventsByArtist.pm
@@ -22,7 +22,8 @@ triggers startend =>
     'live',
     'live show',
     'live shows',
-    'gigs';
+    'gigs'
+    'tickets';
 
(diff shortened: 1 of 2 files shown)


marianosimone:
There's an alternative for this solution, that would involve merging [EventsByArtist](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SeatGeek/EventsByArtist.pm) with [Sports](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SeatGeek/Sports.pm), and deciding if the `taxonomies.name`(as documented [here](http://platform.seatgeek.com/#events)) should be included or not, depending on some keywords to be present (e.g. "game", "match" vs "concert", "gig")

moollaza:
LGTM :+1:

moollaza:
@marianosimone thanks again!

PR #1977, opened 2015-06-30
Improve the hackage IA.

### LINKED ISSUE #1950: HackagePackages: Find new data source, re-implement IA with templates
This IA was using a JSON file hosted on GitHub as it's source. The file has since been removed so I've taken the IA offline as it is now throwing a console error.

This IA should be revamped entirely. Let's use a more official data source that allows for actual package search so we don't have to dig through a large JSON file each time the IA triggers.

Once we have the API figured out, this IA should be re-implemented using our built-in templates. There no reason this one should need a custom template anymore :+1: 

------

IA Page: https://duck.co/ia/view/hackage

cskksc:
Fetches packages or methods based on signatures for Haskell.

Instant access to package names and versions. Or method signatures.

https://www.haskell.org/hoogle/
The hackage api doesn't support json responses for search operations.

It provides accurate information about packages and their versions. Is widely used in the haskell community.

[Hayoo](http://hayoo.fh-wedel.de/) is also comparable. But the library version information is missing. So I think [Hoogle](https://www.haskell.org/hoogle/) is better.

```hask json```, ```hask [a]-> a```, ```map hask```

Haskell programmers.

Not sure. There's one specific for Hayoo and not generic haskell packages. [Idea](https://duck.co/ia/view/hayoo)

hackage, hayoo. #1950 

Not right now.

No authorization required. And its free.

Friends

- Package listing:
  package

- Method signature
  method_sig
## Checklist

```
[] Added metadata and attribution information

    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 8

    - Mac OSX
        [x] Google Chrome
        [x] Firefox
        [] Opera
        [x] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [x] Firefox
        [] Native Browser
        [x] Google Chrome
        [] Opera
```

### CODE REVIEW

**Comment by moollaza** [comment]
>Which existing Instant Answers will this one supersede/overlap with?
hackage, hayoo. #1950 

@cskksc I'm not a Haskell user, so I'm not quite sure which of these IA's/Sources are necessary. Do you think with this IA merged, the others are still useful? Should we remove them?

**Comment by cskksc** [comment]
The functionality offered by this is very similar to the Hayoo IA. But IMHO the results from Hoogle are better than Hayoo. For example, [hayoo list](https://duckduckgo.com/?q=hayoo+list) vs [hask list](http://twist-joker.codio.io:5000/?q=hask+list&ia=hackage). Also, Hoogle returns version numbers for packages which would allow us to sort them. Plus hoogle is a part of haskell.org. Having an api for the actual [Hackage](https://hackage.haskell.org/) site would have been nice. But its search doesn't support json responses, only HTML. https://hackage.haskell.org/api#search .

File: share/spice/hackage/hackage.js [hunk 1]
```diff
@@ -0,0 +1,97 @@
...
+
+            normalize : function(item) {
+                var v = version(item.docs);
+                var info = {
+                    title: item.self,
+                    description: docString(item.docs),
+                    url: item.location
+                };
+                if (v) {
+                    info.subtitle = "Version: " + v;
```
moollaza: please switch to `info.altsubtitle` this will prevent the space between the title and subtitle

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,97 @@
...
+            name: "Hackage",
+            data: results,
+            meta: {
+                sourceName: "Hackage",
+                sourceUrl: "https://hackage.haskell.org/packages/search?terms=" + query
+            },
+            templates: {
+                group: 'text',
+                detail: false,
+                item_detail: false
```
moollaza: Let's define some variants to format things a little better:

```js
templates: 
  variants: {
    tileTitle: "2line",
    tileSnippet: "large"
  }
}
```

moollaza:
@cskksc this is off to a great start. I see now that we're parsing the "docs" portion of the result to get what we need -- is there no API endpoint with better structured data? Can we ask the API maintainers to improve the API for us?

Also, I see a few results have no `docs` string -- perhaps we should skip these results so we don't a mostly blank tile. Alternatively we could put `-` for the Version and `No description available` so it's more clear why it's missing.

@abeyang any thoughts on design? We don't have anything more available to use beside the title, version and snippet it seems.

cskksc:
@moollaza @abeyang Made the changes to the ui. And skipped the results with empty ```docs``` field. This is how it looks now,

ui

I have also raised an [issue](https://github.com/ndmitchell/hoogle/issues/120) on the hoogle repository. Lets see if we can get any help in improving the api.

abeyang:
@cskksc Is there any more info provided by the API, such as number of downloads or last updated? Would be nice to have it more consistent with other similar IAs:

sublime-v1 1

cskksc:
@abeyang Not right now. But the API creator has been very helpful and confirmed that they are working on a release of Hoogle 5.0 which would have an extensive API. Then we can further improve the results for packages and method signatures both. I'm not sure about the release date. https://github.com/ndmitchell/hoogle/issues/120

abeyang:
@cskksc Ok, if that's the case, then instead of using `.tile__content--lg` (associated with `.tile__content`), can we use `.tile__content--m`? There's too much empty space under the snippets, as it currently is.

cskksc:
@abeyang Yeah that looks better. I couldn't find any documentation on how to assign css classes to html elements. Could you point me to some resources ?

moollaza:
>I couldn't find any documentation on how to assign css classes to html elements. Could you point me to some resources ?

@cskksc sorry about that! It's a little bit buried in the docs but we cover it here: https://duck.co/duckduckhack/templates_reference#directly-specifying-classes

If you look through the repo for `elClass` you'll be able to find other working examples.

---

However, the usual way to set the class for elements like the snippet are to use the Variants:

```js
variants: {
   tileSnippet: 'medium'
}
```

But it seems we haven't implemented the "medium" option for some reason. So setting that with `elClass` will work.

We also have an optional `snippetChars` property you can use to control the number of characters allowed in the snippet, but those classes will determine the height of the snippet div.

cskksc:
@moollaza @abeyang Ah, my bad. I thought elClass is only used for text & colours. Now I've applied the ```tile__content--m``` class. Looks like this.

screen shot 2015-07-14 at 4 51 35 pm

moollaza:
@cskksc terribly sorry, but it looks like @abeyang was mistaken about the existence of a `tile__content--m` class. We don't actually have one, but in modifying the `--lg` class on the element, it made it look like it existed as the height changed.

It looks like we'll need to set an explicit height for the tiles via CSS. Sorry for the back and forth but can you set the tiles to a height of `10em` using CSS?

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,108 @@
...
+            }
+            // Assuming semantic version groups do not exceed 4
+            toAppend = 4 - v.length;
+            v = v + Array(toAppend + 1).join("0");
+            return parseInt(v);
+        }
+
+
+        // limit display results to 30
+        var results = api_result.results;
```
jagtalon: Check if `api_result` exists first before accessing `api_result.results`.

```javascript
if(!api_result || api_result.error) {
    return Spice.failed('hackage');
}
```

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,108 @@
...
+        if (results.length > 30)
+            results = results.splice(0,30);
+
+        // Get the original query
+        var script = $('[src*="/js/spice/hackage/"]')[0];
+        var source = $(script).attr("src");
+        var query = source.match(/hackage\/([^\/]*)/)[1];
+
+        if (api_result.error) {
```
jagtalon: Does the API actually return an `error` property if things don't go well? Also, see https://github.com/duckduckgo/zeroclickinfo-spice/pull/1977/files#r34880825.

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,108 @@
...
+        // converts "1.2.3" to 1230
+        // converts "0.0.1" to 10
+        // helps in comparing version strings in different format.
+        function versionInt(v) {
+            if (v == undefined) {
+                return -1;
+            }
+            while(v.indexOf(".") != -1) {
+                v = v.replace(".","");
+            }
```
jagtalon: What about `"1.2.3".replace(/\./g, "")`?

lib/DDG/Spice/Hackage.pm
```diff
@@ -0,0 +1,16 @@
+package DDG::Spice::Hackage;
+# ABSTRACT: Search for information about Hackage packages
+
+use DDG::Spice;
+
+triggers startend => 'hask';
+
+spice to => 'http://www.haskell.org/hoogle/?mode=json&hoogle=$1';
```
jagtalon: This redirects to `https://www.haskell.org/hoogle/?mode=json&hoogle=` (note the HTTPS part). Our Nginx config doesn't handle redirects at the moment, so we can't have any endpoints that redirect. Please change "HTTP" to "HTTPS".

lib/DDG/Spice/Hackage.pm
```diff
@@ -0,0 +1,16 @@
+package DDG::Spice::Hackage;
+# ABSTRACT: Search for information about Hackage packages
+
+use DDG::Spice;
+
+triggers startend => 'hask';
```
jagtalon: Trigger on "hackage", "haskell", "hayoo" or "hoogle"?

share/spice/hackage/hackage.js
jagtalon: Also, we should probably check if `results` has a length greater than 0

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,108 @@
...
+            return Spice.failed('hackage');
+        }
+
+        Spice.add({
+            id: "hackage",
+            name: "Hackage",
+            data: results,
+            meta: {
+                sourceName: "Hackage",
+                sourceUrl: "https://hackage.haskell.org/packages/search?terms=" + query,
```
jagtalon: Shouldn't this link to `https://www.haskell.org/hoogle/?hoogle=` since we're using Hoogle?

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,108 @@
...
+        // helps in comparing version strings in different format.
+        function versionInt(v) {
+            if (v == undefined) {
+                return -1;
+            }
+            while(v.indexOf(".") != -1) {
+                v = v.replace(".","");
+            }
+            // Assuming semantic version groups do not exceed 4
+            toAppend = 4 - v.length;
```
jagtalon: When declaring a new variable, add `var`.

cskksc:
@moollaza @abeyang Thats perfectly fine. It did seem that ```tile__content--m``` change had that effect on the sizes :smile: . I've changed the height to ```10em``` now. Also added a chars limit of 85. Looks like this,

screen shot 2015-07-17 at 4 38 31 pm

cskksc:
@jagtalon I've made the changes according to your comments. The ```hask``` trigger was temporary to avoid conflicts with [Hayoo IA](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Hayoo.pm#L15).

cskksc:
@abeyang ping.

share/spice/hackage/hackage.css
```diff
@@ -0,0 +1,3 @@
+.tile__body {
+    height: 10em;
```
moollaza: @cskksc looks like this is causing problems? It's cutting off the bottom of the tile for me:

hackage_json_at_duckduckgo

Removing the CSS causes this to look OK:

hackage_json_at_duckduckgo_and_1____projects_zeroclickinfo-goodies__ssh_

cskksc:
@moollaza Yeah, it looks bad. Was looking ok on my machine. Have removed the CSS.

moollaza:
@cskksc thanks!

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+        var query = source.match(/hackage\/([^\/]*)/)[1];
+
+        Spice.add({
+            id: "hackage",
+            name: "Hoogle",
+            data: results,
+            meta: {
+                sourceName: "Hoogle",
+                sourceUrl: "https://www.haskell.org/hoogle/?hoogle=" + query,
+                snippetChars: 85
```
moollaza: Lets increase this to `180`, there's lot more text we can show!

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+            },
+            templates: {
+                group: 'text',
+                detail: false,
+                item_detail: false,
+                variants: {
+                    tileTitle: "2line"
```
moollaza: let's set a large variant for the snippet so the height of the tiles is consistent:

`tileSnippet: "large"`

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+                }
+            },
+
+            normalize : function(item) {
+                var v = version(item.docs);
+                var description = docString(item.docs);
+                if (description == "") {
```
moollaza: Seems like this might not be working correctly? Let's make sure the description also contains more than 1 character:

Looks like it has no description:
hackage_math_at_duckduckgo

Description is too short:
hackage_json_at_duckduckgo

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+            v = v + Array(toAppend + 1).join("0");
+            return parseInt(v);
+        }
+
+
+        if(!api_result || api_result.results.length == 0) {
+            return Spice.failed('hackage');
+        }
+        var results = api_result.results;
+        if (results.length > 30)
```
moollaza: Any chance there's an API param that will limit us to 30 results?

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+//    {"location"=>"http://hackage.haskell.org/package/JSON-Combinator",
+//     "self"=>"package JSON-Combinator",
+//     "docs"=> "A combinator library on top of a generalised JSON type.\n\nVersion 0.2.8 "}]}
+
+(function (env) {
+    "use strict";
+    env.ddg_spice_hackage = function(api_result) {
+
+        // get documentation string
+        function docString(doc) {
```
moollaza: Let's define these private functions outside the callback, above `env.ddg_spice_hackge`. We're working in a closure so they'll be accessible in the `ddg_spice_hackge` but won't get added to the global `env` :smiley:

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+            return doc.replace(/Version .*/,"");
+        }
+
+        // get version from docs.
+        // supported version formats: [1.2, 1.2.3, 1.2.3.4]
+        function version(doc){
+            try {
```
moollaza: I think we can do away with the `try..catch` block here:

```js
var version = doc.match(/\d\.\d\.?\d?\.?\d?/);
return version || undefined;
```

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+            catch (err) {
+                return undefined;
+            }
+        }
+
+        // converts "1.2.3" to 1230
+        // converts "0.0.1" to 10
+        // helps in comparing version strings in different format.
+        function versionInt(v) {
+            if (v == undefined) {
```
moollaza: Let's use `===`

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,102 @@
...
+                return -1;
+            }
+            v = v.replace(/\./g,"")
```
moollaza: Doesn't seem like we need a regex here to match `"."`?

moollaza: Ah nevermind, I see the `/g`. :+1:

share/spice/hackage/hackage.js
cskksc: It was ```large``` before, but @abeyang suggested that it should be medium because of the empty space in many tiles. So we wanted  ```medium``` tile-height, but there's no CSS class for it. Then we settled on using ```height: 10 em```. But having both the ```height``` and ```snippetChars``` seems to be causing the text to cut off like that, so I have removed the css. So only ```snippetChars: 170``` should be used then ?

cskksc:
@moollaza hey, I've made the other changes, have removed ```snippetChars``` and set tile height to small for now.

moollaza:
@cskksc we actually wanted the taller tile (`"large"`) and higher charsnippet (`175`) so we can show more! Let's go with that :smiley:

share/spice/hackage/hackage.js
moollaza: @cskksc sorry for the confusion. Let's go with the "Large" Snippet, and "170" char snippet. That way the text fills up more space and we don't have consistent gaps below the snippets.

/cc @abeyang

cskksc:
@moollaza Okay, now tile snippet size is ```large```, & char limit is ```170```. It did fill up the empty space. Looks like this,
screen shot 2015-08-11 at 9 05 12 am

abeyang:
Looks good :+1: . Can we change the tab title to something more generic, like "Package" or "Software"?

cskksc:
@abeyang Cool :smiley:. Changed it to "Software", as not all results would be packages.

lib/DDG/Spice/Hackage.pm
```diff
@@ -0,0 +1,16 @@
+package DDG::Spice::Hackage;
+# ABSTRACT: Search for information about Hackage packages
+
+use DDG::Spice;
+
+triggers startend => 'haskell', 'hoogle', 'hackage';
```
moollaza: Wondering if we should add "haskell packge"? I'm not sure if users are likely to have the word "package" in their query or not.

Alternatively we could strip it out of the remainder if think it's going to be there.

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,94 @@
...
+// supported version formats: [1.2, 1.2.3, 1.2.3.4]
+function version(doc){
+    var version = doc.match(/\d\.\d\.?\d?\.?\d?/);
+    return version === null ? undefined : version[0];
+}
+
+// converts "1.2.3" to 1230
+// converts "0.0.1" to 10
+// helps in comparing version strings in different format.
+function versionInt(v) {
```
moollaza: Whoops, these private functions need to be inside our closure (`(function(env) { ... } ()`), but outside our callback (`env.ddg_spice_hackage`).

Let's move them down, so they aren't declared in the global namespace.

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,94 @@
...
+    v = v + Array(toAppend + 1).join("0");
+    return parseInt(v);
+}
+
+
+(function (env) {
+    "use strict";
+    env.ddg_spice_hackage = function(api_result) {
+
+        if(!api_result || api_result.results.length == 0) {
```
moollaza: Should probably add `!api_result.results ||` in the middle here.

share/spice/hackage/hackage.js
```diff
@@ -0,0 +1,94 @@
...
+                    tileTitle: "2line",
+                    tileSnippet: "large"
+                }
+            },
+
+            normalize : function(item) {
+                var v = version(item.docs);
+                var description = docString(item.docs);
+                if (description === "" || description.length < 2) {
+                    return;
```
moollaza: this should be `return null`

moollaza:
@cskksc this is looking great. I've made some final comments and then this should be good for merge! Thanks for doing a tremendous job :+1:

lib/DDG/Spice/Hackage.pm
cskksc: Lets remove words "package" or "packages" from the query to be safe.

cskksc:
Thanks for all the help @moollaza. My first open source contribution has been to DDG and the experience has been great ! :smiley:

moollaza:
@cskksc sorry for the delay! This LGTM :+1:

cskksc:
Cool :smile:. Right now, Hayoo IA and this one shares a trigger, "haskell". This won't cause any problems right ?

moollaza:
@cskksc do we still need hayoo? Is it better when it comes to returning results for the "haskell" trigger? We could narrow the triggers for hayoo, or let it only trigger on "hayoo"?

cskksc:
@moollaza IMO "hackage" source is better, but I'm not a regular haskell user so I can't say for sure. May be we can take some feedback on the haskell mailing list, after this IA is live. I'm changing the trigger to only "hayoo" for now. If both IA's are working simultaneously, on what parameters would one get preference ? latency, no of results etc ?

cskksc:
Follow up [PR](https://github.com/duckduckgo/zeroclickinfo-spice/pull/2121).

moollaza:
@cskksc to answer your question:

>If both IA's are working simultaneously, on what parameters would one get preference ? latency, no of results etc ?

It's a race for Spices. Whichever calls `Spice.add()` first get's to display first, assuming nothing else, more important also tries to show an answer

PR #1980, opened 2015-06-30
Custom Item Detail Wrappers

sdougbrown:
Item detail templates are now going to require an extra wrapping div.  This makes handling the templates internally a lot easier.  This shouldn't be a problem going forward since it's generally easier to use the `base_item_detail` template rather than crafting a custom one. :smile: 

to @bsstoner


sdougbrown:
Internal PR merged. :+1:

PR #1982, opened 2015-06-30
Icon - find icons from IconFinder

iambibhas:
Finds icons from iconfinder.com for queries like `human icon` or `free human icon`. It also detects if free/premium is mentioned and returns relevant results.

There are hundreds and thousands on icons, free and premium, that developers/designers need on daily basis. Current way is to search for icon related websites, find them, search for required term, find icon, download/buy. This spice tries to bring the whole thing down to one/two steps.

[iconfinder.com](http://iconfinder.com)

They have the best collection of icons.

I'm not aware of any.

`cloud icons`, `cloud free icon` etc

Most probably designers/developers/people who make presentations and maybe more.

Yes. https://duck.co/ideas/idea/14/search-icons-via-iconfinder-iconfinder-com

None.

Nope.

None. IconFinder's search API is public.

selection_059
selection_060
## Checklist

```

    - Ubuntu
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [] IE 10

    - Android
        [X] Firefox
        [X] Native Browser
        [X] Google Chrome
        [] Opera
```

https://duck.co/ia/view/iconfinder

### DIFF
--- a/lib/DDG/Spice/Icon.pm
+++ b/lib/DDG/Spice/Icon.pm
@@ -1,0 +1,52 @@
+package DDG::Spice::Icon;
+# ABSTRACT: Find icons for the given term
+
+use DDG::Spice;
+
+# Caching - https://duck.co/duckduckhack/spice_advanced_backend#caching
+spice is_cached => 1;
+
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
+name "Icon";
+source "IconFinder";
+icon_url "https://api.iconfinder.com/favicon.ico";
+description "Find icons for the given term";
+primary_example_queries "first example query", "second example query";
+secondary_example_queries "optional -- demonstrate any additional triggers";
+category "reference";
+topics "web_design";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Icon.pm";
+attribution github => ["iambibhas", "Bibhas D"],
+            twitter => "bibhasdn";
+
+# API endpoint - https://duck.co/duckduckhack/spice_attributes#spice-codetocode
+spice to => 'https://api.iconfinder.com/v2/icons/search?query=$1&style=&vector=all&premium=$2';
+spice wrap_jsonp_callback => 1;
+
+spice from => '(.+)/(.+)';
+
+# Triggers - https://duck.co/duckduckhack/spice_triggers
+triggers startend => "icon", "icons", "free icon", "icon free",
+                     "premium icon", "icon premium";
+
+# Handle statement
+handle query_lc => sub {
+    $_ =~ s/\s?icon\s?//g;
+    my $premium = 'all';
+
+    if (index($_, 'free') != -1) {
+        $premium = '0';
+        $_ =~ s/\s?free\s?//g;
+    } elsif (index($_, 'premium') != -1) {
+        $premium = '1';
+        $_ =~ s/\s?premium\s?//g;
+    }
+
+    # for queries like `free premium icons`
+    $_ =~ s/\s?(free|premium)\s?//g;
+
+    return unless $_;
+    return $_, $premium;
+};
+
+1;
--- a/lib/DDG/Spice/Icon.pm
+++ b/lib/DDG/Spice/Icon.pm
@@ -31,7 +31,7 @@ triggers startend => "icon", "icons", "free icon", "icon free",
 
 # Handle statement
 handle query_lc => sub {
-    $_ =~ s/\s?icon\s?//g;
+    $_ =~ s/\s?icons?\s?//g;
     my $premium = 'all';
 
     if (index($_, 'free') != -1) {
--- a/lib/DDG/Spice/Icon.pm
+++ b/lib/DDG/Spice/Icon.pm
@@ -3,33 +3,28 @@ package DDG::Spice::Icon;
 
 use DDG::Spice;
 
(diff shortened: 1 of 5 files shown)


jagtalon:
@iambibhas sweet! Checking it out

lib/DDG/Spice/Icon.pm
```diff
@@ -0,0 +1,47 @@
...
+attribution github => ["iambibhas", "Bibhas D"],
+            twitter => "bibhasdn";
+
+spice to => 'https://api.iconfinder.com/v2/icons/search?query=$1&style=&vector=all&premium=$2';
+spice wrap_jsonp_callback => 1;
+
+spice from => '(.+)/(.+)';
+
+triggers startend => "icon", "icons", "free icon", "icon free", "icons free",
+                     "premium icon", "icon premium", "icons premium";
```
jagtalon: what about "premium icons"?

lib/DDG/Spice/Icon.pm
```diff
@@ -0,0 +1,47 @@
...
+category "reference";
+topics "web_design";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Icon.pm";
+attribution github => ["iambibhas", "Bibhas D"],
+            twitter => "bibhasdn";
+
+spice to => 'https://api.iconfinder.com/v2/icons/search?query=$1&style=&vector=all&premium=$2';
+spice wrap_jsonp_callback => 1;
+
+spice from => '(.+)/(.+)';
```
jagtalon: Can you change it to `(.+?)/(.+)` so that it's non greedy?

lib/DDG/Spice/Icon.pm
iambibhas: Right. I really wish I could use regex trigger. This icon/icons thing is much redundant.

share/spice/icon/icon.js
```diff
@@ -0,0 +1,90 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_icon = function(api_result){
+
+        // Validate the response (customize for your Spice)
+        if (!api_result || api_result.error || api_result.icons.length == 0) {
```
jagtalon: Check if `api_result.icons` first before accessing the length property?

share/spice/icon/icon.js
```diff
@@ -0,0 +1,90 @@
...
+            name: "Icon",
+            data: api_result.icons,
+            meta: {
+                itemType: 'Icons',
+                searchTerm: decodedQuery,
+                sourceName: "iconfinder.com",
+                sourceUrl: 'https://www.iconfinder.com/search/?q=' + decodedQuery
+            },
+            normalize: function(item) {
+                if (item.raster_sizes.length == 0) {
```
jagtalon: Check `item.raster_sizes` before getting the `length` property

share/spice/icon/icon.js
```diff
@@ -0,0 +1,90 @@
...
+        Spice.add({
+            id: "icon",
+
+            // Customize these properties
+            name: "Icon",
+            data: api_result.icons,
+            meta: {
+                itemType: 'Icons',
+                searchTerm: decodedQuery,
+                sourceName: "iconfinder.com",
```
jagtalon: Change to "Iconfinder" instead of "iconfinder.com". We usually go with the title of the website like in https://duckduckgo.com/?q=ray+bradbury+books&ia=products we say "Amazon" instead of "amazon.com".

share/spice/icon/buy.handlebars
```diff
@@ -0,0 +1,3 @@
+<a href="{{url}}" class="icon__btn btn btn--primary">
+  {{#if is_premium}}Buy{{else}}View{{/if}} at IconFinder
```
jagtalon: Looks like it should be styled as "Iconfinder" instead of "IconFinder"?

share/spice/icon/icon.js
jagtalon: Or maybe go with `total_count`.

jagtalon:
One thing that I'm worried about is that "icon" can mean other things too like "popular icons" and such. We'll look into it more and see if it needs tweaking.

lib/DDG/Spice/Icon.pm
iambibhas: Can you explain this a little? We're definitely returning both the params from the handler.

iambibhas:
@jagtalon yeah, but I can't think of many of those cases. There are only so many things `icon` can mean.

moollaza:
@iambibhas thanks a lot for making this! I wish you would have double-checked with us first though! We've been speaking to the IconFinder team recently and they were just gearing up to build an IA themselves. We've discussed a design internally, but need to confirm some details.

I'm marking this one on hold for now so we can let the IconFinder team know this IA is already under way. It's likely that this will use the Images view to mimic the design of our current Images IA.

iambibhas:
:disappointed: Sure.

moollaza:
@iambibhas sorry for the delay. IconFinder never got back to us, unfortunately. Let's migrate this to the Image template (like the Gif Spice) and go from there.

moollaza:
@iambibhas I've spoken with the team and the we decided icon search should be integrated into our existing image search. We're going to integrate inconfinder internally as an additional image source so that icon searches work through the static images tab. This way we won't have a separate tab for what should really be "Images" results.

PR #1984, opened 2015-07-02
Crunchbase organization info view

rubinovitz:
# Spice Pull Request Template

Using this template will help us better understand your Instant Answer and assist you when necessary.  Simply copy and paste the markdown below into the description of your GitHub pull request and complete as appropriate.

Check if a crunchbase organization exists for the query, and then renders it. Organizations is the only endpoint that can be searched like this, if we'd like to do the other Crunchbase collections (products, people), we'd probably have to mine them ourselves in advance.

Query Crunchbase

It will point you to the most relevant crunchbase organization to your query, saving time/clicks.

The [Crunchbase developer API](https://developer.crunchbase.com/)

It is a site I find myself querying very often.

I do not think there are any crowdsourced startup data sources that are as
comprehensive.

* crunchbase duckduckgo
* crunchbase uber

(the trigger is "crunchbase")

Anyone interested in learning more about startups.

No. There are some [Crunchbase related ideas](https://duck.co/ideas/search?q=crunchbase), but I wanted to start out simple with just querying by organization.

None, I think.

If you look at the screenshot you'll see the apostrophe in the DuckDuckGo is not rendered correctly. In the past I've handled encoding issues on the backend, but do not know what you'd prefer I'd do with this one.

I've been speaking with @moollaza and @nilnilnil about how best to render the results and how to make multiple API calls. I think I have the multiple API call thing worked out, but definitely would like to use item.set for rendering the Spice when its available, and open to design help.

Yes, DDG will need to apply for an API key [here](http://data.crunchbase.com/v3/page/accessing-the-dataset). It is referenced in the code under the environment variable DDG_SPICE_CRUNCHBASE_APIKEY.

Probably at a hackathon years ago.

screenshot 2015-07-01 17 51 09
screenshot 2015-07-01 17 51 37

## Checklist

```
[] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)

    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 7
        [] IE 8

    - Mac OSX
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [X] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [X] Google Chrome
        [] Opera
```

-----
IA Page: https://duck.co/ia/view/crunchbase_collections


LabLayers:
This is awesome! :+1:

altern8tif:
:+1: too...

But maybe the `title` should be changed to something shorter ie. Crunchbase?  "Crunchbase Organization" seems a bit long.  Just my 2 cents' worth.

jagtalon:
@rubinovitz Sweet! Checking it out!

moollaza:
@rubinovitz thanks a lot for contributing. We'll be sure to review this shortly and give you some feedback!

>But maybe the title should be changed to something shorter ie. Crunchbase? "Crunchbase Organization" seems a bit long. Just my 2 cents' worth.

@altern8tif you're correct -- we're trying to avoid using brand names for the tabs, so we'll need to update that. `Answer` might be the best tab name for this IA.

lib/DDG/Spice/Crunchbase/Collections.pm
```diff
@@ -0,0 +1,31 @@
...
+category "reference";
+topics "economy_and_finance";
+
+# Attribution 
+attribution github => ["https://github.com/rubinovitz", "rubinovitz"],
+            twitter => ["https://twitter.com/rubinovitz", "rubinovitz"];
+
+# Triggers
+triggers startend => 'crunchbase';
+spice to => 'http://api.crunchbase.com/v/2/organizations?user_key={{ENV{DDG_SPICE_CRUNCHBASE_APIKEY}}}&query=$1';
```
jagtalon: I don't have a key yet, but it looks like http://api.crunchbase.com redirects to HTTPS? If it does, please update the `spice to` to "https://api.crunchbase.com/..." to prevent any `301` responses (since we don't handle redirects when we deploy)

lib/DDG/Spice/Crunchbase/Organizations.pm
```diff
@@ -0,0 +1,21 @@
...
+# Start at https://duck.co/duckduckhack/spice_overview if you are new
+# to instant answer development
+
+use DDG::Spice;
+
+use strict;
+
+# Triggers
+triggers startend => '///***never trigger***///';
+spice to => 'http://api.crunchbase.com/v/2/organization/$1?user_key={{ENV{DDG_SPICE_CRUNCHBASE_APIKEY}}}';
```
jagtalon: Same as in https://github.com/duckduckgo/zeroclickinfo-spice/pull/1984/files#r33839657

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,42 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_crunchbase_collections = function(api_result){
+        // Validate the response
+        if (!api_result || api_result.error || api_result.data.items.length === 0) {
```
jagtalon: Check if `api_result.data` and `api_result.data.items` exists before accessing it's length property just in case they don't exist.

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,42 @@
...
+        var metadata = api_result.metadata, // get metadata for base urls
+            data = api_result.data, // get org data
+            items = data.items, // get all orgs
+            wwwPathPrefix = metadata.www_path_prefix,
+            imagePathPrefix = metadata.image_path_prefix,
+            firstResult = api_result.data.items[0], // get first org to render for now
+            permalink = firstResult.path; // get endpoint of first org
+
+        // get organization information from organization endpoint
+	$.getJSON('js/spice/crunchbase/organizations/'+ permalink.split('/')[1],function(response) {
```
jagtalon: I think we should check if there is a response like the `if(!api_result ...` above. Call `Spice.failed('crunchbase');` as well when there is no response data.

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,42 @@
...
+            // Render the response
+            Spice.add({
+                id: "crunchbase",
+		name: "Crunchbase Organization",
+		data: response.data,
+		meta: {
+                    sourceName: "Crunchbase",
+                    sourceUrl: api_result.metadata.www_path_prefix + permalink
```
jagtalon: A bit similar to https://github.com/duckduckgo/zeroclickinfo-spice/pull/1984/files#r33839830: Check if `api_result.metadata` exists before accessing `www_path_prefix`.

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,42 @@
...
+                },
+                normalize: function(item) {
+                    return  { 
+                        description: DDG.strip_html(item.properties.description),
```
jagtalon: Check `item.properties` before accessing `item.properties.description`. Also is it possible for description to contain just `""` or maybe just whitespace `"    "`? We might want to not display the Spice at all if it's missing stuff.

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,42 @@
...
+                        url: wwwPathPrefix + permalink,
+                        image: imagePathPrefix + item.relationships.primary_image.items[0].path,
```
jagtalon: Check if `item.relationships.primary_image.items` has at least one element before getting `items[0]`

jagtalon:
@rubinovitz An easy way to check if, say, `api_result.thing.another_thing.something` exists is to use the `DDG.getProperty` function. You can find the documentation in https://duck.co/duckduckhack/spice_js_api#getpropertyobj-pathname. Basic usage:

```javascript
> DDG.getProperty({test: {hello: "something"}}, 'test.hello');
"something"
```

rubinovitz:
@jagtalon Thanks for the review and for the DDG.getProperty function. I got caught up in trying to get this out there to make sure I was doing the 2 API calls correctly. Will implement your and @moollaza 's changes shortly.

rubinovitz:
@moollaza changed title to "Answer" in [change crunchbase AI title to 'Answer' in crunchbase_collections.js 	d7d8c1c](https://github.com/rubinovitz/zeroclickinfo-spice/commit/d7d8c1ce7a71b2712079864f6325f871b63617dd)

@jagtalon  Changed endpoints to https in [8dac0a9](https://github.com/rubinovitz/zeroclickinfo-spice/commit/8dac0a9f23b6eb273008e494b99b9c11c43c29de)
Checked properties before accessing them in [aeef5e9](https://github.com/rubinovitz/zeroclickinfo-spice/commit/aeef5e94f4d53b3f63f38762832d580cfc6a025d) and [48dd27d](https://github.com/rubinovitz/zeroclickinfo-spice/commit/48dd27d6aebe37fb0d3f25c54ba9f10726070f9b)

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,66 @@
...
+            var wwwPathPrefix = metadata.www_path_prefix,
+                imagePathPrefix = metadata.image_path_prefix;
+	}
+
+        // get organization information from organization endpoint
+	$.getJSON('js/spice/crunchbase/organizations/'+ permalink.split('/')[1],function(response) {
+            // return error if no response to organization query
+            if (!response) {
+                return Spice.failed('crunchbase');
+            } else {
```
moollaza: I think we can ditch these `else` blocks. It should help keep the code from being so heavily nested

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,66 @@
...
+
+	var metadata = api_result.metadata; // get metadata for base urls
+	// get base urls if metdata exists
+	if (metadata){
+            var wwwPathPrefix = metadata.www_path_prefix,
+                imagePathPrefix = metadata.image_path_prefix;
+	}
+
+        // get organization information from organization endpoint
+	$.getJSON('js/spice/crunchbase/organizations/'+ permalink.split('/')[1],function(response) {
```
moollaza: `item.set()` is now available and can be used here. Amazon.js contains a live example.

My only concern is that (it seems like) we get so little information from the API initially and I'm wondering if we have enough after the first API call to decide whether the result is relevant.

I was initially thinking this would be a tile view that shows multiple IA's but it makes sense to display a single result when we have an exact match.

How does the API behave when you don't query an exact company name? e.g "crunchbase pets"

If it's returning a bunch of irrellevant results, we're probably going to want to add a `relevancy: {}` block to the `Spice.add()` call so we can ensure we don't show anything totally irrellevant.

moollaza:
@abeyang I'm wondering how we should handle the results for this IA. There are likely to be searches that contain and exact result (e.g. "crunchbase spotify", but I can imagine categorical style searches like "crunchbase music")

Should we just always show a tile view for all the results returned?

share/spice/crunchbase/collections/crunchbase_collections.js
rubinovitz: @moollaza Looking at the [docs](https://developer.crunchbase.com/docs) for the organizations endpoint again, these are the the paramaters it makes sense to search by. We're currently using 'query', which is why I started out using the tile view until you advised not to over email. It's imprecise. (Also, I took a screenshot of all available orgs parameters and put it at the bottom of this comment for reference):

screenshot 2015-07-06 14 35 26
We could do something like:
'Crunchbase name $COMPANYNAME' -> to hit the name option, info view
'Crunchbase $ORGANIZATIONTEXT' -> hit the query option, tile view

and then there's a category parameter as well that requires us to get the available categories and their UUIDs from the categories endpoint beforehand:
screenshot 2015-07-06 14 37 31



Right now the paging data for the categories endpoint is:
```
  "paging": {
   "total_items": 864,
   "number_of_pages": 1,
   "current_page": 1,
   "sort_order": "name ASC",
   "items_per_page": 5000,
   "revision": 1436207068
  }
```

I could add categories, but it would take substantially more time than using the query parameter.

And to save time, here's all the parameters for the organizations endpoint for reference as a screenshot:
screenshot 2015-07-06 14 44 51

moollaza:
>We're currently using 'query', which is why I started out using the tile view until you advised not to over email. 

Sorry for the confusion! By suggesting the Info group I didn't mean for you to abandon the tile view. You can still use the Info group and have tiles. It just means that when you click the tile the detail pane would be different. It looked like you were using the `Text` group in your screenshot?

>We could do something like:
'Crunchbase name $COMPANYNAME' -> to hit the name option, info view
'Crunchbase $ORGANIZATIONTEXT' or 'Crunchbase query $ORGANIZATIONTEXT' -> hit the query option, tile view

This is a good idea, but sadly most users likely won't end up using that syntax unless they know about it beforehand. Looking at the autocomplete results for "crunchbase" it's pretty cleary that user's aren't added any qualifiers to their CB queries:

yolo__at_duckduckgo

I don't have an API key yet so I can't test their API right now, but it sounds to me like the Name endpoint would be ideal to use, if it gives us multiple results (otherwise the Query endpoint could work too). Either way, if it's returning irellevant results, we can use a relevancy block to ensure the **name** of each company is relevant to the query.

Can we modify the API call to only return a certain number of results (i.e. reduce the result per page)? If so let's drop it down to about 20 or 30 (something saner than 5000!)

rubinovitz:
@moollaza You cannot limit the amount of results Crunchbase sends back. 
I'll work on implementing the new query and display. I think I'll need to either cut the results rendered in the JS or Perl.

rubinovitz:
@moollaza item.set is coming up as undefined in normalize (whereas it is defined on onItemShown, but does not update the tile).

rubinovitz:
My key ended up getting upgraded to the Crunchbase v3 API which only returns 10 items per page and returns the key organization details without making a second query. Pushed the changes now. The only thing I'm not happy with is that the sizes of the images and the templates do not always get alone. Open for suggestions on this. 
screenshot 2015-07-12 17 49 57
Bad sizing on the expanded details

moollaza:
> My key ended up getting upgraded to the Crunchbase v3 API which only returns 10 items per page and returns the key organization details without making a second query. 

That's awesome! Glad to hear we have some basic info available to us. I think we should still make the additional API call if we can get even more info for the detail pane, if it's available.

>The only thing I'm not happy with is that the sizes of the images and the templates do not always get alone. Open for suggestions on this. 

@abeyang how should we handle the logo? I assume we should vertically center them with CSS?

> Bad sizing on the expanded details

That seems like a bug -- will investigate!

>I'm also not rendering a tile if it is missing an image, but am wondering if we should stick with that and if so if I should be making sure at least one tile is rendered or else fail the spice and if there's a best practices for that.

Yes l think we should continue to skip results without pictures. We could possibly fallback to a placeholder if we want to, but we don't typically do that.

The best way to enforce this requirement is with a `relevancy`block: https://duck.co/duckduckhack/display_reference#coderelevancycode-emobjectem-optional

In the case that no results have pictures it should automatically fail when we see that the array passed to `data` is empty.

moollaza:
@rubinovitz sorry for the delay! I have an API key now and will be able to put this on a dev machine for further testing.

moollaza:
@rubinovitz @abeyang PR in it's current state is up here: https://moollaza.duckduckgo.com/?q=crunchbase+uber&ia=answer&iai=0

abeyang:
@rubinovitz does the api provide more standardized logos (such as square-sized)?

rubinovitz:
@abeyang no, I am returned the width and height of the image if that's helpful, however.

abeyang:
@rubinovitz got it. Can you provide a sample of the returned api?

abeyang:
Also, can we have more verbatim results? Eg, for the results below, "snapchat" should only return the first 2, whereas "snap" may return more:

pasted_image_7_22_15__11_24_am

abeyang:
For the tiles, can we make them look more like [those in Product](https://duckduckgo.com/?q=bike&ia=products)? So:

- Apply the gray gradient over the images (again, as found in Product)
- Ensure images fit within the thumbnail box. They should be centered, both horizontally and vertically.
- Let's make the tiles taller, too; let's use the 3-line snippet (instead of 2-lines).

@rubinovitz would you be able to take a crack at this? After you have an initial working update, we can make additional tweaks as necessary. :smile:

rubinovitz:
@abeyang 
I.  Sample from API /organizations?name=duckduckgo 

https://gist.github.com/rubinovitz/3aa767daa8a83870eca3

sample from api /organizations/<permalink> where permalink = duck-duck-go 
https://gist.github.com/rubinovitz/14a053347e52364828e3

II. Yes, that's doable.

III. Yes, I can style.

Will let you know as I get II, III done.

rubinovitz:
Hi sorry for the delay. Been doing some other DDG stuff. I think what makes the most sense is for me to try to wrap this on Global Quack & Hack Day Aug 29.

moollaza:
:+1:

rubinovitz:
@abeyang I don't think the product example you sent me is in production any longer. Do you have another example?

share/spice/crunchbase/collections/crunchbase_collections.js
```diff
@@ -0,0 +1,66 @@
...
+        
+        var reCompany = /crunchbase (.*)/; // all words proceeding crunchbase
+        var query = DDG.get_query().match(reCompany)[1]; // get org query
+        var verbatimMatches = items.filter(filterVerbatim); // filter by query
+        
+        if (verbatimMatches){ // if matches by query 
+            var items = verbatimMatches; // limit items to those matches
+        }
+
+	    var metadata = api_result.metadata; // get metadata for base urls
```
mintsoft: nit: there's a hard-tab here, @jagtalon will declare war if it's not switched out to 4 spaces!

abeyang:
@rubinovitz Just checked, it's still in [production](https://duckduckgo.com/?q=bike&ia=products). You might need to change your region to `United States`?

image

moollaza:
Closing as abandoned.

PR #1986, opened 2015-07-02
Quandl: quandl.com redirects to www.quandl.com

jagtalon:
Nginx can't handle redirects, so we modify the Spice endpoint.


moollaza:
LGTM :+1: 

/cc @brianrisk

brianrisk:
Hunh.... is the test environment not Nginx?  Regardless:  thank you!

moollaza:
>is the test environment not Nginx?

Nope. DuckPAN doesn't use NGINX (it would be a pain to have to get users to install it locally). We re-create most of the functionality in Perl. It works *pretty* well, but it's not a 100% solution. We have plans to address this though :+1:

PR #1987, opened 2015-07-02
Zaahir/cryptocurrency ie8 support

moollaza:
Better support for IE8:
- removed trailing whitespace
- switched to use moment.js to handle date/time strings
- ~~update CSS for better UI~~ (wasn't necessary once icons were working again)
- fix JS to provide uppercase filenames for png assets

Result:
dashboard

//cc @claytonspinner

### DIFF
--- a/share/spice/cryptocurrency/cryptocurrency.css
+++ b/share/spice/cryptocurrency/cryptocurrency.css
@@ -88,6 +88,10 @@
     display: inline-block;
 }
 
+    .ie8 .zci--cryptocurrency-result {
+        margin: 0.75em 3.5em 0.5em 0em;
+    }
+
 .zci--cryptocurrency-group {
     white-space: nowrap;
     display: inline-block;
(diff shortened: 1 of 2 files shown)


jagtalon:
@moollaza are you seeing this? It looks like the vertical bar is overlapping on the content, and looks like there's a lot of space after "More at ..."

jagtalon:
screen shot 2015-07-02 at 2 38 46 pm

moollaza:
@jagtalon forgot to mention it breaks in duckpan. It seems that we're not adding the `ie8` class to `<html>` which is why I thought it was so badly broken initially.

In production it works correctly.

Regarding the space after the More at, that still occurs in IE8 but I'm not entirely sure why. It's a minor issue though so I'm not too worried about it.

Now that we have the `moreText` this IA could be updated to let the templates create the more at and exchange rate link.

jagtalon:
@moollaza ok cool :+1:

PR #1988, opened 2015-07-03
Finish today in history

moollaza:
Continuation of #1393 

Now that Sports has gone live, we now have an easier method of implementing expanding tiles. This PR cleans up the CSS and JS for TodayInHistory and updates it to use some new templates features and the new expanding_item template.

/cc @abeyang @killerfish


jagtalon:
@moollaza I like it, and it works well on mobile. The links on the tiles are currently broken though: https://vid.me/GIRQ

lib/DDG/Spice/TodayInHistory.pm
```diff
@@ -3,33 +3,83 @@ package DDG::Spice::TodayInHistory;
...
+description "List of events which occured on this day in history";
+primary_example_queries "today in history", "this day in history";
+secondary_example_queries "historical events on 27 June";
+category "dates";
+topics "everyday","trivia";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/TodayInHistory.pm";
+attribution github => ["https://github.com/killerfish", "Usman Raza"],
+            twitter => ["https://twitter.com/f1shie", "Usman Raza"];
+
+spice to => 'http://en.wikipedia.org/w/api.php?action=query&titles=$1&prop=revisions&rvlimit=1&rvprop=content&format=json&continue=&redirects&callback={{callback}}';
```
jagtalon: Wikipedia now redirects all HTTP to HTTPS. We can't have any redirects in the Spice endpoint. Please change "http://" to "https://"

lib/DDG/Spice/TodayInHistory.pm
```diff
@@ -3,33 +3,83 @@ package DDG::Spice::TodayInHistory;
...
+sub useToday {
+    my $t = DateTime->now;
+    $t->set_time_zone($loc->time_zone);
+    return $t->mday."_".$t->month_name;
+}
 
 handle remainder => sub {
-    return '' if $_ eq '';
+  # check current date in users timezone
+    if (my $date = readParams($_)) {
```
jagtalon: Should probably fail if a day is not given (or use the first day of the month)

screen shot 2015-07-03 at 1 34 42 am

jagtalon: Having more than one space between the day and the month fails as well (it shouldn't):

screen shot 2015-07-03 at 1 36 51 am

jagtalon: This should fail as well screen shot 2015-07-03 at 1 39 26 am

lib/DDG/Spice/TodayInHistory.pm
```diff
@@ -3,33 +3,83 @@ package DDG::Spice::TodayInHistory;
...
+        return $value if (($value = validateParams($parameter2, $parameter1)) ne 'fail');
+    }
+    # no parameter - use today date
+    else {
+        return useToday();
+    }
+}
+
+# validate parameters and return day_month on success
+sub validateParams {
```
jagtalon: Can we use https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/GoodieRole/Dates.pm here instead? But only change if it will solve issues like https://github.com/duckduckgo/zeroclickinfo-spice/pull/1988/files?diff=split#r33841410 for us and none of this is problematic (like we'd probably want support for "jul 10" as well). If it's not any better it's best to come back to it

t/TodayInHistory.t
```diff
@@ -2,22 +2,37 @@
...
+my $day = $t->mday;
+my $month = $t->month_name;
+my $query = $day."_".$month;
+
+my $req  = DDG::Request->new( 
+    query_raw => "today in history", 
+    location => $loc 
+);
 
 ddg_spice_test(
```
jagtalon: Add some failing examples here like "historical events on august"

share/spice/today_in_history/today_in_history.js
```diff
@@ -1,33 +1,141 @@
...
+            source = $(script).attr('src'),
+            ourquery = source.match(/today_in_history\/([^\/]+)/)[1];
+            ourquery = ourquery.replace("_", " ");
 
-    if (!item) {
-        return Spice.failed('today_in_history');
-    }
+        // Extract data
+        var item;
+        for (var key in api_response.query.pages) {
```
jagtalon: Check `api_response.query` before accessing the `pages` property? It might be unlikely that the API would change, but if it does, this could break things.

lib/DDG/Spice/TodayInHistory.pm
moollaza: :+1:

jagtalon:
@moollaza good to go?

share/spice/today_in_history/today_in_history.js
moollaza: Fixed

moollaza:
Yup :+1:

jagtalon:
Looks good :+1:

Issue #1927, opened 2015-06-15
Seasons: Wrong dates for Southern Hemisphere!


moollaza:
A user pointed out that our results for Australia (and the rest of the Southern Hemisphere are incorrect!)

https://twitter.com/nauzilus/status/609086192814923777

We incorrectly assumed that "summer" implies June, but the seasons depend on the hemisphere of the country. We need to be able to determine the hemisphere of the country we're working with in order to choose the correct solstice (June vs December).

/cc @mayo 

---

IA Page: https://duck.co/ia/view/seasons

moollaza:
An interim fix for this would be to make a 2nd API call to Time&Date for the given country in order to get the Lat/Lon of it. Using that, we can determine the hemisphere and choose the right date.

mayo:
@moollaza I've been pondering about this, but didn't have a chance to dig in. I think the DDG location API already gives lon/lat, so it shouldn't need a second request. We already use that to get the country name if none is specified.

I'll try to get a patch tonight.

mayo:
Pull request https://github.com/duckduckgo/zeroclickinfo-spice/pull/1989

moollaza:
IA is no longer live

PR #1989, opened 2015-07-03
#1927 Incorrect dates shown in southern hemisphere

mayo:
Incorrect dates were shown to users in southern hemisphere.

Timeanddate API always returns results in array indexed 0-3 (spring, summer, fall, winter as viewed from northern hemisphere). This means that in southern hemisphere, one has to look at index 3 ("winter") to get the summer dates down south.

### DIFF
--- a/lib/DDG/Spice/Seasons.pm
+++ b/lib/DDG/Spice/Seasons.pm
@@ -9,8 +9,8 @@ my $seasons_qr = join "|", @seasons;
 
 triggers any => @seasons;
 
-spice from => "(.+)/(.+)/.*";
-spice to => 'https://api.xmltime.com/holidays?accesskey={{ENV{DDG_SPICE_TIME_AND_DATE_ACCESSKEY}}}&secretkey={{ENV{DDG_SPICE_TIME_AND_DATE_SECRETKEY}}}&callback={{callback}}&country=$2&year=$1&types=seasons&version=1';
+spice from => "(.+)/(.+)/.+/.*";
+spice to => 'https://api.xmltime.com/holidays?accesskey={{ENV{DDG_SPICE_TIME_AND_DATE_ACCESSKEY}}}&secretkey={{ENV{DDG_SPICE_TIME_AND_DATE_SECRETKEY}}}&callback={{callback}}&country=$2&year=$1&types=seasons';
 
 spice proxy_cache_valid => "200 30d";
 spice is_cached => 1;
@@ -48,6 +48,10 @@ handle query_lc => sub {
         $country = lc $loc->country_code;
     }
 
+    # Get hemisphere
+    my $hemisphere = ($loc->latitude > 0) ? 'north' : 'south';
+
+
     # Detect year
     my $current_year = (localtime(time))[5] + 1900;
     my $year = $+{year} // $current_year;
@@ -66,7 +70,7 @@ handle query_lc => sub {
     }
 
     # Season is not required for the API call, but used in the frontend
-    return $year, $country, $season, {is_cached => $caching};
+    return $year, $country, $season, $hemisphere, {is_cached => $caching};
 
 };
 
--- a/share/spice/seasons/seasons.js
+++ b/share/spice/seasons/seasons.js
@@ -12,8 +12,17 @@
         // Grab the matching search terms and figure out the season
         var script = $('[src*="/js/spice/seasons/"]')[0];
         var source = $(script).attr("src");
-        var query = source.match(/seasons\/[0-9]{4}\/[a-zA-Z]{2}\/([a-z]+)/)[1];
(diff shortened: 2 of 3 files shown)


lib/DDG/Spice/Seasons.pm
```diff
@@ -60,6 +60,10 @@ handle query_lc => sub {
         $country = lc $loc->country_code;
     }
 
+    # Get hemisphere
+    my $hemisphere = ($loc->latitude > 0) ? 'north' : 'south';
```
jagtalon: Wouldn't this get the latitude of the person searching? This would work if I typed in "first day of spring" in Australia, but typing in "summer in australia" fails

screen shot 2015-07-03 at 1 58 53 am

Maybe there's a Perl module that accepts a country and spits out the hemisphere its in?

Another solution would be to ignore countries for now, i.e., accept things like "first day of spring" but don't trigger on "summer in australia". This is not ideal, of course, but it might be good enough. Would still investigate if the above is possible, though.

mayo: doh! You're right. I wouldn't want to remove the location triggering since it's already there. I'll poke at it some more.

jagtalon: @mayo no problem! ping us if you run into anything.

moollaza:
@mayo I spoke to the TimeAndDate Team and they suggest we move to use a specific city when providing the hemisphere. The reason being, not all countries are in a single hemisphere, so if we go with, say the capital city we will always know for sure what the hemisphere is. We can also make a 2nd API call to our existing Time Spice endpoint `/js/time/` with the city & country to get the latitude which will give us the hemisphere.

How does that sound?

mayo:
Sounds reasonable, but that also means we need to look up capital cities somewhere, as from what I know the DDG location api gives the city at user's location. (There's probably a module though) I'll try to get something going this weekend.

mayo:
@moollaza There is also Caida:Countries.pm (not sure how sizeable chunk of caida we'd need to install) which gives lat/lon of countries' geographic centre, but that may not be what we want based on the comments from Time and Date.

jagtalon:
@moollaza does it have to be the capital or is the city and country good enough?

Maybe calling `/js/time/` is not a good idea because it's going to want to call `ddg_spice_time` since `ddg_spice_time(...)` is what's returned by the API (because of the callback parameter). So we'd have to do something like:

```js
env.ddg_spice_seasons = function() { 
    ...
    $.getScript(...);
};

env.ddg_spice_time = function() { ... };
```

Which is weird and could clash if `time.js` was ever loaded along with this. I could be wrong though.

moollaza:
>Maybe calling /js/time/ is not a good idea because it's going to want to call ddg_spice_time since ddg_spice_time(...) is what's returned by the API (because of the callback parameter).

@jagtalon right. We can make another perl module then, that has the same `spice to` url as `DDG::Spice::Time` except it should have no triggers, and no callback defined, like https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Xkcd/Latest.pm

jagtalon:
@moollaza sweet. @mayo please ask if you have any questions!

moollaza:
@mayo @MrChrisW ping! I had forgotten about this one. This issue still exists yet? We're showing incorrect answers for users in the Sourthern hemisphere?

MrChrisW:
@moollaza Unable to reproduce, is this IA disabled? https://ddh10.duckduckgo.com/js/spice/seasons/2015/us/spring

moollaza:
@MrChrisW yup, I forgot that I disabled it when we discovered this bug -- I can deploy the PR to beta

jagtalon:
Hey, @moollaza! please deploy to beta :)

jdorweiler:
@mayo can you update your branch.  There are some conflicts so I can't deploy it to beta.

mayo:
@jdorweiler Updated, but I think the problem from above still persists. This IA uses user's location to determine hemisphere, not taking into account the country named in the search. I don't think the above discussion on how to look up a country/hemisphere has ever been really resolved.

jdorweiler:
Sounds like @moollaza has a potential solution in a comment above.  @moollaza do you think this approach will still work?

moollaza:
Just wanted to chime in here.

The PR is currently deploying to Beta. A lot has changed and creating the additional API endpoint no longer requires a new Perl Module, we just need to create a new `alt_to` (XKCD can be used as an example) that points to the same endpoint as the Time Spice, or we need to create a JSON file that defines all the capitals of all the countries and load that via AJAX to determine the hemisphere

moollaza:
This PR is being closed due to inactivity. If you're still interested in finishing, please let me know and I'll gladly re-open it for you!

PR #1994, opened 2015-07-04
Gravatar - Check for null profiles

altern8tif:
Continuation of #1925 

@moollaza Fixed.  Hope this works now...

screen shot 2015-07-04 at 6 54 44 pm
screen shot 2015-07-04 at 6 54 58 pm
screen shot 2015-07-04 at 6 55 04 pm


moollaza:
LGTM :+1:

PR #1998, opened 2015-07-05
Removed old meta text attribute

### LINKED ISSUE #1794: Update old meta text attributes
We have a shim internally to convert old attribute names to the new ones, but it would be nice if we could just migrate to using the new ones so we could deprecate the shim:

`meta.heading` => `meta.primaryText`
`meta.altMeta` => `meta.secondaryText`

Forecast is good example with both:
https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/forecast/forecast.js#L274

gokul1794:
Removed some old meta text attributes from Transit. Fixes #1794

### DIFF
--- a/share/spice/transit/njt/transit_njt.js
+++ b/share/spice/transit/njt/transit_njt.js
@@ -13,6 +13,7 @@
             signal: 'high',
             data: api_result.routes,
             meta: {
+                primaryText: api_result.origin + " to " + api_result.destination,
                 sourceUrl: api_result.url,
                 sourceName: "NJ Transit"
             },
--- a/share/spice/transit/njt/transit_njt.js
+++ b/share/spice/transit/njt/transit_njt.js
@@ -13,7 +13,6 @@
             signal: 'high',
             data: api_result.routes,
             meta: {
-                heading: api_result.origin + " to " + api_result.destination,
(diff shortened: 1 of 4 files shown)


MariagraziaAlastra:
Hi @gokul1794, thanks for taking care of this!

share/spice/transit/njt/transit_njt.js
```diff
@@ -13,7 +13,6 @@
             signal: 'high',
             data: api_result.routes,
             meta: {
-                heading: api_result.origin + " to " + api_result.destination,
```
MariagraziaAlastra: @gokul1794 this line shouldn't be totally removed, you just need to replace <pre>heading</pre> with <pre>primaryText</pre>, like it's done [here](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1826/files#diff-f43444f36ddec5dfc14010d40ef963e1R263). Thanks!

gokul1794:
@MariagraziaAlastra Sorry about that!

MariagraziaAlastra:
@gokul1794 No problem at all, I see how the issue title might be confusing :)

moollaza:
LGTM :+1:

moollaza:
Thanks @gokul1794 !

PR #2014, opened 2015-07-08
codemodedev's GetGPSCoordinates Spice

codemodedev:
Displays the GPS coordinates for any city, and a link to those coordinates in OpenStreetMap.

It's quicker and easier than wading through search results that may or may not actually show the specific coordinates for the requested city.

Wikipedia

It appears to have GPS coordinates for every major city and many small towns as well, with a fast open API

Possibly, but none that I found had an API or at least not one as fast as Wikipedia's.

gps for denver, chicago gps coordinates, gps coordinates for paoli pa

cartographers, people who work with weather stations, data geeks

no

I don't see any (at least in the Spice tree?) that mention either GPS or coordinates

When running the tests, some of my negative tests are failing.  For example, "gps for denver" is valid for my IA, but "gps for sale" is not.  This works properly in the browser (the IA is simply not displayed), but in the test it fails, because my code intentionally returns Spice.failed() from the JS in cases where the API returns no GPS coordinates, but the command-line test can't test for such JS returns (according to the DDG guys in the Gitter chat).
The test output is:

  Failed test 'Checking for not matching on gps for sale'
  at /home/codio/.../Test/Block.pm line 62.
  got: 'DDG::ZeroClickInfo::Spice=HASH(0x365ccd8)'
  expected: undef

For now I've just commented out those types of negative tests (other negative tests work properly, such as 'gps coordinate conversion' => undef).  I guess that's fine, but should there be a way to test for them that uses the JS output in the test?

No API key or costs are involved

Just browsing around the DDG site and the about-us page

## Checklist

```

    - Linux (LinuxMint 17)
        [x] Google Chrome
        [x] Firefox

    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [x] Google Chrome
        [x] Firefox
        [x] Opera
        [] IE 8
        [x] IE 9
        [x] IE 10
        [x] IE 11

    - Windows XP
        [x] IE 8

    - Mac OSX
        [x] Google Chrome
        [x] Firefox
        [x] Opera
        [x] Safari

    - iOS (iPhone)
        [x] Safari Mobile
        [x] Google Chrome
        [x] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [] Google Chrome
        [] Opera
```
screenshot

### DIFF
--- a/lib/DDG/Spice/GetGPSCoordinates.pm
+++ b/lib/DDG/Spice/GetGPSCoordinates.pm
@@ -17,7 +17,8 @@ category "geography";
 # Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
 topics "geography";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GetGPSCoordinates.pm";
-attribution github => ["codemodedev", "Anthony DiSante"];
+attribution github => ["codemodedev", "Anthony DiSante"],
+            email => ['codes@anthonydisante.com', "Anthony DiSante"];
 
 spice to =>'https://en.wikipedia.org/w/api.php?action=query&titles=$1&prop=revisions&rvprop=content&format=json&callback={{callback}}';
 
--- a/lib/DDG/Spice/GetGPSCoordinates.pm
+++ b/lib/DDG/Spice/GetGPSCoordinates.pm
@@ -1,0 +1,120 @@
+package DDG::Spice::GetGPSCoordinates;
+# ABSTRACT: displays the latitude and longitude for a given city, from the city's Wikipedia page.
+
+use DDG::Spice;
+
+spice is_cached => 1; 
+
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
+name "GetGPSCoordinates";
+source "Wikipedia";
+#icon_url "";
+description "Displays the GPS coordinates for the requested city, with a link to OpenStreetMap.";
+primary_example_queries "gps for paoli, pa", "gps coordinates for philadelphia";
+secondary_example_queries "denver gps coordinates";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#category
+category "geography";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
+topics "geography";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/GetGPSCoordinates.pm";
+attribution github => ["codemodedev", "Anthony DiSante"];
+
+spice to =>'https://en.wikipedia.org/w/api.php?action=query&titles=$1&prop=revisions&rvprop=content&format=json&callback={{callback}}';
+
+triggers start => 'gps coordinates for', 'coordinates for', 'gps for';
+triggers startend => 'gps coordinates';
+
+# Handle statement
+handle remainder => sub {
+
+    # optional - regex guard
+    # return unless qr/^\w+/;
+
+    return unless $_;    # Guard against "no answer"
+
+    # Uppercase the first letter of each word, b/c wikipedia wants "Camden, New Jersey"
+    # not "camden, new jersey".
+    s!(^|\s+)(\w)(\S+)!$1\u$2$3!g;
+
+    # For unusual names like "King of Prussia, Pennsylvania" wikipedia will only accept
+    # "of" and not "Of":
+    s! Of ! of !g;
+
+    # Wikipedia also requires a comma in many cases (e.g. "Paoli, Pennsylvania" not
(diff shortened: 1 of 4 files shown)


MrChrisW:
@codemodedev Thanks for your contribution! We'll be sure to review this shortly.

lib/DDG/Spice/GetGPSCoordinates.pm
```diff
@@ -0,0 +1,121 @@
...
+    s!(,|\s+)TN$!$1Tennessee!i;
+    s!(,|\s+)TX$!$1Texas!i;
+    s!(,|\s+)UT$!$1Utah!i;
+    s!(,|\s+)VT$!$1Vermont!i;
+    s!(,|\s+)VI$!$1Virgin Islands!i;
+    s!(,|\s+)VA$!$1Virginia!i;
+    s!(,|\s+)WA$!$1Washington!i;
+    s!(,|\s+)WV$!$1West Virginia!i;
+    s!(,|\s+)WI$!$1Wisconsin!i;
+    s!(,|\s+)WY$!$1Wyoming!i;
```
MrChrisW: @codemodedev You may be able to use the [Locale-SubCountry](https://github.com/cebjyre/Locale-SubCountry) package for these abbreviations. See [SalesTax](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/SalesTax.pm) as an example.

t/GetGPSCoordinates.t
```diff
@@ -0,0 +1,38 @@
...
+        caller => 'DDG::Spice::GetGPSCoordinates'
+    ),
+    'denver gps coordinates' => test_spice(
+        '/js/spice/get_gpscoordinates/Denver',
+        call_type => 'include',
+        caller => 'DDG::Spice::GetGPSCoordinates'
+    ),
+    # Try to include some examples of queries on which it might
+    # appear that your answer will trigger, but does not.
+    #'gps for sale' => undef,
```
MrChrisW: Comments can be removed.

share/spice/get_gpscoordinates/get_gpscoordinates.js
```diff
@@ -0,0 +1,108 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_get_gpscoordinates = function(api_result){
+
+        if (api_result.error) {
```
MrChrisW: Also need to check if we get a vaild response at all eg.  `!api_result | api_result.error`

lib/DDG/Spice/GetGPSCoordinates.pm
```diff
@@ -0,0 +1,121 @@
...
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
+name "GetGPSCoordinates";
+source "Wikipedia";
+#icon_url "";
+description "Displays the GPS coordinates for the requested city, with a link to OpenStreetMap.";
+primary_example_queries "gps for paoli, pa", "gps coordinates for philadelphia";
+secondary_example_queries "denver gps coordinates";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#category
+category "geography";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
```
MrChrisW: Remove comments above, leave only ABSTRACT

MrChrisW:
@codemodedev This looks ok, however it would be preferable to use an API that returns formatted JSON with `lat` and `long` properties, this would avoid doing all the regex searching/parsing of the Wikipedia response. 

A good free (hit limit) API is the GeoNames database: 

Example: http://api.geonames.org/searchJSON?formatted=true&q=Pennsylvania&maxRows=10&lang=es&username=demo&style=full

Docs: http://www.geonames.org/export/web-services.html

moollaza:
@codemodedev thanks for contributing. This is great, but I think I mentioned before that we should probably just have our existing Maps IA handle these queries, and then update the frontend to show the lat/lon for the map pins.

As @MrChrisW said parsing Wiki with regex's really isn't ideal and probably won't have as good coverage/relevancy checking as our maps data.

@bsstoner what are your thoughts on this? Should we implements this is a separate IA (it looks like we could copy Maps.pm?) with diffferent triggers and set it up to show the Lat/Lon, or should this just be an improvement to our existing Maps IA?

moollaza:
@codemodedev we've made some recent changes to our Maps and sources and we now support these kinds of queries and results, but its slightly broken it seems right now and needs some repairing. Would you be interested in adding some triggers. I'm not sure if the open source JS is where the existing bug lies, but you're welcome to take a look. Here's an example of the functionality: 

https://duckduckgo.com/?q=longitude+of+toronto+canada&t=ffab&ia=maps

Note the `NaN` value in the map pin, which is a bug!

PR #2020, opened 2015-07-09
Remove Drinks Spice

moollaza:
API (and domain) have gone offline

- Sent an email to the domain owner (via whois lookup), and will wait for a reply

to @jdorweiler 
/cc @jagtalon


javathunderman:
@moollaza Is there a replacement API available (perhaps one that we're using for the Recipes IA right now)?

PR #2022, opened 2015-07-10
Commented url in 'example.pm'  referencing spice caching documentation updated.

alohaas:
I noticed that the url to the spice docs was out of date. It should reflect the current page anchor now.


moollaza:
LGTM

PR #2026, opened 2015-07-10
New Commodities functionality for Quandl spice

brianrisk:
Queries triggered by "price" and "cost of" + any of the following:

aluminum alloy
aluminum
coal
copper
crude oil
gasoline spot
gold
natural gas
palladium
platinum
silver

-----
IA Page: https://duck.co/ia/view/quandl_commodities


share/spice/quandl/commodities/quandl_commodities.js
```diff
@@ -0,0 +1,137 @@
...
+        // we need two data points to get percent change
+        if (result.data.length < 2) {
+            return Spice.failed('quandl_commodities');
+        }
+        
+        var quandlCode = result.source_code + "/" + result.code;
+        
+        // setting title and units based on database/dataset string
+        result.header = "Price of ";
+        if (quandlCode === "OFDP/ALUMINIUM_21") {
```
moollaza: May be a little cleaner to use a `switch` here, or better yet, an object that maps the codes to the header and unit so you have:

```js
var commodities = {
    "OFDP/ALUMINIUM_21": {
        header: "Aluminium",
        units: "$USD/tonne",
    },
    ...
}

result.header = "Price of " + commodities[quandleCode].header;
result.units = commodities[quandleCode].units;
```

brianrisk: Love it!  Will do.

share/spice/quandl/commodities/quandl_commodities.js
```diff
@@ -0,0 +1,137 @@
...
+        result.urlTitle = 'View more data at Quandl';
+
+        var recentValue = result.data[0][1];
+        var previousValue = result.data[1][1];
+
+        // month array to make string version of date
+        var months = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"];
+
+        // adding in the previous date point date
+        var fromDate = new Date(result.data[1][0]);
```
moollaza: I think it would be a little more "Spice-esque" take advantage of Moment.js so you don't have to make a `months` array or deal with the other date stuff:

```js
DDG.require('moment.js', function() {
Spice.add({
  ...
  normalize: function(data) {
    return {
      date: moment(data.to_date).format('MMMM Do YYYY')
    }
}
```

brianrisk:
@moollaza your suggested changes have been implemented; those were good suggestions!

share/spice/quandl/commodities/quandl_commodities.js
```diff
@@ -0,0 +1,134 @@
...
+    env.ddg_spice_quandl_commodities = function(api_result){
+
+        if (!api_result) {
+            return Spice.failed('quandl_commodities');
+        }
+
+        var result = api_result;
+
+        // we need two data points to get percent change
+        if (result.data.length < 2) {
```
jagtalon: Check if `result.data` exists before getting the `length` property.

brianrisk: @jagtalon  Done.  Also done for the other Quandl spices.

share/spice/quandl/commodities/quandl_commodities.js
```diff
@@ -0,0 +1,134 @@
...
+        // getting rounded percentage
+        var percentChange = 10000 * ((recentValue - previousValue) / Math.abs(previousValue));
+        percentChange = Math.round(percentChange);
+        percentChange /= 100;
+        result.change_percent = Math.abs(percentChange);
+
+        // setting style based on change
+        if (percentChange > 0) result.changeDirection = 'up';
+        if (percentChange < 0) result.changeDirection = 'down';
+        if (percentChange == 0) result.changeDirection = 'up';
```
jagtalon: Could maybe simplify to this?

```javascript
if (percentChange >= 0) {
    result.changeDirection = 'up';
} else {
    result.changeDirection = 'down';
}
```

brianrisk: @jagtalon there used to be a "no change" option that this addressed.  I kept the code this way in case we opted to return to that.

jagtalon: Ah got it.

lib/DDG/Spice/Quandl/Commodities.pm
```diff
@@ -0,0 +1,65 @@
...
+category "facts";
+attribution web => ["https://www.quandl.com", "Quandl"],
+            twitter => "quandl";
+            
+# hash associating triggers with indicator codes
+my $primary_hash = LoadFile(share('commodities_primary.yml')); 
+
+# triggers sorted by length so more specific is used first
+my @primary_keys = sort { length $b <=> length $a } keys($primary_hash);
+my $primary_qr = join "|", @primary_keys;
```
jagtalon: Looks like `$primary_qr` isn't being used?

share/spice/quandl/commodities/commodities_primary.yml
```diff
@@ -0,0 +1,4 @@
+price: price
+prices: prices
+cost of: cost of
+azzz: azzz
```
jagtalon: what's `azzz`?

brianrisk:
@jagtalon my bad on the 'azzz'.  The Zanran spice was getting triggering queries on "price" there fore I added the 'azzz' trigger for testing; I forgot to take it out.  (BTW Zanran seem broken as it is not returning results).  

RE $primary_qr -- good catch.  I adapted this spice from the Quandl Worldbank spice; this was old code.

moollaza:
@brianrisk I had a chance to put this one on a machine and start testing it. I noticed that some (all?) of these sources aren't providing real time information which I'm a little skeptical about. Do you know if any of these are real time?

I think we'd be better off using a source like Xignite (who we already use for Stocks) to get most of this info as they'll have more accurate numbers. I noticed for example that the natural gas data was over a month old.

A couple other things:
- The result for coal has a misspelled title, I think there's a typo in the API or the data file? It says "Cloal" (https://moollaza.duckduckgo.com/?q=price+of+coal&ia=commodities)
- We actually have prices for Gold, Silver, Platinum and Palladium via XE.com: https://duckduckgo.com/?q=gold&t=ffab&ia=currency -- but the triggering definitely needs improving as "price of gold" doesn't show the result

brianrisk:
@moollaza: Bummer.  However, this makes sense.  For these kinds of prices, users are after up-to-the-moment values.  I didn't know Xignite was an option; I'm fine with scrapping this Quandl implementation in favor of refining the Xignite spice.

moollaza:
>  I'm fine with scrapping this Quandl implementation in favor of refining the Xignite spice.

To clarify, we till need to speak to Xignite about getting access to that data set. Right now we only have access to the Stocks API. I'll make a task to reach out to them, but commodities isn't really a high priority because we haven't seen any demand for this info from our users.

PR #2035, opened 2015-07-14
News Articles Spice

brianrisk:
New IA triggered by names of popular news sites.  The spice fetches the most recent articles from the site's RSS feed and displays as items.  Example triggers:
* engadget
* huffington
* nyt

### DIFF
--- a/lib/DDG/Spice/Articles.pm
+++ b/lib/DDG/Spice/Articles.pm
@@ -8,7 +8,7 @@ use YAML::XS 'LoadFile';
 primary_example_queries "engadget";
 secondary_example_queries "nyt";
 description "Shows latest articles for popular news sites with Atom/RSS feeds";
-name "Articles";
+name "Articles By";
 code_url "https://github.com/brianrisk/zeroclickinfo-spice/lib/DDG/Articles.pm";
 topics "economy_and_finance";
 category "facts";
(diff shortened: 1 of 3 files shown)


brianrisk:
@MrChrisW @moollaza sorry for the branch mixup.

moollaza:
@brianrisk when you mentioned a new IA I thought it was going to be Quandl related, but I should have still asked what the IA was for so I could have made sure it was OK to build before hand. 

I have a couple concerns with this IA:

- We already have our own existing News Spice as I'm sure you've seen, but haven't added too many sources yet. We do have plans to though and I'll check if there's any timeline on that.

- The data source is a Google API. We don't really have any rules on this, but I'd prefer if we could avoid relying on Google services where possible. Right now the only Google API we use is the Google+ API for our Google+ Spice. In this case there are obviously no alternatives, and they're the most official source. In most other cases there are alternatives that should work.

I'll speak with the team about this during our weekly IA meeting tomorrow and get back to your with our thoughts!

brianrisk:
@moollaza I had similar thoughts about the Google API... I wanted to put this IA out there as a proof of concept to get the conversation going.  If this is deemed useful, I can write an endpoint that converts RSS to JSON.   (I didn't want to spend a ton of time on the backend if it turned out moot)

Regarding with how this fits in with News, I think it locks in perfectly.  The News spice should fire with current topics *about* a given query while the Articles will fire when news items are produced *by* a an entity.  Eg "Apple Watch" queries signal a desire to learn about the Apple Watch while "Engadget" queries are performed by users with an underlying interest in articles by Engadget.

moollaza:
Closing -- I believe this is a duplicate of #2273? (which we're handling internally)

PR #2040, opened 2015-07-16
set signal 'high'

### LINKED ISSUE #1834: Currency: Change signal to high for better triggering
The Currency Spice isn't always displaying when it has a tab - we need to set the `signal: 'high'` for this IA so that it works for queries like these:

https://duckduckgo.com/?q=25+USD+in+DKK

------

https://duck.co/ia/view/currency

gokul1794:
fixes #1834


moollaza:
LGTM :+1: @gokul1794 thanks

Issue #2000, opened 2015-07-06
Airlines: flight tracker not showing all flights happening today


whmountains:
Expected behavior: show all four "UA1567" flights happening today:
* DEN -> GEG (left yesterday but arrived today)
* IAH -> LAX
* LAX -> DEN
* DEN -> GEG

Actual Behavior: flight tracker only shows first two flights:
* DEN -> GEG
* IAH -> LAX

Interestingly enough, both Flightstats themselves and also Google show the last three flights but miss the first one.  They probably query the flight database based on departure date rather than arrival but that's a topic for another time.

![duckduckgo_flight_error](https://cloud.githubusercontent.com/assets/8680147/8530287/f4c5962a-23d4-11e5-9498-0c39c65ff472.jpg)
![google_flight_info](https://cloud.githubusercontent.com/assets/8680147/8530294/027afa30-23d5-11e5-978d-4f397d039e80.jpg)

------
IA Page:  http://duck.co/ia/view/airlines

moollaza:
@whmountains thanks for reporting this! It's very helpful :+1: 

@tommytommytommy I thinking we'll should increase the window when checking for flights?

tommytommytommy:
Thanks for the heads up, @whmountains and @moollaza. Right now, we show flights based on arrival times for a window 6 hours in the past and 18 hours ahead. Judging by the times shown above, we should've been able to get all the flights because they all fall within 12 hours of each other. But, we don't explicitly state that we want 24 hours worth of data, so perhaps FlightStats is returning less to us. Let's see if specifying a 24 hour window gets us more results. I'm going to wait for this flight to cycle around again next week and test before submitting a PR.

tommytommytommy:
I took a look at this, and I think there are two complicating factors...the first is it seems that the FlightStats API only returns results for one day (not a 24 hour rolling window starting from the time that we specify--specifying the time and a window only allows you to choose a subset of the current day's results), and the second is that the UTC day and the local time day may be different.

I don't believe that widening the window will get us all the results above with the current API call. We need to figure out another way to get that information, either by making a different API call with a rolling window. I took a look at the docs and didn't find such a function, but worth a second pair of eyes exploring and/or contacting FlightStats. We could also make multiple API calls, but it gets pretty awful (and FlightStats probably would not like this) because we'd have to make three calls to consider night time flights from the previous day,  flights for the current day, and early flights for the next day...

whmountains:
[...]

PR #2047, opened 2015-07-20
Widen flights window

tommytommytommy:
Re: #2000

Stop-gap fix to address the missing flights by making an additional request for either the prior day or the next day depending on the UTC time. I also added some heuristic changes to suppress flights that arrived more than three hours in the past, as well as flights that have scheduled departures more than 24 hours into the future. I would like to merge a fix for this issue first, and then do a second round where we clean up flights by route.

### DIFF
--- a/share/spice/airlines/airlines.js
+++ b/share/spice/airlines/airlines.js
@@ -60,8 +60,8 @@
         var callParameters = decodeURIComponent(source).split('/');
         
         // determine the original date that we made an upstream request
-        var apiRequestDayOfMonth = callParameters[8];
-        var apiRequestHour = callParameters[9];
+        var originalRequestDayOfMonth = callParameters[8];
+        var originalRequestHour = callParameters[9];
 
         // for every API request, we make an additional request to 
         // grab results for late-night arrivals from the night before and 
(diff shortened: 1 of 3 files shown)


tommytommytommy:
screen shot 2015-07-20 at 1 21 02 am 2

tommytommytommy:
To test, I re-enabled the disabled trigger. So...you'll have to do the same to verify :)

moollaza:
@tommytommytommy this LGTM :+1:

share/spice/airlines/airlines.js
```diff
@@ -49,16 +49,64 @@
...
+                                       originalRequestHour, 0, 0));
+
+        // if this response is for the original same-day request...
+        if (originalRequestDayOfMonth == apiResponseDate.getUTCDate()) {
+
+            var apiRequestDate;
+
+            //  and the current hour is before 3AM, get the previous day's flights as well
+            //  (so we can show arrivals for late-night flights)
+            if (originalRequestHour < 3) {
```
moollaza: We may want to hold on to a list of possible fallbacks in case the 2nd API call has nothing useful.

For example "flight EgyptAir 995" we currently show a result on the live site for this but not with these changes because it landed over 3 hours ago. However there's nothing else coming so we should probably still show this.

moollaza: https://duckduckgo.com/?q=flight+EgyptAir+995&ia=airlines

vs https://moollaza.duckduckgo.com/?q=flight+EgyptAir+995

PR #2051, opened 2015-07-22
Ensure a valid $loc and $loc->time_zone exists before creating DateTime object

moollaza:
/cc @jagtalon 

Mostly needed for internal purposes, but it doesn't hurt to be safe :+1:


russellholt:
LGTM, assuming it's tested :)

PR #2052, opened 2015-07-22
Product Hunt: Relevancy Block Continuation

moollaza:
Continuation of https://github.com/duckduckgo/zeroclickinfo-spice/pull/1750

/cc @MrChrisW @altern8tif 

I've moved to a relevancy block because ultimately that is the correct way to handle the relevancy checks. I noticed that when we put it in the `normalize` function, we don't call `Spice.failed()` in the case when there's nothing relevant which is bad as it means all the other IA's that may trigger will end up waiting.

I'm happy with this and will double check that the relevancy checking get's fixed internally.

Any thoughts?

---

IA Page: https://duck.co/ia/view/product_hunt


jagtalon:
@moollaza looks like it's a bit too aggressive:

Current:
screen shot 2015-09-17 at 2 24 46 pm

New:
screen shot 2015-09-17 at 2 25 22 pm

moollaza:
@jagtalon that's because of the internal relevancy bug I discovered earlier: https://github.com/duckduckgo/zeroclickinfo-spice/pull/1750#issuecomment-92426959

I bet if you use "producthunt" instead of "product hunt" you'll have most or all of the results.

Either way I'd rather we have false negatives than false positives. Right now it's a little too lenient on the matching which is worse.

PR #2060, opened 2015-07-31
Twitch: Remove minItemsForModeSwitch

jagtalon:
There doesn't seem to be any reason to show 3 items in a grid so it's probably best to remove this. Removing this will use the default instead which is 12 items.


moollaza:
LGTM :+1:

PR #2065, opened 2015-08-03
First implementation of Citi Bike NYC Bike Sharing IA

marianosimone:
# CitiBike NYC Bike Sharing Stations

Show stations of the Citi Bike Bike Sharing Program (in NY), along with information about bike and docks availability.

This is a way to bring into DDG the functionality of the [official map](https://member.citibikenyc.com/map/), and it's useful for both tourists (to discover stations) and NYC citizens (to know current status)

**Data Source**

The data source is the [official Citi Bike API](https://www.citibikenyc.com/stations/json), and its policy can be found [here](http://www.citibikenyc.com/data-sharing-policy)

```
citibike ny
citi bike nyc
citi bike ny stations
bike sharing brooklyn
```
**What does the Instant Answer look like?**

citibike_brooklyn_at_duckduckgo

## Checklist

```

    - Windows 8
        [X] Google Chrome
        [X] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 8

    - Mac OSX
        [X] Google Chrome
        [] Firefox
        [] Opera
        [] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [] Google Chrome
        [] Opera
```

-----
IA Page: https://duck.co/ia/view/citi_bike_nyc

### DIFF
--- a/lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm
+++ b/lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm
@@ -1,0 +1,31 @@
+package DDG::Spice::BikeSharing::CitiBikeNYC;
+
+use strict;
+use DDG::Spice;
+
+name 'Citi Bike NYC';
+source 'Citi Bike NYC';
+icon_url 'http://www.citibikenyc.com/assets/images/favicon.ico';
+description 'Search Citi Bike stations in New York City';
+primary_example_queries 'citibike nyc', 'citi bike ny';
+secondary_example_queries 'citibike stations near brooklyn';
+category 'geography';
+topics 'everyday', 'travel', 'entertainment', 'geography';
+code_url 'https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm';
+attribution github => 'marianosimone',
+            web  => ['http://www.marianosimone.com',  'Mariano Simone'];
+
+triggers any => 'citibike', 'citi bike', 'bike sharing', 'bike share';
+
+spice to => 'https://www.citibikenyc.com/stations/json';
+spice wrap_jsonp_callback => 1;
+spice is_cached => 0;
+spice proxy_cache_valid => '200 304 15m';
+
+handle remainder => sub {
+    $loc->city.' '.$_ =~ /(ny|new york|nyc|new york city|brooklyn|manhattan|queens)/i;
+    return unless $1;
+    return $1;
+};
+
+1;
--- a/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
+++ b/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
@@ -27,6 +27,23 @@
 
         DDG.require('moment.js', function() {
             DDG.require('maps', function() {
+                moment.locale('en', {
+                    relativeTime : {
+                        past:   "%s ago",
+                        s:    "a sec",
+                        ss:   "%d sec",
+                        m:    "a min",
+                        mm:   "%d min",
+                        h:    "an hour",
+                        hh:   "%d hours",
+                        d:    "a day",
+                        dd:   "%d days",
+                        M:    "a month",
+                        MM:   "%d months",
+                        y:    "a year",
+                        yy:   "%d years"
+                    }
+                });
                 Spice.add({
                     id: 'citi_bike_nyc',
                     name: 'Citi Bike NYC',
--- a/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
+++ b/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
@@ -1,0 +1,78 @@
+(function(env) {
+    "use strict";
+
+    var references = {
+        manhattan: {
+            lat: 40.7590615,
+            lon: -73.969231
+        },
+        brooklyn: {
+            lat: 40.645244,
+            lon: -73.9449975
+        },
+        queens: {
+            lat: 40.651018,
+            lon: -73.87119
+        }
+    };
+
+    env.ddg_spice_bike_sharing_citi_bike_nyc = function(api_result) {
+        if (!api_result || !api_result.stationBeanList) {
+            return Spice.failed('citi_bike_nyc');
+        }
+
+        var script = $('[src*="/js/spice/bike_sharing/citi_bike_nyc/"]')[0],
+            source = $(script).attr("src"),
+            query = source.match(/citi_bike_nyc\/([^\/]+)/)[1];
+
+        DDG.require('moment.js', function() {
+            DDG.require('maps', function() {
+                Spice.add({
+                    id: 'citi_bike_nyc',
+                    name: 'Citi Bike NYC',
+                    meta: {
+                        sourceName: 'Citi Bike NYC',
+                        sourceUrl: 'https://www.citibikenyc.com/stations',
+                        itemType: 'Bike Stations'
+                    },
+                    model: 'Place',
+                    view: 'Places',
+                    templates: {
+                        item: 'base_item',
+                        options: {
+                            content: Spice.bike_sharing_citi_bike_nyc.content
+                        },
+                        variants: {
+                            tile: 'narrow'
+                        }
+                    },
+                    sort_fields: {
+                        distance: function(a, b) {
+                            return a.distanceToReference - b.distanceToReference;
+                        }
+                    },
+                    sort_default: 'distance',
+                    data: api_result.stationBeanList,
+                    normalize: function(item) {
+                        if (item.testStation || item.statusKey != 1) {
+                            return null;
+                        }
+                        if (references[query]) {
+                            var pointOfReference = L.latLng(references[query]);
+                        }
+                        return {
+                            name: item.stationName,
+                            lat: item.latitude,
+                            lon: item.longitude,
+                            distanceToReference: pointOfReference ? L.latLng(item.latitude, item.longitude).distanceTo(pointOfReference) : item.id,
+                            title: item.stationName,
+                            availableDocks: item.availableDocks,
+                            availableBikes: item.availableBikes,
+                            lastCommunication: moment(new Date(item.lastCommunicationTime)).fromNow()
+                        };
+                    }
+                });
+            });
+        });
+    }
+}(this));
--- a/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
+++ b/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
@@ -50,7 +50,9 @@
                     meta: {
                         sourceName: 'Citi Bike NYC',
                         sourceUrl: 'https://www.citibikenyc.com/stations',
-                        itemType: 'Bike Stations'
+                        itemType: 'Bike Stations',
+                        pinIcon: 'ddgsi-circle',
+                        pinIconSelected: 'ddgsi-star'
                     },
                     model: 'Place',
                     view: 'Places',
--- a/t/BikeSharingCitiBikeNYC.t
+++ b/t/BikeSharingCitiBikeNYC.t
@@ -1,0 +1,26 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+use Test::More;
+use DDG::Test::Spice;
+
+ddg_spice_test(
+    [qw( DDG::Spice::BikeSharing::CitiBikeNYC )],
+    'citi bike ny' => test_spice(
+        '/js/spice/bike_sharing/citi_bike_nyc/ny',
+        call_type => 'include',
+        caller => 'DDG::Spice::BikeSharing::CitiBikeNYC',
+    is_cached => 0,
+    ),
+    'citibike stations near brooklyn' => test_spice(
+        '/js/spice/bike_sharing/citi_bike_nyc/brooklyn',
+        call_type => 'include',
+        caller => 'DDG::Spice::BikeSharing::CitiBikeNYC',
+    is_cached => 0,
+    ),
+    'citibike' => undef,
+    'citibike washington' => undef,
+);
+
+done_testing;
--- a/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.css
+++ b/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.css
@@ -1,0 +1,32 @@
+.tile--citi_bike_nyc {
+    text-align: center;
+    height: 12em;
+}
+
+.tile--citi_bike_nyc .zci__header {
+    height: 2.5em;
+}
+
+.tile--citi_bike_nyc .side_by_side {
+    overflow: auto;
+    margin-left: auto;
+    margin-right: auto;
+    width: 100%;
+}
+
+.tile--citi_bike_nyc .side_by_side .left {
+    float: left;
+    border-right: 1px solid #E5E5E5;
+}
+
+.is-mobile .tile--citi_bike_nyc .side_by_side .left {
+    width: 50%;
+}
+
+.tile--citi_bike_nyc .side_by_side .right {
+    float: right;
+}
+
+.is-mobile .tile--citi_bike_nyc .side_by_side .right {
+    width: 50%;
+}
--- a/share/spice/bike_sharing/citi_bike_nyc/content.handlebars
+++ b/share/spice/bike_sharing/citi_bike_nyc/content.handlebars
@@ -1,0 +1,15 @@
+<div class="tile__body">
+    <h1 class="zci__header tx--l7">{{title}}</h1>
+    <div class="tile__content side_by_side">
+        <div class="left half">
+            <p class="tx-clr--dk">Bikes</p><br/><p class='tx-clr--dk t-xxl'>{{availableBikes}}</p>
+        </div>
+
+        <div class="right half">
+            <p class="tx-clr--dk">Docks</p><br/><p class='tx-clr--dk t-xxl'>{{availableDocks}}</p>
+        </div>
+    </div>
+    <div class="tile__foot ">
+        <span class="tx--12 tx-clr--grey">Updated {{lastCommunication}}</span>
+    </div>
+</div>


t/BikeSharingCitiBikeNYC.t
```diff
@@ -0,0 +1,20 @@
...
+use DDG::Test::Spice;
+
+ddg_spice_test(
+    [qw( DDG::Spice::BikeSharing::CitiBikeNYC )],
+    'citi bike ny' => test_spice(
+        '/js/spice/bike_sharing/citi_bike_nyc/ny',
+        call_type => 'include',
+        caller => 'DDG::Spice::BikeSharing::CitiBikeNYC',
+    is_cached => 1,
+    ),
```
iambibhas: Can you add a test case for the secondary query above `'citibike stations near brooklyn'`? Otherwise it looks good.

marianosimone: @iambibhas added the test (and modified the code a bit to actually capture the geographic reference).

Still, I'm not doing anything with that in the js... It would be nice to be able to sort/filter. Any ideas?

iambibhas: The API doesn't seem to have any search feature. The best you can do is probably focus the searched area on the map. Make a list of approximate center coordinated for those 3 or 4 areas and focus them initially. Not sure if that's possible with the maps api. Haven't used it yet.

marianosimone: @iambibhas thanks for the idea... I've just commited a change that takes a reference point in one of the specified places, and calculates the distance to it when sorting the stations

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -0,0 +1,79 @@
...
+
+        DDG.require('moment.js', function() {
+        DDG.require('maps', function() {
+            var script = $('[src*="/js/spice/bike_sharing/citi_bike_nyc/"]')[0];
+            var source = $(script).attr("src");
+            var query = source.match(/citi_bike_nyc\/([^\/]+)/)[1];
+
+            if (query) {
+                var references = {
+                    'brooklyn': L.latLng(40.645244, -73.9449975),
```
marianosimone: this might not be ideal... but I think is a good start :D

zekiel:
sidebar: @marianosimone  the folks at Code 4 Philly are interested in doing a similar implementation via CyclePhilly open data http://www.cyclephilly.org/

marianosimone:
@zekiel I expected that (that's the reason behind the dir structure). It would be awesome to get this code ready to be merged, and work with them on how to share as much code as possible :D

zekiel:
@marianosimone I'll put you in touch!

marianosimone:
@moollaza @iambibhas @zekiel any other feedback?

marianosimone:
@zekiel @moollaza BTW, @AcriCAA has been doing some work on getting a similar implementation for Philly: https://github.com/AcriCAA/zeroclickinfo-spice/tree/05132575baa2aa2a71ea15cea83a8718eab0c6ed

Once one of them is merged, it would be cool to extract common features and see how to extend to new cities!

AcriCAA:
Thanks, Mariano. I intend to submit a pull request on the Philly version
today.

lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm
```diff
@@ -0,0 +1,30 @@
...
+topics 'everyday', 'travel', 'entertainment', 'geography';
+code_url 'https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm';
+attribution github => 'marianosimone',
+            web  => ['http://www.marianosimone.com',  'Mariano Simone'];
+
+triggers any => 'citibike', 'citi bike', 'bike sharing', 'bike share';
+
+spice to => 'https://www.citibikenyc.com/stations/json';
+spice wrap_jsonp_callback => 1;
+spice proxy_cache_valid => '200 304 15m';
```
moollaza: We need to also set `spice is_cached => 0` because you're using the location API for generic queries.

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -0,0 +1,79 @@
...
+                        title: item.stationName,
+                        altSubtitle: subtitle,
+                        footer_content: Spice.bike_sharing_citi_bike_nyc.foot_back,
+                        lastCommunication: moment(new Date(item.lastCommunicationTime)).fromNow()
+                    }
+                };
+            }
+
+            Spice.add({
+                id: 'CitiBikeNYC',
```
moollaza: Let's go with `"citi_bike_nyc"`

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -0,0 +1,79 @@
+(function (env) {
+    "use strict";
+
+    env.ddg_spice_bike_sharing_citi_bike_nyc = function(api_result) {
+        if (!api_result || !api_result.stationBeanList) {
+            return Spice.failed('citibikenyc');
```
moollaza: This need's to match our (updated) `id`, `"citi_bike_nyc"`

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        DDG.require('moment.js', function() {
+        DDG.require('maps', function() {
+            var script = $('[src*="/js/spice/bike_sharing/citi_bike_nyc/"]')[0];
```
moollaza: I think we should move these outside the `require` block

lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm
```diff
@@ -0,0 +1,30 @@
...
+
+handle remainder => sub {
+    $loc->city.' '.$_ =~ /(ny|new york|nyc|new york city|brooklyn|manhattan|queens)/i;
```
moollaza: Why does the user's location matter? Shouldn't I, from Toronto, Canada, be able to look for "citi bike nyc"?

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -0,0 +1,79 @@
...
+            var source = $(script).attr("src");
+            var query = source.match(/citi_bike_nyc\/([^\/]+)/)[1];
+
+            if (query) {
```
moollaza: Is it possible to have no `query`? Should the Perl should prevent that?

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
moollaza: Seems like the return from this is just a standard object with a `Lat: ...` and `Lon: ...` -- do we need to use Leaflet here?

moollaza: Ah I see the prototype has some other stuff methods. But if we're not using them, then I image we can just use a static, normal object

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -0,0 +1,79 @@
...
+                    'queens': L.latLng(40.651018, -73.871192)
+                };
+                var pointOfReference = references[query];
+            }
+
+            function normalize(item) {
+                if (item.testStation || item.statusKey != 1) {
+                    return null;
+                }
+                var subtitle = item.availableBikes + ' bike' + (item.availableBikes == 1 ? '' : 's') + ' | ' + item.availableDocks + ' dock' + (item.availableDocks == 1 ? '' : 's');
```
moollaza: We have some awesome helpers you can use to do this:

```js
var subtitle = [
                    item.availableBikes + DDG.pluralize(item.availableBikes, "bike"),
                    item.availableDocks + DDG.pluralize(item.availableDocks, "dock")
                ];
```

If you pass an array to `subtitle` we'll automatically add the separator for you. `DDG.Pluralize()` will handle pluralization as well :smile:

moollaza:
@marianosimone is there any chance of the API data getting updated to include the components of the address? like city, state, street, etc? It would really help because then we could show the full address on the back of the tile, and allow for a map search link like we usually do for Places tiles.

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
moollaza: We might as well define that inline, below so we don't need the extra variable.

moollaza:
@marianosimone after playing around with this a bit, I think I'd like the JS to look more like this:

https://gist.github.com/moollaza/22e4d2bc78a271b6db78

This way we don't need to make redundant calls to Leaflet to make LatLon objects in the references until we need it. I think this code looks a bit more like our other IA's.

I've also move to a 1-sided tile, because the flipping isn't doing much for us right now. We can also add more content to the tile if needed.

@abeyang any thoughts on the tile design? Unfortunately we don't have a proper address except for the name we're showing in the tile so we can't break it down much like we usually do.

The result looks like this:
citi_bike_queens_at_duckduckgo

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
marianosimone: Ohhh.. good catch. Fixing!

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
marianosimone: :+1:

marianosimone:
@moollaza I applied most of your changes (cache, id renaming, re-organizing the js function, pluralizer, not checking for query 2 times - perl does it already).

Regarding the Address: they DO include it. I'm not sure what changes to do to the way we show that. If you point me in the right direction I can fix that

As per your example screenshot (Queens), the problem in your gist was that the distance calculator was removed... I added it again and now it shows the nearest stations (still, that's Brooklyn, which is not as good as Queens). Luckily, they are pushing their expansion pretty hard: http://expansion.citibikenyc.com

Finally, regarding the location: If you are in Toronto, you can search "citi bike nyc" and we'll show you the IA, but not for "citi bike". If you are in NY, though, just searching for "citi bike" is enough. Makes sense?

jagtalon:
@marianosimone looks like I'm getting an error in the console. Are you seeing the same thing?

screen shot 2015-08-18 at 2 17 48 pm

marianosimone:
@jagtalon sorry about that... you can `pull --rebase` now and get the fix :D

moollaza:
@marianosimone excellent work on the fixes! Thanks a lot :+1: 

> Finally, regarding the location: If you are in Toronto, you can search "citi bike nyc" and we'll show you the IA, but not for "citi bike". If you are in NY, though, just searching for "citi bike" is enough. Makes sense?

That totally makes sense. Thanks for explaining.

moollaza:
Alright! This is now up for testing at https://comserv.duckduckgo.com/?q=citi+bike+queens&ia=citibikenyc&iax=1&iai=437

@abeyang this one is ready for a design review! Any thoughts?

------

>Regarding the Address: they DO include it. I'm not sure what changes to do to the way we show that. If you point me in the right direction I can fix that

@marianosimone looking at the JSON, I just see the intersection's given in the data? The "city", "street", "postal code" etc seem to be blank?

marianosimone:
@moollaza indeed... the address comes in empty. I'll follow up with the people at Citi Bike to see if we can get that on (it would be even better to also have neighborhood or at least borough!)

If I get that data from them... how should I add it?

AcriCAA:
So cool that it is ready for testing already.

@abeyang.  I mentioned in my Philadelphia iteration
<https://github.com/duckduckgo/zeroclickinfo-spice/pull/2102> of this that
the numbered markers are very misleading.  Based on my user research as
well as talking to the folks who run the Philly bikeshare, when users see a
number on a marker in relation to bikeshare, their expectation is that the
number signifies the number of bikes or docks available.

marianosimone:
I think @AcriCAA's point about the pin numbers is a good one... @moollaza is there any way to remove them?

abeyang:
Hi @marianosimone and @AcriCAA -- from your experience, and the people you've talked to, which is more important: the number of bikes available or the number of docks free? In other words, do people (in general) care more about one over the other?

And @marianosimone I've noticed you got a "last checked" time for each tile. But aren't they generally all checked at around the same time (+/- few minutes)?

Your answers will help direct the design for these types of IAs. Thanks!

marianosimone:
@abeyang I don't have data on the pins data (as a matter of fact, [Citi Bike's map](https://member.citibikenyc.com/map/) doesn't show a number in the pin at all): my intuition would be to NOT have a number or have a half/half pin (but that could be confusing)

As per the communication time, it can be different between stations (at least for NYC), because it station reports back to the main server at its own pace, or might be offline at any given time.

abeyang:
@marianosimone well let's not worry about implementation for now -- just wondering if one metric is more important than the other.

AcriCAA:
@abeyang their expectation is actually to see some some of pie chart
indicator or fuel gauge type of effect showing how full or empty the
station is. There are other apps that let you choose between a dock or bike
number (see below) but that is too complicated I think.

We are working with a gradient effect in the Indego implementation for the
CyclePhilly app I helped build (see below).

That being said, like @marianosimone said, no numbers or indicators are
better than what we have now and still useful and effective. So if that's
the simplest solution, I'd say just strip the numbers. .

Here are some examples of gradients:

From CyclePhilly prototype:




From SpotCycle:



From Transit app which is my favorite and perhaps most suitable here if
possible:

abeyang:
Sorry @AcriCAA , I don't see any of your images. 

But I take it that users typically care about both metrics? Again, I'm not talking about what to show / what not to show on the pins, I just want to know what's most important to users.

moollaza:
@marianosimone @AcriCAA you can specify the number for the pin if you pass a `num: ##` property in the `normalize()` function :+1:

AcriCAA:
@abeyang: just the screenshots here: 


Here are some examples of gradients:

From CyclePhilly prototype:
unnamed




From SpotCycle (allows you to choose between bikes/docks at the top and then shows the respective number):

unnamed-1


From Transit app which is my favorite and perhaps most suitable here if
possible:

unnamed-2

AcriCAA:
@abeyang To answer the rest of your question.  My assessment is that users just want a pin they can tap on on-the-fly which then shows the bikes/docks available in some sort of pop-up.  The importance of the bike or dock number depends on whether the user is starting his or her trip or ending his or her trip.  If it is the beginning of the trip, users want to know bike numbers, if it is the end of the trip users what to know docks available.

abeyang:
Thanks, @AcriCAA -- that makes a lot of sense!

abeyang:
FYI @marianosimone , we're working on the design at the moment. Will share with you (and @AcriCAA ) once we're done!

abeyang:
@marianosimone , we've considered a wide variety of options -- mostly having to do with the pins. but for now, let's go with the simple pins -- would love to do one where it would show visually which stations have free bikes/docks, but let's save that for Version 2.

bike sharing - v3d

The tiles here are custom, but they're very similar to the ones we're using for [GoWatchIt](https://duckduckgo.com/?q=stream+jurassic+park&ia=watch). If you need help getting started, @moollaza will be able to help ya. The idea is that the tiles that are shown will be represented by the pins, and everything else will be dots. Selected pin is the dark one with the star.

Hopefully everything else is straight-forward, but if you have any questions, please don't hesitate to holler! /cc @AcriCAA

jagtalon:
@marianosimone thanks for the fixes! Can't wait for this and https://github.com/duckduckgo/zeroclickinfo-spice/pull/2102 to get merged in.

marianosimone:
@abeyang thanks a lot, awesome design! I started working on it, and I think I'm getting there:

citibike_nyc_at_duckduckgo

There are a couple of issues and things I need help with:

1. Some of the addresses are 1, 2 or 3 lines long, and that messes up the alignments.. any ideas on how to fix?
2. The "updated" part sometimes gets clipped. Not a huge bummer, but would be nice to make it fit.
3. How do I avoid the numbers being added to the pins? (I tried sending an empty string as `num` in the normalization as @moollaza suggested, but it didn't work :cry: - BTW sending a number doesn't work either. It would be nice to have documentation on what things are supported there)
4. How do I get a star to be shown in the pin?

A quick & dirty way to fix 1 and 2 would be NOT using the "narrow" variant, but I don't think the cards look good being so wide

marianosimone:
Also, will it be possible to share templates/css between this and other bike sharing IAs? Seems like all this layout woork could be extracted somehow

abeyang:
@marianosimone--

1. For the header, let's keep it at 2-lines and truncate anything that goes beyond that (make sure you're using `.tx--17` class). I think if you have a set height, that would fix the alignment issue...
2. Use "min" instead of "minutes" :smile: 

For 3 and 4, I think @moollaza will have to address that.

And yes, we do hope to turn this into a template, but in the meantime, others can take you existing custom style and apply it to theirs.

marianosimone:
@abeyang thanks! I fixed 1 with your idea.

2. Won't work... we are using [Moment.js' fromNow](http://momentjs.com/docs/#/displaying/fromnow/). I could reduce the font size and padding to make it fit.. but I think it gets too small :(
3 and 4. Agreed, let's wait for @moollaza

moollaza:
@marianosimone @abeyang we don't really have a way yet to get 

>How do I avoid the numbers being added to the pins? (I tried sending an empty string as num in the normalization as @moollaza suggested, but it didn't work :cry: - BTW sending a number doesn't work either.

We actually don't have a way to do this yet, I'll need to speak with the frontend team to have them add support for all this. 

>It would be nice to have documentation on what things are supported there)

We'll document this stuff as soon as it's available :)

>How do I get a star to be shown in the pin?

This also needs to be implemented...

mattr555:
> Won't work... we are using Moment.js' fromNow. I could reduce the font size and padding to make it fit.. but I think it gets too small :(

You could try [customizing the locale](http://momentjs.com/docs/#/customization/relative-time/), but a simple `.replace('minutes', 'min')` should suffice.

marianosimone:
Thanks @mattr555, good find! Applied, and looking good

xuv:
Hey. It's Julien. Just jumping here, following a discussion with @marianosimone during the NYCmeetup. 
In Europe, a big contractor for bike sharing in cities is JCDecaux And here is a link to the APIs for the different cities they operate in. https://developer.jcdecaux.com/#/opendata/vls?page=static
Would sure would like to help push this project for these. Let me know of I can help.

bsstoner:
@marianosimone, the changes to customize the pin icons went live yesterday. You should now be able to pass any of our icon class names into the meta block of the Spice.add() call, i.e:
```
Spice.add({
  ...
  meta: {
    ...
    pinIcon: 'ddgsi-circle',
    pinIconSelected: 'ddgsi-star'
  }
});
```

marianosimone:
@bsstoner that looks awesome (and super easy config! is it already documented?). Thanks for making it happen!

@zekiel @moollaza what do you guys think?

bikeshare

@AcriCAA it would be nice to add the same config to Philly

abeyang:
Looking good, @marianosimone ! Still need some few tweaks though:

First, let's make the dots within the pins a tad smaller:

```css
.mapview-marker__icn.ddgsi-circle {
  font-size: 6px;
}
```

Second, let's give each tile some more breathing room:

```css
.tile--citi_bike_nyc.tile--c--n {
  width: 13em;
  height: 13em;
}
```

Next, for the header `h1`, instead of the class `tx--l7`, let's replace that with `tx--14`.
Let's add the `tx--14` class to "Bikes", "Docks", and the footer; and add `tx--25` to the two numbers under bikes/docks:

pasted_image_9_2_15__12_45_pm

Let's also remove the `<br>` between Bikes and the number (as well as Docks and the number) -- not necessary. Next, let's close the gap between the text and numbers, as well as scooting it down:

```css
p.tx-clr--dk.tx--14 {
  padding-top: 0;
}

p.tx-clr--dk.tx--25 {
  padding-top: 0;
  padding-bottom: 0;
  line-height: 1.1;
}

div.tile__content.side_by_side {
  margin-top: 18px;
}
```

Finally, let's space out the bikes and docks a bit:
```css
.tile--citi_bike_nyc .side_by_side .left {
  padding-right: 7px;
}

.tile--citi_bike_nyc .side_by_side .right {
  padding-left: 7px;
}
```

At the end of it all, your tiles should look like this one:

image

Happy hacking! :smile:

moollaza:
@marianosimone @AcriCAA BTW it's live here: with Brian's suggested updates: https://moollaza.duckduckgo.com/?q=citi+bike+nyc&ia=bikesharing

I'll implement these final touches for you and merged this in :+1:

marianosimone:
Thanks @moollaza and @abeyang, glad to see this finally merged :d
Looks like @xuv could start working on those he found using this PR as a model, right?

xuv:
@marianosimone Glad to hear it's merged. Congrats. Seen on the @moollaza instance and it looks great.
Looking forward to port this for other cities like Brussels and Paris.

moollaza:
@marianosimone @xuv checkout https://github.com/duckduckgo/zeroclickinfo-spice/pull/2151 :)

AcriCAA:
@marianosimone @moollaza Using this in brooklyn.  I am wondering if "bikeshare" (one word) should be added to the query. I am getting this result.  

img_7489

PR #2075, opened 2015-08-06
Bible - decoding html entities, changed template group, fixed source url

### LINKED ISSUE #2055: Bible: HTML encoding in API response
The instant answer for hosea 11:9 shows some mojobake:

I cannot carry out my fierce anger! I cannot totally destroy Ephraim! Because I am God, and not man &#8211; the Holy One among you &#8211; I will not come in wrath! 

------
IA Page:  http://duck.co/ia/view/bible

iambibhas:
Fixes #2055.

screenshot from 2015-08-06 18 49 13


moollaza:
@iambibhas thanks a lot! I was testing with some random values and came across a result the wad "[[EMPTY]]". So it looks like we need to add a check for that:

matthew_23_14_at_duckduckgo

iambibhas:
@moollaza I was thinking about an empty string case. Didn't think it'll appear like this though.`[[EMPTY]]`!!

iambibhas:
@moollaza can I use [`.trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)? It's supported from IE9.

moollaza:
>can I use .trim()? It's supported from IE9.

Should be OK. We're dropping support for IE8. jQuery offers a trim fn though.

iambibhas:
@moollaza this should be good to go.

jagtalon:
Looks good!

jagtalon:
:+1:

PR #2079, opened 2015-08-07
Lastfm - added relevancy block, stop words and tests

### LINKED ISSUE #2050: Last.fm Artist: Add relevancy block to require description for results
Currently the query "apple watch bands" triggers this IA and shows a pretty useless result as the "band" has no picture or description.

- https://duckduckgo.com/?q=apple+watch+bands&ia=music
- https://duckduckgo.com/?q=bands+of+the+80%27s&ia=music

We should add a relevancy block and use it to ensure any potential result has a description.

We may also need to set `/watch bands?/` as a negative trigger by checking the query against that and bailing if it matches.

------
IA Page:  http://duck.co/ia/view/lastfm_artist

iambibhas:
Fixes #2050.

 - Added relevancy block to ensure that artist description is available
 - Added stop words `/watch bands?/`
 - Added tests, there were none.

**Bug Found:** While the spice has code for showing similar artists, it doesn't implement the method to show them. `ddg_spice_lastfm_artist_similar` callback method is called for any `similar artist` queries, but it's not implemented. Hence the query fails with a JS error. 

    Uncaught ReferenceError: ddg_spice_lastfm_artist_similar is not defined

/cc @jagtalon: Am I missing something?

### DIFF
--- a/lib/DDG/Spice/Lastfm/Artist.pm
+++ b/lib/DDG/Spice/Lastfm/Artist.pm
@@ -26,6 +26,10 @@ triggers any => 'similar', 'band', 'bands', 'musician', 'musicians', 'artist', '
 handle query_lc => sub {
     my $synonyms = "bands?|musicians?|players?|artists?|performers?|singers?|rappers?|djs?|vocalists?|songsters?";
 
+    #Ignore Queries like "apple watch bands"
+    if(m{watch\ bands?}) {
+        return;
+    }
     #Queries like "bands similar to incubus" or "artists similar ben folds"
     if(m{(?:$synonyms)\s+similar\s+(?:to\s+)?(\S+(?:\s+\S+)*)}) {
         return $1, 'similar';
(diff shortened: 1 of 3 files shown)


moollaza:
> Bug Found: While the spice has code for showing similar artists, it doesn't implement the method to show them. ddg_spice_lastfm_artist_similar callback method is called for any similar artist queries, but it's not implemented. Hence the query fails with a JS error. 

@iambibhas that's definitely a bug that should be fixed!

jagtalon:
@iambibhas looks good! Hmm the similar artists view might need a separate PR and input from @abeyang because it might need some design love. Can you disable the calling of ddg_spice_lastfm_artist_similar for the mean time?

iambibhas:
@jagtalon sure. For that I'll have to disable some trigger and some tests and comment some parts of the code though. We can implement the similar view after this one.

jagtalon:
@iambibhas thanks!

iambibhas:
@jagtalon removed the trigger `similar` for now.

jagtalon:
@iambibhas thanks for disabling that endpoint! :+1:

PR #2080, opened 2015-08-08
RandWord - moved to randomWords endpoint, moar triggers

### LINKED ISSUE #2062: RandWord: Allow "random words", use `randomWords` API endpoint
Right now the Rand Word Spice only returns a single, random word. We don't show anything for queries like "random words" or "5 random words", etc.

We can easily extend this IA to provide more than one word when requested. We should switch to the `/randomWords` endpoint to allow for this. http://developer.wordnik.com/docs.html#!/words/getRandomWords_get_3

------
IA Page:  http://duck.co/ia/view/rand_word

iambibhas:
Fixes #2062.

**Bug Found:** For the `concat` helper of Handlebars, there is a bug with the use of `,` - 

screenshot from 2015-08-08 12 26 05

There is an extra space before the comman. Also, above 5 words, there is an extra comma before `and`. How to fix this?

-----
IA Page: https://duck.co/ia/view/rand_word


moollaza:
Checkout the Rhymes Spice, the concat helper is working correctly there.

https://github.com/duckduckgo/zeroclickinfo-spice/search?utf8=%E2%9C%93&q=concat

iambibhas:
@moollaza even on the rhymes there is a `,` before the `and`. The prefixed space on this one is a bit confusing. Not sure why it's being added.

iambibhas:
@moollaza fixed the prefixed space issue. Apparently it was being added because of indented template code. Changed this - 

```
{{#concat words sep=", " conj=" and "}}
    <a href="http://wordnik.com/words/{{word}}">{{word}}</a>
{{/concat}}
```

to this 

```
{{#concat words sep=", " conj=" and "}}<a href="http://wordnik.com/words/{{word}}">{{word}}</a>{{/concat}}
```

.. and now it works. There is still a space before `and` though.

screenshot from 2015-08-08 20 15 09

moollaza:
>There is still a comma before and though.

Oh sorry, for the confusion, that's expected. Checkout https://en.wikipedia.org/wiki/Serial_comma. 

Its meant to separate the last two items in the list so that the "and" doesn't look like its joining the final two items into 1.

moollaza:
>fixed the prefixed space issue. Apparently it was being added because of indented template code.

You can actually tell Handlebars to ignore any whitespace inside the block using `~`. Checkout http://handlebarsjs.com/expressions.html#whitespace-control

In your case:

```handlebars
{{#concat words sep=", " conj=" and "~}}
    <a href="http://wordnik.com/words/{{word}}">{{word}}</a>
{{~/concat}}
```

iambibhas:
> Oh sorry, for the confusion, that's expected. Checkout https://en.wikipedia.org/wiki/Serial_comma.

That, is news to me. Thanks.

> You can actually tell Handlebars to ignore any whitespace inside the block using ~.

Oh, ok. Lemme put that in place and then you can go ahead and merge.

iambibhas:
@moollaza done.

jagtalon:
@iambibhas sweet! Getting @abeyang to comment on it.

iambibhas:
@abeyang probably wont like the blue hyperlinks on that grey. :P

lib/DDG/Spice/RandWord.pm
```diff
@@ -15,19 +15,29 @@ icon_url "/i/wordnik.com.ico";
...
-        return $1;
-    } elsif ($_ eq '') {
-        return '0-100';
+    my $minlen = 1;
+    my $maxlen = 100;
+    my $limit = 1;
+
+    if ($_ =~ /^([0-9]+)\-([0-9]+)$/) {
+        # Quesries like "random word 5-10"
+        $minlen = $1;
```
moollaza: Should probably compare these and make sure min is actually smaller?

iambibhas: Hmm. Right.

lib/DDG/Spice/RandWord.pm
```diff
@@ -15,19 +15,29 @@ icon_url "/i/wordnik.com.ico";
...
+        $maxlen = $2;
+    } elsif ($_ =~ /^\d+$/) {
+        # Queries like "5 random words"
+        $limit = $_;
```
moollaza: Is there a limit on this from the API? We may want to give it an upper-bound like 20. I haven't checked to see how that looks yet...

share/spice/rand_word/rand_word.js
```diff
@@ -2,13 +2,15 @@
     "use strict";
     env.ddg_spice_rand_word = function(api_result) {
 
-        if (!api_result || !api_result.word) {
+        if (!api_result || api_result.length == 0) {
```
moollaza: Let's use `===`

lib/DDG/Spice/RandWord.pm
iambibhas: There is no upper limit mentioned in the doc. I've tried 200 and it worked. :-/

moollaza:
@iambibhas @abeyang my only thought is if we should actually be using the `list` template here? Seems silly that we need a handlebars helper to create this for us. The template should be able to do this if we give it the right data...

Also the list template would enable us to show more results as they'd be suppressed initially.

We don't have a nice way of showing a list like this though with the template so this is probably best for now.

iambibhas:
@moollaza which one is the `list` template? Not

moollaza:
@iambibhas looks like "random words" doesn't actually return multiple words yet. Let's default to show...10? words for that query.

lib/DDG/Spice/RandWord.pm
```diff
@@ -15,19 +15,33 @@ icon_url "/i/wordnik.com.ico";
...
+
+    if ($_ =~ /^([0-9]+)\-([0-9]+)$/) {
+        # Quesries like "random word 5-10"
+        $minlen = $1;
+        $maxlen = $2;
+
+        if ($minlen > $maxlen) {
+            ($minlen, $maxlen) = ($maxlen, $minlen)
+        }
+    } elsif ($_ =~ /^\d+$/) {
```
moollaza: I'm not sure if this should be an `elsif`. I think it may be best to `handle_query_lc` and then use a regex to check for `/\d+? random words? (\d+\-\d+)?/`

So that we can allow the number of words and the min/max length of each to be specified.

jagtalon:
Hey @iambibhas. @abeyang recommended that we use the list template similar to NameDays list having 4 columns: https://duckduckgo.com/?q=name+day+Maria

Each column would just contain the random word which links to Wordnik, and we'll see how that looks.

What do you think?

iambibhas:
@jagtalon lemme try and show you how it looks.

jagtalon:
@iambibhas thank you

iambibhas:
@jagtalon looks like it's not a template but custom HTML? https://github.com/duckduckgo/zeroclickinfo-goodies/blob/master/lib/DDG/Goodie/NameDays.pm#L129

jagtalon:
Hey, @iambibhas! Sorry for not being clear. You can use the list template (https://duck.co/duckduckhack/template_groups#list-template-group) and enabling the `meta.template.options.list_content` (https://github.com/duckduckgo/duckduckgo-answerbar-templates/blob/578e094324133f235c9ef06d3f8860443a9a9d20/detail/list_detail.handlebars#L17...#L23) by adding this to `Spice.add`:

```javascript
templates: {
    options: {
        list_content: Spice.rand_word.list_content
    }
}
```

Inside `list_content.handlebars`, you can have a custom grid as seen in https://duckduckgo.com/styleguide:

screen shot 2015-09-17 at 1 59 57 pm

So in your case you'd have something like:

```handlebars
<div class="gw">
  <div class="g quarter">You can use</div>
  <div class="g quarter">custom</div>
  <div class="g quarter">grid layouts!</div>
 <div class="g quarter">grid layouts!</div>
</div>
```

iambibhas:
@jagtalon sorry for the confusion because I didn't see `list` template group under *Template Group Reference* here - https://duck.co/duckduckhack/template_groups

We should probably add *List* to the list.

iambibhas:
This is what it looks like now - 

selection_016


@jagtalon @moollaza TBH, I've been feeling this for a while - when we're starting to work on an issue with the UI, the designer should look at it first and then we should work on any frontend code. Or at least designer should review after the first commit. Because this has happened multiple times, including this issue, that we work on one type of UI for sometime and then we have to change it to something totally different after the designer review.

moollaza:
@iambibhas thanks for the honesty. We're actively trying to improve the process and we should have handled this PR differently. We're going to start posting more Visual Design update tasks so they can be handled separately by the community, rather than using other opportunities (like a bug fix) to also have developers change the UI unless its absolutely necessary.

Apologies for the frustrations and as always, thanks a lot for the excellent work, we really appreciate it.

moollaza:
@jagtalon @iambibhas @abeyang live on beta here: https://beta.duckduckgo.com/?q=random+words&ia=answer

moollaza:
This looks pretty good. I think Chris had mentioned that we should aim to have "random word" return a dictionary result for a random word. That can be made into a v2 task as well.

moollaza:
Also, we may want to put an upper-bound on the number of words, but the result is OK given that the template supresses the overflow.

https://beta.duckduckgo.com/?q=1000+random+words&ia=answer&iax=1

iambibhas:
@moollaza the problem with imposing limits is that we have to let the user somehow about that.

moollaza:
>@moollaza the problem with imposing limits is that we have to let the user know somehow about that.

@iambibhas I suppose that information would belong in the IA Page description? But I don't think we need to indicate it in the SERP. They'll be able to figure it out pretty quickly by testing if they want to.

We obviously can't allow someone to request crazy amounts like 5000 or 1M random words (I'm sure the API at some point will stop giving us results too) so we need to impose a hard limit and any queries above that won't garner a response. The same thing happens for most other IAs that accept input parameters. For example the calculator, at some point it stops responding when the calculations are too complex.

iambibhas:
@moollaza cool. So what should the limit be? 500? 1000 seems a bit much.

moollaza:
@iambibhas sure 500 seems good for now

PR #2087, opened 2015-08-11
turbopape's Clojure Packages Search from Clojars

turbopape:
This IA accesses Clojars.org, the central repository of clojure packages, to search for an artifact and give all possible matches with latest versions and pointers to it on Clojars website.

This IA solves the problem of getting a description of a Clojure package from Clojars as well as a pointer to it while searching for Clojure artifacts on DuckDuckGo.

https://clojars.org/search?q=some-lib&format=json

Clojars is the defacto standard repository for clojure packages. Leiningen, for instance, uses it.

As far as know, no. There is cross-clj but it is more or less based on clojars to derive its dependency graphs.

clojure instaparse
clojure lib zip
clojure library something
clojars a-lib
clojure package some-lib
clojure package some lib 
clojars library some lib and some other therms

Clojure developers and learners.

Yes. to and I A I suggested but I could not claim, as it was mine: 
https://duck.co/ideas/idea/6216/search-for-a-clojure-package-on-clojars

No one, as far as I know.

no.

clojars API is free to use.

I heard about DuckDuckHack after I got interested by DuckDuckGO and adopted it as my search engine. I've red a bit about its open-source architecture, and loved the concept.

ia-clojars-2

-----
IA Page: https://duck.co/ia/view/clojars


moollaza:
@turbopape thanks a lot for this! We'll be sure to review this shortly and give you some feedback!

lib/DDG/Spice/ClojurePackagesFromClojars.pm
```diff
@@ -0,0 +1,43 @@
+package DDG::Spice::ClojurePackagesFromClojars;
```
moollaza: Lets simplify this name! `Clojars` (like Npm, RubyGems, Hoogle) is fine, or `ClojurePackages` (similar to SublimePackages)

share/spice/clojure_packages_from_clojars/clojure_packages_from_clojars.js
```diff
@@ -0,0 +1,51 @@
+(function(env) {
+    "use strict";
+    env.ddg_spice_clojure_packages_from_clojars = function(api_result) {
+
+        // Validate the response (customize for your Spice)
+        var search_terms = DDG.get_query().split(" ");
+
+        function keep_it(candidate) {
```
moollaza: take a look at using a `relevancy: {}` block in the docs. It'll handle all of this for you :+1:

share/spice/clojure_packages_from_clojars/clojure_packages_from_clojars.js
```diff
@@ -0,0 +1,51 @@
...
+                sourceUrl: 'https://clojars.org/search?q=' + clojar_artifact
+            },
+            templates: {
+                group: 'text',
+                detail: false,
+                item_detail: false,
+
+                variants: {
+                    tileTitle: 'basic4',
+                    detail: 'light'
```
moollaza: Don’t think you need this for anything. You're not showing a detail pane :+1:

share/spice/clojure_packages_from_clojars/clojure_packages_from_clojars.js
```diff
@@ -0,0 +1,51 @@
...
+                }
+            },
+            normalize: function(item) {
+                return {
+                    title: item.group_name + "/" + item.jar_name,
+                    subtitle: item.version,
+                    description: item.description,
```
moollaza: Seeing as `item.description` exists, you don't need to re-specify it here. We're actually extending the `item` obj with whatever you return here, so you don't lose any of the original properties from `item`

turbopape: Thank you for your feedback. I'll adress all of this and let you know when
I'm finished.

turbopape:
@moollaza I pushed a modified version as per your suggestions. But your suggestion about using relevancy is not appliable in my case: actually, I use it to get the terms of the query, and not to filter results. I saw that, for instance, news.js (in the tutorial) used a similar approach than mine.

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,50 @@
+(function(env) {
+    "use strict";
+    env.ddg_spice_clojars = function(api_result) {
+
+        // Validate the response (customize for your Spice)
+        var search_terms = DDG.get_query().split(" ");
+
+        function keep_it(candidate) {
```
moollaza: Let's move this out of the callback function, right before `ddg.spice_clojars`

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,50 @@
...
+        if (!api_result || api_result.error) {
+            return Spice.failed('clojars');
+        }
+
+        // Render the response
+    
+        Spice.add({
+            id: "clojars",
+            // Customize these properties
+            name: 'On Clojars',
```
moollaza: Let's change this to "Software" or "Packages"

t/Clojars.t
```diff
@@ -0,0 +1,41 @@
...
+use strict;
+use warnings;
+use Test::More;
+use DDG::Test::Spice;
+
+spice is_cached => 1;
+
+ddg_spice_test(
+    [qw( DDG::Spice::Clojars)],
+    # At a minimum, be sure to include tests for all:
```
moollaza: nit: you can remove this comment, and the one below. They're just for you to read and see

lib/DDG/Spice/Clojars.pm
```diff
@@ -0,0 +1,43 @@
+package DDG::Spice::Clojars;
+# ABSTRACT: This spide searches for an artifact on Clojars.
+
+
+use DDG::Spice;
+
+# Caching - https://duck.co/duckduckhack/spice_advanced_backend#caching-api-responses
```
moollaza: You can also remove this comment and any others that don't explain this specific code.

lib/DDG/Spice/Clojars.pm
```diff
@@ -0,0 +1,43 @@
...
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/ClojurePackagesFromClojars.pm";
+attribution github => ["turbopape", "Rafik Naccache"],
+            twitter => "turbopape";
+
+
+spice to => 'https://clojars.org/search?q=$1&format=json&callback={{callback}}';
+spice wrap_jsonp_callback => 1;
+
+
+triggers any => "clojure", "clojars","clojure lib", "clojure library";
```
moollaza: Let's use `startend` to avoid over triggering. Otherwise we'll be triggering this for any query that has "clojure" anywhere. Most of those queries are likely not to be a search for a packge/module.

lib/DDG/Spice/Clojars.pm
```diff
@@ -0,0 +1,43 @@
...
+
+# Handle statement
+handle remainder => sub {
+
+    # optional - regex guard
+    # return unless qr/^\w+/;
+
+    return unless $_;    # Guard against "no answer"
```
moollaza: Do we need to cleanup the query at all? Would user's search something like "clojure package $name" or "clojars package for $name"? I'm not exactly sure how a user might search for a Clojure package, but if those are likely we should handle them appropriately so we don't send extra text to the API.

lib/DDG/Spice/Clojars.pm
```diff
@@ -0,0 +1,43 @@
...
+category "programming";
+
+topics "programming","computing";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/ClojurePackagesFromClojars.pm";
+attribution github => ["turbopape", "Rafik Naccache"],
+            twitter => "turbopape";
+
+
+spice to => 'https://clojars.org/search?q=$1&format=json&callback={{callback}}';
+spice wrap_jsonp_callback => 1;
```
moollaza: You don't want `wrap_jsonp_callback` if you're passing a `callback` param to the API. If the API supports the callback param it will wrap it for you. In case it doesn't, then you can use this, and remove `&callback=...` from the `spice to`

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,50 @@
...
+                sourceName: "clojars.org",
+                sourceUrl: 'https://clojars.org/search?q=' + clojar_artifact
+            },
+            templates: {
+                group: 'text',
+                detail: false,
+                item_detail: false,
+
+                variants: {
+                    tileTitle: 'basic4'
```
moollaza: This should be `tile`: 'basic4' for it to work correctly :+1:

turbopape:
I addressed all of your remarks.
As per the point regarding cleaning the entry, I will just replace space between terms with hyphens, so if wen enter : clojure incanter cascalog, we'll search incanter-cascalog, as this is the format fo clojure packages naming.

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,51 @@
...
+                sourceName: "clojars.org",
+                sourceUrl: 'https://clojars.org/search?q=' + clojar_artifact
+            },
+            templates: {
+                group: 'text',
+                detail: false,
+                item_detail: false,
+
+                variants: {
+                    tile: 'basic4'
```
moollaza: Instead of `basic4` I think we should use:

```js
tileTitle: '2line',
tileSnippet: 'small'
```

This will reduce the tile height, and give the title 2lines of room so we don't clip it as much.

moollaza:
@turbopape thanks for the updates! This is looking great. I think we just need to adjust the snippet size and title size and this should be about done.

@abeyang any thoughts on design? I've suggest we move to a 2line title and small snippet. The result looks like this:

clojars_packages_for_math_at_duckduckgo

turbopape:
I changed the design as per your recommendation @moollaza @abeyang.

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,52 @@
...
+        return DDG.isRelevant(candidate, ['clojure', 'clojars', 'lib', 'library','pacakge'], 1, 0);
+    }
+
+        var clojar_artifact = search_terms.filter(keep_it);
+    env.ddg_spice_clojars = function(api_result) {
+
+        // Validate the response (customize for your Spice)
+      
+
+        if (!api_result || api_result.error) {
```
jagtalon: Does `api_result.error` actually show up when there's an error?

t/Clojars.t
```diff
@@ -0,0 +1,45 @@
...
+
+ddg_spice_test(
+    [qw( DDG::Spice::Clojars)],
+    'clojure instaparse' => test_spice(
+        '/js/spice/clojars/instaparse',
+        call_type => 'include',
+        caller => 'DDG::Spice::Clojars'
+    ),
+    'clojure lib instaparse' => test_spice(
+    '/js/spice/clojars/instaparse',
```
jagtalon: Nitpick: Please indent this as:

```perl
    'clojure lib instaparse' => test_spice(
        '/js/spice/clojars/instaparse',
        call_type => 'include',
        caller => 'DDG::Spice::Clojars'
    ),
```

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,52 @@
...
+                detail: false,
+                item_detail: false,
+
+                variants: {
+                    tile: '2line',
+                    tileSnippet: 'small'
+                }
+            },
+            normalize: function(item) {
+                return {
```
jagtalon: Check if `group_name`, `jar_name`, and `version` exists before accessing them?

lib/DDG/Spice/Clojars.pm
```diff
@@ -0,0 +1,39 @@
...
+
+use DDG::Spice;
+
+spice is_cached => 1;
+
+name "Clojars";
+source "clojars.org";
+icon_url "https://clojars.org/favicon.ico";
+description "This spice accesses clojars.org, the clojure libraries reposisitory, to search for a clojure artifact";
+primary_example_queries "clojure some-lib", "clojars some-lib","clojure lib some-lib","clojure library some-lib";
```
jagtalon: Could you add in real packages in here?

jagtalon: Like probably `clojure template` to show its capabilities.

abeyang:
Suggestion is good, @moollaza

jagtalon:
Created an IA page for this: https://duck.co/ia/view/clojars

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,52 @@
...
+        if (!api_result || api_result.error) {
+            return Spice.failed('clojars');
+        }
+
+        // Render the response
+    
+        Spice.add({
+            id: "clojars",
+            // Customize these properties
+            name: 'Packages',
```
jagtalon: Please change the tab name to "Software" like in https://duckduckgo.com/?q=rubygems+sass&ia=software

share/spice/clojars/clojars.js
turbopape: Right, there's no such thing as api_result.error. Instead, I've put api_result.count, which when shows 0, means that no result was shown.

t/Clojars.t
turbopape: Done.

share/spice/clojars/clojars.js
turbopape: I Added the check, I return nothing if any of these fields is not found.

lib/DDG/Spice/Clojars.pm
turbopape: Done.

share/spice/clojars/clojars.js
turbopape: Done.

turbopape:
New PR as per all of the remarks.

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,54 @@
...
+            },
+            normalize: function(item) {
+                if (item.group_name && item.jar_name && item.version){
+                    return {
+                        title: item.group_name + "/" + item.jar_name,
+                        subtitle: item.version,
+                        url: "https://clojars.org/" + item.group_name + "/" + item.jar_name
+                    };
+                } 
+                return;
```
moollaza: this should `return null`

share/spice/clojars/clojars.js
```diff
@@ -0,0 +1,54 @@
+(function(env) {
+    "use strict";
+    var search_terms = DDG.get_query().split(" ");
+
+    function keep_it(candidate) {
+        return DDG.isRelevant(candidate, ['clojure', 'clojars', 'lib', 'library','pacakge'], 1, 0);
+    }
+
+        var clojar_artifact = search_terms.filter(keep_it);
```
moollaza: indent is a little off here

moollaza:
@turbopape this LGTM :+1: A couple non-essential fixes to be made. But I'll merge for now :smile:

turbopape:
Cool ! (Y)

moollaza:
@turbopape congrats! It's live: https://duckduckgo.com/?q=clojure+template&ia=software

PR #2088, opened 2015-08-11
twitter - not showing divider if no location is available

### LINKED ISSUE #2085: Twitter: Don't show vertical separator when location undefined
It looks like we're always displaying the vertical separator that goes between the handle and location of the user. We don't need to show it though when there's no location!

Example: https://duckduckgo.com/?q=twitter+testing&t=ffab&ia=social

------
IA Page:  http://duck.co/ia/view/twitter

iambibhas:
Fixes #2085.

screenshot from 2015-08-12 00 51 43

### DIFF
--- a/share/spice/twitter/twitter.js
+++ b/share/spice/twitter/twitter.js
@@ -35,10 +35,15 @@
                 sourceName: "Twitter",
             },
             normalize: function(item) {
+                var subtitle = ["@" + item.user, ];
+                if (item.location) {
+                    subtitle.push(item.location);
+                }
+
                 return {
                     image: bigger_picture(item.profile_image),
                     title: item.name,
-                    subtitle: ["@" + item.user, item.location],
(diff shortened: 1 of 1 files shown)


share/spice/twitter/twitter.js
```diff
@@ -35,10 +35,15 @@
                 sourceName: "Twitter",
             },
             normalize: function(item) {
+                var subtitle = ["@" + item.user, ];
```
moollaza: looks like you have a trailing `,` in there

moollaza:
@iambibhas thanks! Just a small fix and it should be good :+1:

share/spice/twitter/twitter.js
iambibhas: Sorry, habit from Python. It should be ok though. Means that there is just one item in the array.

PR #2089, opened 2015-08-11
twitch - increased limit of stream search

### LINKED ISSUE #2084: Twitch: Add limit= param to API call, set limit to 20
By default the Twitch API only returns 10 items. We require 12 items to allow the GridMode button to work but that can be changed. It seems like we're best off pulling in a few more results anyways so let's add `&limit=20` to the API calls for both Twitch Instant Answers.

iambibhas:
Fixes #2084.

screenshot from 2015-08-12 01 00 17


moollaza:
@iambibhas LGTM :+1: 

Thanks!

Issue #2077, opened 2015-08-07
Timer: visual modifications


abeyang:
@andrey-p , timer is almost there! Here are several comments to improve it visually and experientially. /cc @chrismorast 

First, by default, I think we should set it at 1:00, like so (also note that the outer ring, `.progress`, is a lighter gray, `#eee`):

![image](https://cloud.githubusercontent.com/assets/794084/9143545/f993deb0-3cfb-11e5-9b01-89ea3c71abce.png)

Second, as the timer is going, the Title should **not** have the input field behind it, like so:
![image](https://cloud.githubusercontent.com/assets/794084/9143594/382719e4-3cfc-11e5-94c1-30a10396662e.png)

However, **on hover** we should be able to see the input field behind the title, and it should be **editable**.

 Moreover, we need to align the fields a bit:

```css
.zci--timer .time_display {
  margin-top: -1.6em;
  margin-left: .3em;
}
```

Third, the +1 should look more like Chris' design -- LMK if you want css help with this:
![image](https://cloud.githubusercontent.com/assets/794084/9143916/1eda3154-3cfe-11e5-85fd-605f398b59af.png)

Fourth, the progress ring inside seems a bit off... like a couple of pixels to the top, and a couple of pixels to the left. We'll need to fix that... this is what it looks like at 800%..
![image](https://cloud.githubusercontent.com/assets/794084/9143977/95099afe-3cfe-11e5-8c4e-70a3d8f54e17.png)

A few more soon to come..

mattr555:
Woah, this looks really cool (commenting to follow, lmk if you need my help with the old code)

andrey-p:
Hey @abeyang, I'm moving the discussion to a PR so we can comment on this as I add new bits to it. I'll implement this round of changes shortly.

@mattr555 I've cc'd you there :)

PR #2091, opened 2015-08-12
Timer improvements

andrey-p:
This never had a PR, it just lived on a branch in my fork while I was working on it. Continuing the discussion on this from https://github.com/duckduckgo/zeroclickinfo-spice/issues/2077

@abeyang 

cc @moollaza @chrismorast @mattr555


andrey-p:
@abeyang Tweaks added (https://ddh10.duckduckgo.com/?q=timer&ia=timer). The one thing I haven't added from there was the timer input field thing - I think I'd like to talk a bit more about the behaviour when we chat later on.

Also, let me know if the +1 button is up to spec  - if not some CSS would be appreciated.

screenshot_2015-08-12_13-45-56

share/spice/timer/timer.handlebars
```diff
@@ -1,20 +1,33 @@
...
+            <div class="time_input">
+                <input type="text" maxlength="2" class="hours" placeholder="HH">
+                <span class="separator">:</span>
+                <input type="text" maxlength="2" class="minutes" placeholder="MM">
+                <span class="separator">:</span>
+                <input type="text" maxlength="2" class="seconds" placeholder="SS">
+            </div>
+        </div>
+    </div>
+    <a href="" class="corner_btn reset">RESET</a>
```
moollaza: What about using `<button>` elements? I think those would be more appropriate?

share/spice/timer/timer.js
```diff
@@ -7,290 +7,430 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
...
+    }
+
+    //play the alarm sound
+    function playLoopingSound() {
+        function requirePlayer(player) {
+            player.play('alarm-sound', soundUrl, {
+                autoPlay: true
+            });
         }
+        DDG.require('audio', requirePlayer);
```
moollaza: @andrey-p I noticed on the live site (and here) the sound no longer works. Any ideas what's going on? I haven't investigated it yet.

andrey-p: That's definitely an issue, and not only on the Timer IA. Will investigate.

moollaza: :+1:

share/spice/timer/timer.css
```diff
@@ -1,233 +1,321 @@
...
 
-.zci--timer .timer__btn.timer__start[disabled] {
-    background-color: #ccc;
-    border-color: #ccc;
-    color: #f2f2f2;
+.zci--timer #add_timer_container {
+    display: block;
+    width: 140px;
+    height: 140px;
+    background-color: #e6e6e6;
```
moollaza: I think we need to use our special `bg-clr--foo` and `tx-clr--foo` CSS classes so that colours work across themes:

timer_at_duckduckgo

share/spice/timer/timer.css
```diff
@@ -1,233 +1,321 @@
...
+.zci--timer .play_pause .fill {
+    display: none;
+    top: -4px;
+    left: -4px;
+    position: absolute;
+    width: 100px;
     height: 100px;
-    margin-left: 5px;
-    margin-right: 5px;
+    clip: rect(0, 100px, 100px, 50px);
```
moollaza: I haven't seen this property before! MDN seems to suggest we should use `clip-path` instead, but that seems to be a largely experimental/unsupported feature.

https://developer.mozilla.org/en-US/docs/Web/CSS/clip

share/spice/timer/timer.css
```diff
@@ -1,233 +1,321 @@
...
+}
+
+.zci--timer .timer_container.status_paused .rotated_fill,
+.zci--timer .timer_container.status_paused .first_half_fill {
+    border-color: #959595;
+}
+
+.zci--timer .play_pause a {
+    font-family: "ddg-serp-icons";
+    color: #333;
```
moollaza: Using `tx-clr--slate` will make sure this looks good in the dark theme :+1:

share/spice/timer/timer.js
```diff
@@ -7,290 +7,430 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
...
+    function padZeros(n, len) {
+        var s = n.toString();
+        while (s.length < len) {
+            s = '0' + s;
+        }
+        return s;
+    }
+
+    // split seconds into hours, minutes, seconds
+    function getHrsMinsSecs(totalSeconds) {
```
moollaza: Wondering if Moment.js will make any of this easier?

moollaza:
@andrey-p it looks like we're changing the value of the timer too quickly when you start the timer. We immediately change the value to 1 second less then the input, and so for short times (< 5sec) there's actually a 1 second delay between the number reaching 0 and the circle filling completely. We should wait 1 second to update the timers after they start.

share/spice/timer/timer.js
mattr555: Odd, sound works for me here and on the live site (Linux, Chrome)

mattr555:
> We immediately change the value to 1 second less then the input

To fix this we did `Math.ceil()` to the difference [last time](https://github.com/mattr555/zeroclickinfo-spice/commit/3a77b1c9e830d54edec5029e464c9c3d0913b1cf?w=1#diff-186e35eeb541cc28cf6434a65c435dceR38)

share/spice/timer/timer.js
andrey-p: Possibly, but using a library might add extra overhead and we're running this every 100ms. Also, Moment.js hadn't been added to the libraries when I worked on this last :)

share/spice/timer/timer.css
andrey-p: Yeah, it'll probably take years to be fully phased out. It's what we use on the progress loaders in the Soundcloud IA too.

share/spice/timer/timer.handlebars
andrey-p: Not sure if there's a massive difference and I generally find them more of a hassle to style... I don't mind changing them if you'd rather have them as buttons though.

share/spice/timer/timer.js
andrey-p: @mattr555 Yeah, it was a weird browser-specific issue - a fix is in the pipeline for it.

andrey-p:
> it looks like we're changing the value of the timer too quickly when you start the timer.

@moollaza Good catch! @mattr555's suggestion did the trick.

abeyang:
Here's the css fix for the input field alignment, as promised:

``` css
.zci--timer .name_input {
  width: 9.05em;
  margin-bottom: 0.8em;
}

.zci--timer .time_display {
  margin-top: -1.2em;
}
```

andrey-p:
@abeyang Thanks!

Um, I ran into a bit of a stumbling block making the timer tiles scroll - basically, while it's possible, it'll require a lot of extra work. In the interest of shipping this sooner, how do you feel about limiting the number of timers so we don't have to worry about scroll?

andrey-p:
@abeyang Updated version of this is on ddh10: https://ddh10.duckduckgo.com/?q=timer+1+min+20+sec&ia=timer

There was already a tick in the font library, but it looks slightly different - if you're not cool with the way it looks I'll add the blockier one to the symbol font.

screenshot_2015-08-17_15-19-53

abeyang:
I'll give it a whirl. But are we still gonna implement the `timer mm:ss` query? So, `timer 1:15` will set the timer to 1 min 15 sec.

abeyang:
Minor tweaks to the `+1`:

pasted_image_8_17_15__11_04_am

abeyang:
Finally, I thought I asked this elsewhere (but I could be trippin'), how much harder would it be to get the scrolling to work?

abeyang:
Sorry, a few more things. With regards to the tile margins, can we remove the spacing above and below it, like so:

``` css
.zci--timer .tile {
  margin-top: 0;
  margin-bottom: 0;
}
```

And with the "More" tile, can we make it more like **Images** for Desktop, and **Recipes** for mobile? So it should look like--

For desktop (doesn't have to say "more", and it should still be a square):
image

For mobile (show more => add more):
image

andrey-p:
Tweaks added. That silver typo is embarrassing, thanks for spotting that... :)

screenshot_2015-08-19_11-57-40
screenshot_2015-08-19_12-02-35

>  But are we still gonna implement the timer mm:ss query? So, timer 1:15 will set the timer to 1 min 15 sec.

This is implemented, but I'm not sure how to update the triggering on the server without properly deploying it (@moollaza do you know if this is possible?). If you try it from Duckpan it'll work.

> how much harder would it be to get the scrolling to work?

I wrote above:

> Um, I ran into a bit of a stumbling block making the timer tiles scroll - basically, while it's possible, it'll require a lot of extra work. In the interest of shipping this sooner, how do you feel about limiting the number of timers so we don't have to worry about scroll?

By "a lot of extra work" I mean I'll have to circumvent the Spice API and use the core DDG classes. Definitely possible, but it'll be a significant rewrite of the Javascript.

abeyang:
@andrey-p saw the update. First, let's remove that previous fix on `.zci--timer .tile`... let's put the margins back in for each tile. But in addition to that, we'll need this:

``` css
.zci.is-active.zci--timer {
  padding-bottom: 0;
  padding-top: 0;
}
```

andrey-p:
@abeyang I'm pretty happy with this :)

screenshot_2015-08-20_11-21-11

abeyang:
@andrey-p per our conversation:

1. Let's have **RESET** be explicit on the timer when it's completed. Both clicking on it and the checkmark would do the same thing.
2. Mobile: finished sound isn't working
3. Number fields should have [type="number"](http://www.w3.org/TR/html-markup/input.number.html#input.number)
4. When clicking on a number field, if there's a number in there already (like 15 sec), that number should be selected.

Thanks!

andrey-p:
@abeyang Done!

2 will probably require an internal fix, I'll work on that separately.

About 3, I ended up going for `type="tel"` instead as `type="number"` was breaking the leading zeros in some browsers.

abeyang:
Alrighty, looks good! Let's merge this -- we'll work out the audio fixes separately. /cc @chrismorast

PR #2092, opened 2015-08-12
go_watch_it - added stop words and tests

### LINKED ISSUE #2070: GoWatchIt: Should ignore the query "pocket watch"
Unfortunately the query "pocket watch" triggers this IA and get's past our relevancy checking. 

https://duckduckgo.com/?q=pocket+watch&t=ffab&ia=watch

We may need to adjust the relevancy checking threshold to consider all words with 3+ characters, but it may be best to simply check if the query matches this term and if so, `return;`

------
IA Page:  http://duck.co/ia/view/go_watch_it

iambibhas:
Fixes #2070.

 - Added list of stop words
 - added tests for them

### DIFF
--- a/lib/DDG/Spice/GoWatchIt.pm
+++ b/lib/DDG/Spice/GoWatchIt.pm
@@ -18,9 +18,11 @@ attribution github  => ['https://github.com/plexusent', 'GoWatchIt.com'],
             twitter => ['gowatchit', 'GoWatchIt.com'];
 
 my @triggers = ('watch', 'stream', 'watch online', 'on demand', 'watch now', 'stream online', 'buy movie', 'rent movie','movie');
-my @killwords = ('movie', 'show', 'tv', 'online', 'stream');
+my @ignorewords = ('movie', 'show', 'tv', 'online', 'stream');
+my @stopwords = ('apple', 'pocket', 'night', 'tower');
 
-my $killwords = join '|', @killwords;
+my $ignorewords = join '|', @ignorewords;
+my $stopwords = join '|', @stopwords;
 
 triggers startend => @triggers;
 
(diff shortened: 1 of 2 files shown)


lib/DDG/Spice/GoWatchIt.pm
```diff
@@ -30,9 +32,10 @@ spice wrap_jsonp_callback => 1;
 handle remainder_lc => sub {
 
   return unless $_; # Guard against "no answer"
-  $_ =~ s/$killwords//g; # remove killwords
-  return trim($_); # trim spaces and return
+  return if $_ =~ m/^$stopwords$/g; # Guard against stop words "apple watch"
```
moollaza: I think we should compare the full query against these as the order matters: "apple watch" vs "watch apple".

We expose the full query in the `$req` object.

lib/DDG/Spice/GoWatchIt.pm
```diff
@@ -30,9 +32,10 @@ spice wrap_jsonp_callback => 1;
...
 
+  $_ =~ s/$ignorewords//g; # remove ignorewords
```
moollaza: this needs `\b` around the regex so we don't match a substring.

lib/DDG/Spice/GoWatchIt.pm
moollaza: Also need `\b` here

moollaza:
Ping! @iambibhas just checking in here if you're still interested in finishing this up?

iambibhas:
@moollaza I'll make the changes and send a PR tonight(in 6-8 hours).

moollaza:
LGTM :+1:

moollaza:
@iambibhas thanks a ton!

PR #2094, opened 2015-08-12
New IA: Islamic Prayer Times

ozdemirburak:
Shows Islamic prayer times according to given or user's current location and time

Lets users to find the today's prayer times without providing location and time instead of searching within the list of whole locations, dates and times.

~~http://www.aladhan.com/rest-api~~
http://www.muslimsalat.com/api/

~~It's the most reliable one within the public ones.~~
It lets you search with location name, instead of providing both latitude and longitude.

~~Actually there is, where you can get the results without providing latitude and longitude, instead providing the city's name, which may expand this instant answer. However, within that data source,  after 25K requests, which is insufficient I guess, fee is needed for further requests. But maybe if we contact with them, they may grant unlimited requests.~~ 

~~There are only three queries that trigger this Instant Answer which are 'namaz times', 'salat times' and 'islamic prayer times'. Maybe they can be extended, I may not be aware about all of the Islamic search queries about prayer times.~~
`islamic prayer times`, `izmir turkey namaz times`, `salat times in dc`, `namaz times for san marino`

Muslim people

https://duck.co/ideas/idea/971/the-daily-time-for-the-five-daily-islamic-prayers
https://duck.co/ideas/idea/6241/islamic-prayer-time

None

Currently nope. But I wanna ask you a question, if there exists a DDG GeoLocation service which returns latitude and longitude for given location. I've searched through the documentations and forums, but I might have missed it. 

~~Open for public use. There are not any costs.~~
It needs an API key. It's free for 25K daily requests, for more, extra fee is needed. Details can be seen in [here](http://www.muslimsalat.com/panel/signup.php).

Twitter, DDG official primary account.

~~**With the old API**:~~

islamic-prayer-times-ss1

~~**New API (info):**~~

islamic-prayer-times-ss4

**New API (list):**
islamic-prayer-times-ss10

## Checklist

```

    - Ubuntu 14.04
        [x] Google Chrome
        [x] Firefox
        [] Opera

    - Windows 8
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 10

    - Windows 7
        [] Google Chrome
        [] Firefox
        [] Opera
        [] IE 8
        [] IE 9
        [] IE 10

    - Windows XP
        [] IE 8

    - Mac OSX
        [] Google Chrome
        [] Firefox
        [] Opera
        [] Safari

    - iOS (iPhone)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - iOS (iPad)
        [] Safari Mobile
        [] Google Chrome
        [] Opera

    - Android
        [] Firefox
        [] Native Browser
        [] Google Chrome
        [] Opera
```

https://duck.co/ia/view/islamic_prayer_times


moollaza:
@ozdemirburak thanks a lot for contributing, we really appreciate it! We'll be sure to review this shortly and give you some feedback!

>Actually there is, where you can get the results without providing latitude and longitude, instead providing the city's name, which may expand this instant answer. However, within that data source, after 25K requests, which is insufficient I guess, fee is needed for further requests. But maybe if we contact with them, they may grant unlimited requests. Can be seen in http://www.muslimsalat.com/api/

I think this would be a better choice ultimately. Do you know if they have support for specific locations, or the whole world? It seems like the latter, but I can't tell for sure.

>Currently nope. But I wanna ask you a question, if there exists a DDG GeoLocation service which returns latitude and longitude for given location. I've searched through the documentations and forums, but I might have missed it.

Unfortunately we don't have anything like this yet, but it's on our radar and I personally can't wait to have this functionality! It would certainly be useful for many IA's.

moollaza:
@abeyang any thoughts on design here? Seeing as the prayer times are static, I was wondering if we should opt to show tiles for the prayers of the day, and autoscroll to the tile for the next coming prayer?

ozdemirburak:
@moollaza I agree with you on the API choice and I also guess that it's for the whole world as they are probably using some service to obtain latitude and longitude from city, country or state. Should I better contact with the creators of the API ?

moollaza:
>Should I better contact with the creators of the API ?

@ozdemirburak if you wouldn't mind, that would be awesome. Many API providers have let us use their services for free because we clearly attribute them on the page. 

Is the 25K limit daily? Monthly? If they don't let us use it for free, we can go with the free plan for now and if we get close to the limit, we can get in touch with them again and see if we can partner.

ozdemirburak:
@moollaza It's [daily](http://www.muslimsalat.com/panel/signup.php). I've sent an email about the questions, if I got an answer, I'll post it here.

ozdemirburak:
@moollaza I've updated the code and proposal since I've changed the API as the previous one was limiting the scope of this IA. However, I haven't received any answers upon my email that I've sent to the API creators about the daily limits.

Note: I've created an test account for the API, tested with many locations and I can comfortably tell you that it's worldwide.

moollaza:
>It's daily. I've sent an email about the questions, if I got an answer, I'll post it here.

Ah daily is perfect. I don't think we risk hitting that right now, especially if we cache the results for a bit. Hopefully we'll hear back from them soon!

> I've created an test account for the API, tested with many locations and I can comfortably tell you that it's worldwide.

@ozdemirburak excellent, thanks a lot!

ozdemirburak:
@moollaza I've made some changes as you can see from the latest commit. However I have a another question. Let's say user is located in New York and searching for London prayer times, should we show the `getClosest()` result or not. Because it's currently showing the difference between user's local time, which is New York time, and the closest prayer time from London, so it's a little bit weird, but I'm not sure about how we should handle it.

moollaza:
@ozdemirburak are you asking if we should show the prayer time relative to the user in NY, vs the localtime in London? If so, it should be London local time.

That's how we tend to handle things. The sunrise IA is an example: https://duckduckgo.com/?q=sunrise+australia&t=ffab&ia=answer

ozdemirburak:
@moollaza Actually it's showing in London local time in the example that I've provided. But my question is something different.

 `getClosest()` returns the time difference between user's current local time and the closest prayer time from given location.

To give more accurate example, I'm changing the locations. Let's say, when the user from Izmir, Turkey searches for `salat times izmir turkey` right now, it will return `Fajr, in 5 hours` where Fajr is at 4:43 am tomorrow, in Izmir local time. However when the same user searches for `sydney salat times`, it will return `Fajr, in 5 hours.` where Fajr is at 5:12 am, in Sydned local time.

ozdemirburak:
@moollaza Actually it's showing in London local time in the example that I've provided. But my question is something different.

 `getClosest()` returns the time difference between user's current local time and the closest prayer time from given location.

To give more accurate example, I'm changing the locations. 

Let's say, currently, user from Izmir, Turkey searches for `salat times izmir turkey`, it will return `Fajr, in 5 hours` where Fajr is at 4:43 am tomorrow, in Izmir local time. However when the same user searches for `sydney salat times`, it will return `Fajr, in 5 hours.` where Fajr is at 5:12 am, in Sydney local time. 

However, it's already 7:10 AM in Sydney local time and 00:10 in Izmir local time and Fajr event is already past in Sydney and it's not `in 5 hours` for the user who is in Izmir. 

So, I'm asking that should we add the timezone offset for calculating the closest pray time, or completely remove the closest thing when user searches for another location?

moollaza:
Ah, I think I understand. This is mostly a design problem, so I'd like @abeyang thoughts, but I think we should focus on providing the exact local time of the next prayer. And if we have the current time in that location,  indicate the relative time until Salat in terms of local time. 

Trying to convert and account for the time difference is messy and confusing IMO. 

Does that make sense?

ozdemirburak:
@moollaza Now, showing all the differences according to local time. I think screenshots can be more illuminating.

izmir
nyc
dakar

share/spice/islamic_prayer_times/islamic_prayer_times.js
```diff
@@ -0,0 +1,93 @@
...
+    "use strict";
+    env.ddg_spice_islamic_prayer_times = function(api_result) {
+
+        if (!api_result || api_result.error || api_result.status_code === 0 || !api_result.items) {
+            return Spice.failed('islamic_prayer_times');
+        }
+
+        DDG.require('moment.js', function() {
+
+            function capitalize(string) {
```
jagtalon: You can use the built in function `DDG.capitalize`

jagtalon:
Just created an IA page for this: https://duck.co/ia/view/islamic_prayer_times

ozdemirburak:
@jagtalon Thanks for the feedback, updated.

share/spice/islamic_prayer_times/islamic_prayer_times.js
```diff
@@ -0,0 +1,89 @@
...
+                        infoboxData: getInfoboxData(data.items[index]),
+                        closest: getClosest(data.items[index], offset, local),
+                        qibla: data.qibla_direction
+                    };
+                },
+                templates: {
+                    group: 'info',
+                    options: {
+                        content: Spice.islamic_prayer_times.islamic_prayer_times,
+                        moreAt: false
```
moollaza: Can we possibly construct and send them to a link like http://muslimsalat.com/london/daily for more info?

share/spice/islamic_prayer_times/islamic_prayer_times.js
```diff
@@ -0,0 +1,89 @@
...
+                        datum: local.format('LLLL'),
+                        infoboxData: getInfoboxData(data.items[index]),
+                        closest: getClosest(data.items[index], offset, local),
+                        qibla: data.qibla_direction
+                    };
+                },
+                templates: {
+                    group: 'info',
+                    options: {
+                        content: Spice.islamic_prayer_times.islamic_prayer_times,
```
moollaza: nit: the template file should be named `content.handlebars` so its referenced as `Spice.islamic_prayer_times.content`

moollaza:
@ozdemirburak after speaking with the design team we agreed we have so great ideas for a possible V2 of this IA, but we'd like to simplify it for now and get it live.

Let's switch this to the `List` template and simply show a list of the prayer times for the given location. I think we can stick to a simple header that says `<location> on <date>`. Something like this should work:

> **Next prayer: Fajr at 4:40AM, Aug 14 (Tomorrow)**
> New York, USA 
>
> -----
> Fajr: 4:56 AM

> -----
> Sunrise: 6:23 AM

> -----
> Dhuhr: 1:20 PM

> -----
> Asr: 5:11 PM

> -----
> Maghrib: 8:16 PM

> -----
> Isha: 9:44 PM


/cc @abeyang how does that sound?

ozdemirburak:
@moollaza Updated. This is how it looks like now:

islamic-prayer-times-ss5

And for the tomorrow case, its' header will look like this

islamic-prayer-times-ss6

abeyang:
@ozdemirburak Looking good! Few comments:

1. Can we put the info (prayer and time) into two columns? Similar to the [whois IA](https://duckduckgo.com/?q=whois+facebook.com&ia=whois).
2. Let's change the title/subtitle to this:

> **Fajr at 4:40AM, Aug 14 (Tomorrow)**
> Next prayer time at Phoenixville, PA

Note: For any US locations, let's lose the country ("USA") altogether. Thanks!

ozdemirburak:
@abeyang Updated.

US:
islamic-prayer-times-ss7

Non US:
islamic-prayer-times-ss8

abeyang:
:+1:

share/spice/islamic_prayer_times/islamic_prayer_times.js
```diff
@@ -0,0 +1,97 @@
...
+    "use strict";
+    env.ddg_spice_islamic_prayer_times = function(api_result) {
+
+        if (!api_result || api_result.error || api_result.status_code === 0 || !api_result.items) {
+            return Spice.failed('islamic_prayer_times');
+        }
+
+        DDG.require('moment.js', function() {
+
+            function setDst(time, dst) {
```
moollaza: Let's move these functions out of the callback. They can be moved above `env.ddg_spice_islamic_prayer_times`.

We call them from inside the closure, where `moment` is defined, so its OK :+1:

share/spice/islamic_prayer_times/islamic_prayer_times.js
```diff
@@ -0,0 +1,97 @@
...
+                return state !== "" ? state + ', ' + country : country;
+            }
+
+            Spice.add({
+                id: "islamic_prayer_times",
+                name: "Islamic Prayer Times",
+                data: api_result,
+                meta: {
+                    sourceName: "Muslim Salat",
+                    sourceUrl: api_result.link
```
moollaza: Let's prepend "https://" here

ozdemirburak:
@moollaza Updated.

share/spice/islamic_prayer_times/islamic_prayer_times.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    function tz(time, offset) {
+        var clone = time.clone();
+        clone.utcOffset(offset);
+        clone.add(time.utcOffset() - clone.utcOffset(), 'minutes');
+        return clone;
+    }
+
+    function getClosest(timings, offset, local) {
```
moollaza: I think this code warrant some comments to explain what these are for and why we need them.

lib/DDG/Spice/IslamicPrayerTimes.pm
```diff
@@ -0,0 +1,52 @@
...
+category "time_sensitive";
+topics "special_interest", "everyday";
+
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/IslamicPrayerTimes.pm";
+attribution github  => ["https://github.com/ozdemirburak", "Burak Özdemir"],
+            twitter => ["https://twitter.com/ozdemirbur", "Burak Özdemir"],
+            web     => ["http://burakozdemir.co.uk", "Burak Özdemir"];
+
+spice to => 'http://muslimsalat.com/$1/weekly.json?key={{ENV{DDG_SPICE_MUSLIMSALAT_APIKEY}}}';
+spice wrap_jsonp_callback => 1;
```
moollaza: Looks like the API supports a `callback` param? Let's add `&callback={{callback}}` to `spice to`

moollaza:
@ozdemirburak this is just about done! There's just a couple small things to add and I'll be happy to merge this. Great job implementing this! Thanks a lot for contributing to DuckDuckHack!

ozdemirburak:
@moollaza Updated. Thanks, I should also note that I'm glad that I'm contributing to DuckDuckHack. I hope that I'll contribute more in the future.

share/spice/islamic_prayer_times/islamic_prayer_times.js
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }
+        return list;
+    }
+
+    function setFormat(closest, index) {
+        return index === 0 ? closest.format('h:mm A') : closest.format('h:mm A, MMM D ') + '(Tomorrow)';
+    }
+
+    function setAddress(title, country_code, country, state) {
```
moollaza: Looks like there's a bug in here? I think we should always return the `city` + `country` and in the case of US, `city` + `state`?

Right now it looks like we're skipping the City in all cases, and returning `state` + `country` for US:

namaz_times_for_new_york_at_duckduckgo
namaz_times_for_cairo_at_duckduckgo

moollaza:
@ozdemirburak one more small fix and then this should be good to merge :+1:

share/spice/islamic_prayer_times/islamic_prayer_times.js
moollaza: Hmm looks like the API has some issues with Unicode...
namaz_times_for_tokyo_at_duckduckgo

That's a tough case because the city, state, country are all blank and there's no way for use to know if the unicode is correct.

ozdemirburak:
@moollaza Updated. However, some places don't have the city names, or state names, or addresses. So I've checked the all, if you try Cairo again, it will return the same as nor city name neither address is not returned by the API.

I should also note that, some places give error because of the BIGINT scopes, as the pray times are returned as -9223372036854775808, which is the minimum within the BIGINT scope, from the API, try Istanbul for instance ( actually, Istanbul itself is a state as it's really big place, divided to many cities, some must make the search as "Şişli İstanbul" instead, but people may not think about it ). So I've added an error check for that and pushed into list if they are valid.

moollaza:
@ozdemirburak thanks a lot! This LGTM :+1: 

I noticed on more oddity which is a little difficult to fix. If you specify a state as the input the only part of the adress we get back is the state 2 letter code so "namaz times in California" shows "... at CA".

We may need to use the existing capital cities hash (see the Time Spice) and if the query matches a country name or state name, we may have to alter the API call to specify a specific city so it's clear where the result is coming from.

ozdemirburak:
@moollaza I'll be working on that issue within the next week.

Also, after the DDG location service is completed, this one can even be turned into a goodie, as the times can be calculated easily ( Can be seen [here](http://praytimes.org/calculation/) ).

PR #2095, opened 2015-08-12
Risk/places triggering

brianrisk:
@moollaza Added some triggering to Places.pm.  Am seeing merge flag popping up; advice?


brianrisk:
@moollaza found the trouble; had to push my master rebase back to origin.

lib/DDG/Spice/Quandl/Commodities.pm
```diff
@@ -0,0 +1,64 @@
+package DDG::Spice::Quandl::Commodities;
```
moollaza: Whoops, I don't think this is supposed to be in here?

brianrisk: (doh)  removed.

PR #2098, opened 2015-08-13
Add guard for empty lastfm response

bsstoner:
Guard against empty/bad api responses from last.fm

screen shot 2015-08-13 at 8 47 44 am

@moollaza @russellholt

### DIFF
--- a/share/spice/lastfm/artist/lastfm_artist.js
+++ b/share/spice/lastfm/artist/lastfm_artist.js
@@ -2,6 +2,8 @@
     "use strict";
 
     env.ddg_spice_lastfm_artist_all = function(api_result) {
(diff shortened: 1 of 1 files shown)


nilnilnil:
:+1: I'd love to see a helper that you give a name like `api_results.top.tracks` and it walks the object hierarchy to validate that exists. Those statements seem common & overly long.

moollaza:
@nilnilnil  we have `DDG.getProperty()`?

```js
var track = DDG.getProperty(api_resut, "top.tracks");
if (tracks === undefined) { return; }
```

bsstoner:
I think it's much clearer to just enumerate and return. Any method obfuscates what's actually being checked.

PR #2100, opened 2015-08-13
Parking: Reduce results array to 30 elements to increase page load speed

moollaza:
@AlexCutlipParkingPanda We're noticing that the page is taking too long to load on mobile so I'm fixing like this for now.

As a better solution, can your API return fewer results initially? OR can it take a request parameter that specifies the number of results? Right now it's sending ~140 which is too many. 

/cc @bsstoner @andrey-p 

Live here: https://moollaza.duckduckgo.com/?q=parking+manhattan&ia=parking

### DIFF
--- a/share/spice/parking/parking.js
+++ b/share/spice/parking/parking.js
@@ -1,25 +1,25 @@
 (function (env) {
     "use strict";
-    
+
     var userLanguage = navigator.languages || // New (experimental)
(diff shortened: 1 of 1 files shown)


moollaza:
Definitely improves the mobile experience on my Nexus 6.

bsstoner:
:+1: thanks

Issue #2090, opened 2015-08-11
Quandl HomeValues: doesn't work when query is in the form "home values in [state|zip code]"


PalmerAL:
If the search term is in the form "home values in [state]", the result shown is for Indiana no matter what state is entered.
Examples:
https://duckduckgo.com/?q=home+values+in+california&ia=homevalues
https://duckduckgo.com/?q=home+values+in+new+york&ia=homevalues

However, the above search terms work if the word "in" is removed.

------
IA Page:  http://duck.co/ia/view/quandl_home_values

moollaza:
@PalmerAL thanks a lot for reporting! 

/cc @brianrisk

MrChrisW:
Thanks @PalmerAL 

Resolved by @brianrisk in #2105 

https://duckduckgo.com/?q=home+values+in+california&ia=homevalues
https://duckduckgo.com/?q=home+values+in+new+york&ia=homevalues

PR #2105, opened 2015-08-17
quandl home values: fixed indiana bug

brianrisk:
The problem was that the abbreviation for "Indiana" is "IN", so "home values in California" was triggering on the "in".

### DIFF
--- a/share/spice/quandl/home_values/metro.yml
+++ b/share/spice/quandl/home_values/metro.yml
@@ -308,7 +308,6 @@ hudson:	00694
 humboldt:	00448
 huntingdon:	00742
 idaho falls:	00286
-indiana:	00551
 indianapolis:	00035
 indianola:	00927
 iowa city:	00252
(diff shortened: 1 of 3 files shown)


brianrisk:
@moollaza this addresses this bug:
https://github.com/duckduckgo/zeroclickinfo-spice/issues/2090

moollaza:
@brianrisk could we update the Perl logic to better handle this so we can still allow queries for Indiana?

something like `s/\bin\b(\w+)/$1/` perhaps?

brianrisk:
@zaahir, sorry for the confusion.  I'm only getting rid of the "in" abbreviation.  "home values in Indiana" will still trigger the IA.

jagtalon:
@brianrisk looks good, although it would be good to update the tests as well.

brianrisk:
@jagtalon Good suggestion -- Done!

moollaza:
LGTM :+1:

PR #2108, opened 2015-08-17
Remove view from SoundCloud, no longer necessary

bsstoner:
@andrey-p 
cc @moollaza


moollaza:
LGTM :+1:

moollaza:
Whoops, wrong PR. But I'm pretty sure this is good too ;)

bsstoner:
Wait, I jumped the gun here. This will work, but might break in the case where there's only 1 result returned by sound cloud. Want to make a few internal changes first before doing this.

moollaza:
:+1:

PR #2115, opened 2015-08-19
AQI: properly fail on an empty array

mattr555:
When testing I noticed that we cause an error when the array is empty (no result). This tests for that and properly calls `Spice.failed`

https://duck.co/ia/view/aqi


moollaza:
@mattr555 thanks a lot! LGTM :+1:

PR #2119, opened 2015-08-20
(Sports) Fix 'Current' Assignment for MLB

sdougbrown:
This corrects a mistaken assumption that I made when I tried to adjust how `fillBoxScore` gets the 'current' scoring sequence.

to @bsstoner

### DIFF
--- a/share/spice/sports/sports.js
+++ b/share/spice/sports/sports.js
@@ -63,8 +63,9 @@
 
             // Game Finished/In-Progress
             if (attrs.has_started) {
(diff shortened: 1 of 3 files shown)


share/spice/sports/sports.js
```diff
@@ -63,8 +63,9 @@
 
             // Game Finished/In-Progress
             if (attrs.has_started) {
+                console.log(ops);
```
bsstoner: remove console.log please? :)

Issue #2093, opened 2015-08-12
Arch User Repository: undefined parameter in "More at aur.archlinux.org" link


claudiodangelis:
The "More at aur.archlinux.org" link has an undefined parameter (&arg=undefined). Tested on Firefox and Chrome.

http://aur.archlinux.org/rpc.php?type=search&arg=undefined

![image](https://cloud.githubusercontent.com/assets/941963/9224029/a2a29e4a-40ff-11e5-914a-41b2102fcecb.png)

------
IA Page:  http://duck.co/ia/view/aur

moollaza:
@claudiodangelis thanks a lot for reporting this!

moollaza:
@MrChrisW as always, thanks for fixing it :)

PR #2120, opened 2015-08-21
Arch User Repository: Fix sourceUrl

MrChrisW:
Fixing #2093

### DIFF
--- a/share/spice/aur/aur.js
+++ b/share/spice/aur/aur.js
@@ -21,7 +21,7 @@
             data: api_result.results,
             meta: {
                 sourceName: "aur.archlinux.org",
-                sourceUrl: 'https://aur.archlinux.org/packages/?O=0&SeB=nd&K=' + query
+                sourceUrl: 'https://aur.archlinux.org/packages/?K=' + query
             },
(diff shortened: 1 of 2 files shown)


share/spice/aur/aur.js
```diff
@@ -16,7 +21,7 @@
             data: api_result.results,
             meta: {
                 sourceName: "aur.archlinux.org",
-                sourceUrl: 'http://aur.archlinux.org/rpc.php?type=search&arg=' + api_result.name
+                sourceUrl: 'https://aur.archlinux.org/packages/?K=' + query
```
moollaza: Why use the query instead of the package name?

MrChrisW: @moollaza There's no single "package name" returned, if multiple results are found for the query, so `api_result.name` won't work.

This wouldn't even work for a single returned result as the package name is under the results array.

Checkout an example API response:
https://aur.archlinux.org/rpc.php?type=search&arg=firefox

moollaza: :+1:

moollaza:
LGTM :+1:

PR #2121, opened 2015-08-21
Remove 'haskell' trigger from Hayoo IA.

cskksc:
@moollaza  Refers to [PR](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1977).

/cc @headprogrammingczar


jagtalon:
@cskksc maybe https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/t/Hayoo.t should be updated as well. Also does Hoogle give the same results as Hayoo? If that's the case maybe it should trigger on "hoogle" as well?

moollaza:
LGTM :+1:

PR #2122, opened 2015-08-21
Fix number of stories condition for adding news

tommytommytommy:
The correct condition for calling Spice.fail() for news should be if we have less than 3 good stories. The condition was correct in https://github.com/duckduckgo/zeroclickinfo-spice/blob/d010d02cf7dbd095561ee5d36f1505d3cc10f1c4/share/spice/news/news.js#L80, and then the inequality was flipped in https://github.com/duckduckgo/zeroclickinfo-spice/blob/0cfee129b2e052c264d234b57f183524f6b4acb5/share/spice/news/news.js#L80.

This PR flips the inequality condition and only allows us to call either Spice.failed() or Spice.add(), but not both. (I think that the current code always calls Spice.add() even if we call Spice.fail()?)

### DIFF
--- a/share/spice/news/news.js
+++ b/share/spice/news/news.js
@@ -2,6 +2,11 @@
     "use strict";
     env.ddg_spice_news = function (api_result) {
 
+        if (!api_result) {
+            Spice.failed('news');
+            return;
+        }
+
         // Words that we have to skip in DDG.isRelevant.
         var skip = [
             "news",
@@ -77,7 +82,7 @@
 
         var searchTerm = DDG.get_query().replace(/(?: news|news ?)/i, '').trim();
 
-        if (api_result.error || goodStories < 3) {
+        if (goodStories < 3) {
             Spice.failed('news');
 
         } else {
--- a/share/spice/news/news.js
+++ b/share/spice/news/news.js
@@ -79,20 +79,22 @@
 
         if (api_result.error || goodStories < 3) {
             Spice.failed('news');
-        }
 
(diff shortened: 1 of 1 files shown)


moollaza:
@tommytommytommy who should review? me?

tommytommytommy:
@moollaza Sure, please :) Thanks!

share/spice/news/news.js
```diff
@@ -77,22 +77,24 @@
 
         var searchTerm = DDG.get_query().replace(/(?: news|news ?)/i, '').trim();
 
-        if (api_result.error || goodStories >= 3) {
+        if (api_result.error || goodStories < 3) {
```
jagtalon: Does `api_result.error` actually show up when there's an error? It looks like it just returns `Spice.failed` when you put in random stuff. https://duckduckgo.com/js/spice/news/acsd%20asdfasdfasdf

tommytommytommy: Hmmm, I was just copying what was there. Unclear about that...maybe we should just do an `!api_result` guard? Will investigate and get back.

tommytommytommy:
@jag Yeah, it looks like we're creating a hash of story articles and not creating/filling in `api.error` anywhere. Removed it from the check.

moollaza:
@tommytommytommy I made a couple fixes to the `Spice.failed()` calls. I assume they should be `return Spice.failed(...)` like usual?

moollaza:
LGTM :+1:

PR #2124, opened 2015-08-22
Fix flight status

nathanntg:
The flights and routes instant answers do not correctly show the status for delayed flights. After a little investigation, this appears to be related to the handlebars template, which is not passing in the needed scheduled arrival or scheduled departure information. This pull request includes updates to both files to pass the appropriate variables.

I was able to test this for the routes IA only (I think the airline one may only have deep triggers), but it appears to produce the desired effect:

napkin 08-22-15 3 24 29 pm

I only tested in one browser, because this did not change the HTML output or the JavaScript. Let me know if there are further tests or prep work I should complete.

Thanks!

### DIFF
--- a/share/spice/airlines/content.handlebars
+++ b/share/spice/airlines/content.handlebars
@@ -11,7 +11,7 @@
   </div>
   <div>
     <span class="tile__status">
-      {{{airline_status flight departureDate arrivalDate}}}
+      {{{airline_status flight departureDate arrivalDate scheduledDepartureDate scheduledArrivalDate}}}
     </span>
   </div>
 </div>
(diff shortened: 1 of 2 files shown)


moollaza:
@nathanntg thanks a lot, I'll review this shortly!

/cc @tommytommytommy

tommytommytommy:
I don't know if I can be an impartial reviewer, seeing as we live together. Haha. It looks good to me! Thanks for the fix :)

moollaza:
>I don't know if I can be an impartial reviewer, seeing as we live together. 

I thought this "Nathan" guy looked familiar!

> It looks good to me! Thanks for the fix :) 

Tested and working :+1: 

Thanks a ***ton*** @nathanntg 

Proof:
dl_1275_at_duckduckgo_and_delta_flight_1275_at_duckduckgo

PR #2125, opened 2015-08-24
Detail ClassName Updates

sdougbrown:
We're going to be refactoring how detail panes work, in order to make it easier for open-source developers to create and modify these areas.  The first step is to break up the detail 'wrapping' elements from the 'content' elements.  This just tags a bunch of those content elements with the `.c-` prefix, wherever they appear in the Spice repo.

to @bsstoner


bsstoner:
:+1:

bsstoner:
@moollaza this depends on internal changes that need to be released.

moollaza:
:+1: It didn't make it into the most recent release, so I think that one is ok to move ahead?

bsstoner:
:+1: Yep, go for it!

PR #2134, opened 2015-08-26
Better display of flight status

nathanntg:
Inspired by my last PR... :-) Here are a few more tweaks to the flights and routes IAs that I think improve the handling of flight statuses. The exact changes are described below. Some of the changes are worthy of discussion, not sure if there is a right answer.

Changes:

* Eliminated the handlebars helper for flight status, since it was used once and made things more circuitous.
* Made status such as canceled, diverted, redirect and unknown print in "red" (```tile__not```) instead of "green" (```tile__ok```), to more accurate reflect the unexpected status. (See the attached screenshot.)
* Made flights that have landed late, just based on arrival time, print in "red" (```tile__not```) instead of "green" (```tile__ok```) to reflect the deviation from the scheduled arrival time. This change is more debatable, as I can see arguments for either style.

I tried to make changes in both the routes & airline IA at the same time.

Screenshot from testing, showing that the new handlebars code works and highlighting the second change describe above:

napkin 08-26-15 6 55 12 pm

### DIFF
--- a/share/spice/airlines/airlines.js
+++ b/share/spice/airlines/airlines.js
@@ -244,6 +244,9 @@
                     isDeparted: false
                 };
 
+            // Status
+            var status = onTime(flight[i], departureDate, arrivalDate, scheduledDepartureDate, scheduledArrivalDate);
+
             results.push({
                 flight: flight[i],
                 airlineName: airlineName,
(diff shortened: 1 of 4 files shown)


moollaza:
@nathanntg thanks a lot for this. These are great suggestions, but we're actually in the process of finalizing a PR that will add a whole new UI to the flights tiles, which you can see over here: #1636 

I think the updated design takes these into consideration, but if not, I'll make sure it does :+1:

nathanntg:
Awesome, that looks great! It looks like it already touches on the key things, and I love the refreshed designed. I will close this up.

moollaza:
:+1:

PR #2135, opened 2015-08-27
removed use of global isNumber helper

andrey-p:
@moollaza I was doing some cleaning up of the DDG utils and noticed one of them exposed `DDG.isNumber` in the global scope. I'll be removing `isNumber` from the global scope to keep things tidy.

The geometry spice is the only place where the global `isNumber` is used instead of `DDG.isNumber`.


moollaza:
LGTM :+1:

moollaza:
@andrey-p are the internal changes merged yet?

PR #2139, opened 2015-08-29
Polls Spice

alexander95015:
Search for polls on HuffPost Pollster

Polls on HuffPost Pollster are displayed

http://elections.huffingtonpost.com/pollster/api

This source aggregates data and has a good API.

Not sure.

election polls
US polls

People who read polls

No

None

I'm having trouble getting it to link to the chart page. It's provided in the JSON, but I'm not sure how to do the linking inside the DuckDuckHack framework.

No costs. Not sure about the terms of use. http://elections.huffingtonpost.com/pollster/api

screen shot 2015-08-29 at 9 06 53 pm

-----
IA Page: https://duck.co/ia/view/polls


MariagraziaAlastra:
Hey @alexander95015, thanks for your contribution!
Given that you're PR isn't connected to an approved IA idea we need to evaluate it first, but we'll get back to you soon :)

//cc @moollaza

moollaza:
@alexander95015 @MariagraziaAlastra this looks like it should be OK to move ahead.

@alexander95015 please submit an idea for this at https://duck.co/ideas so other's know about this. Then post a link to the thread here so I can mark it as approved, and you can then "claim" it so others know you're working on it.

@alexander95015 My biggest concern right now is the response size and the speed. 300+kb is too big. Let's use the `page=` param to specify the first page initially. Seeing as this doesn't require an API key, you can also set the `next` and `allowMultipleCalls` properties so we can lazy load extra results as users scroll to them.

Checkout https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/amazon/amazon.js#L13 as an example.

You'll need some logic that determines what the URL for the `next` property should be, and when it should stop (perhaps after 5 pages?)

lib/DDG/Spice/Polls.pm
```diff
@@ -0,0 +1,42 @@
+package DDG::Spice::Polls;
```
moollaza: I think `Spice::ElectionPolls` or `Spice::HuffPostPolls` is a little better? I'm worried "Polls" is a little too generic

alexander95015: Some of these polls are not really election polls. How about `Spice::PoliticalPolls`?

share/spice/polls/polls.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+        
+        if(results.length > 50){
+            results = results.splice(50);
+        }
+
+        // Render the response
+        DDG.require("moment.js", function() {
+        Spice.add({
+            id: "polls",
```
moollaza: Once the idea is created, approved and claimed an IA Page will automatically be created for this Instant Answer. We'll need to ensure the `id` here matches the `id` we assign to the page. `election_polls` should be good :+1:

moollaza:
@alexander95015 I forgot to say thanks a lot for contributing, we really appreciate it! :smiley:

lib/DDG/Spice/Polls.pm
moollaza: Sounds good :+1:

MariagraziaAlastra:
@alexander95015 now you can claim the idea, I added the quack and hack tag: https://duck.co/ideas/idea/6315/election-polls

alexander95015:
@MariagraziaAlastra Would there be an error now that I renamed it to `PoliticalPolls` while the IA page says `Election Polls`?

MariagraziaAlastra:
@alexander95015 not really, but it's better to have consistent naming. You can edit the name in the [IA Page](https://duck.co/ia/view/polls) and then click on "submit" to save the changes (that button saves any change for the fields in the top blue header). :)

alexander95015:
@MariagraziaAlastra I can't seem to edit it for some reason.

MariagraziaAlastra:
@alexander95015 Nevermind, I edited it. I'll be sure to give you edit permissions for that page :)

alexander95015:
@moollaza The `page` param you are referring to doesn't apply to `charts` calls, which is what's used here. It's only for `poll` calls. Is there some other way?

MariagraziaAlastra:
@alexander95015 I checked, you should have the edit permissions already...

alexander95015:
@MariagraziaAlastra I can edit most of it, just not the name of the IA.

MariagraziaAlastra:
@alexander95015 oh alright, I forgot the name is only editable by admins!

share/spice/political_polls/political_polls.js
```diff
@@ -0,0 +1,58 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_political_polls = function(api_result){
+
+        // Validate the response (customize for your Spice)
+        if (!api_result || api_result.error) {
```
jagtalon: Maybe you can also check if the  `api_result` array has more than one element?

share/spice/political_polls/political_polls.js
```diff
@@ -0,0 +1,58 @@
...
+            return Spice.failed('political_polls');
+        }
+        
+        var results = api_result;
+
+        if (!results) {
```
jagtalon: Looks like this is redundant because you already checked the truthiness of `api_result`

share/spice/political_polls/political_polls.js
```diff
@@ -0,0 +1,58 @@
...
+            return Spice.failed('political_polls');
+        }
+        
+        if(results.length > 50){
+            results = results.splice(50);
+        }
+
+        // Render the response
+        DDG.require("moment.js", function() {
+        Spice.add({
```
jagtalon: Please indent this, i.e., 

```javascript
DDG.require("moment.js", function() {
    Spice.add({...});
});
```

alexander95015:
@moollaza Is there some other way to fix the response size issue?

moollaza:
> @moollaza Is there some other way to fix the response size issue?

@alexander95015 any reason the pagination doesn't work?

alexander95015:
@moollaza I'm not sure, but apparently the `page` parameter only works for the `polls` method and not the `charts` method this is calling.

share/spice/political_polls/political_polls.js
```diff
@@ -0,0 +1,67 @@
...
+                id: "political_polls",
+
+                // Customize these properties
+                name: "Polls",
+                data: api_result,
+                meta: {
+                    sourceName: "HuffPost Pollster",
+                    sourceUrl: 'http://elections.huffingtonpost.com/pollster',
+                },
+                templates: {
```
jagtalon: Please add 

```
variants: {
    tileSnippet: "large"
}
```

in here so that we'd get something like this

screen shot 2015-09-15 at 1 19 41 pm

Instead of this

screen shot 2015-09-15 at 1 19 56 pm

share/spice/political_polls/political_polls.js
```diff
@@ -0,0 +1,67 @@
...
+                    group: 'text',
+                    options: {
+                        detail: false,
+                        item_detail: false,
```
jagtalon: `detail` and `item_detail` should be in the `template` object instead of the `options` object:

```javascript
templates: {
    group: 'text',
    detail: false,
    item_detail: false,
    ...
}
```

jagtalon:
@alexander95015 cool thanks! We're gonna put this on a server and give it a final design pass

jagtalon:
@alexander95015 @moollaza currently up on https://comserv.duckduckgo.com/?q=election+polls&ia=polls

lib/DDG/Spice/PoliticalPolls.pm
```diff
@@ -0,0 +1,42 @@
...
+topics "special_interest";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/PoliticalPolls.pm";
+attribution github => ["alexander95015", "Alexander"],
+            facebook => "alexander95015",
+            twitter => "alexander95015";
+
+spice wrap_jsonp_callback => 1;
+
+# API endpoint - https://duck.co/duckduckhack/spice_attributes#spice-codetocode
+spice to => 'http://elections.huffingtonpost.com/pollster/api/charts.json';
```
jagtalon: @alexander95015 how often does the data update?

jagtalon: I'm asking because I'd like to cache the data appropriately. :)

lib/DDG/Spice/PoliticalPolls.pm
```diff
@@ -0,0 +1,42 @@
...
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#category
+category "reference";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
+topics "special_interest";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/PoliticalPolls.pm";
+attribution github => ["alexander95015", "Alexander"],
+            facebook => "alexander95015",
+            twitter => "alexander95015";
+
+spice wrap_jsonp_callback => 1;
```
jagtalon: It looks like it can accept a callback parameter so I think we can get rid of this? http://elections.huffingtonpost.com/pollster/api

jagtalon:
@alexander95015 there's a lot of results returned (492 items), and the JSON returned is too big (300kb). Can we filter to just use 2016 polls? http://elections.huffingtonpost.com/pollster/api

lib/DDG/Spice/PoliticalPolls.pm
alexander95015: @jagtalon The data updates whenever there's a new poll.

lib/DDG/Spice/PoliticalPolls.pm
alexander95015: I think they do, but they require the callback function to have a certain name, so it doesn't really work here. http://elections.huffingtonpost.com/pollster/api/charts.json?callback=test

alexander95015:
@jagtalon Wouldn't that mean that we have to constantly update the filter? (i.e. when the election is over and such) How about limiting to just national polls? http://elections.huffingtonpost.com/pollster/api/charts?state=US

jagtalon:
@alexander95015 Ah true. Limiting to just national polls would be great! Anything to get the amount of data down. :)

jagtalon:
@alexander95015 Thanks--checking it now.

jagtalon:
:+1:

moollaza:
@alexander95015 congrats! It's live: https://duckduckgo.com/?q=election+polls&ia=polls

PR #2140, opened 2015-08-29
Update for clarity, add gif demo

moollaza:
/cc @zekiel @tagawa @talsraviv


zekiel:
too large to host on imgur er sthing?

moollaza:
I used cloud app to make the gif, so it automatically hosts them. Can move elsewhere if needed. Can put it in the repo I guess

moollaza:
Looks like GitHub Caches it anyways on their own servers, so the request is going to github

PR #2149, opened 2015-09-02
Timer tweaks

### LINKED ISSUE #2147: Timer creates a new timer every time the tab shows
if you query 'timer', you get one timer;
if you then switch to another tab, and then switch back to the timer, you get two timers, etc.

@andrey-p 

------
IA Page:  http://duck.co/ia/view/timer

andrey-p:
Fixes #2147

This PR will also have the timer sound looping, but I'll add this once we've cleared up some interaction stuff with @abeyang.

cc @russellholt 

-----
IA Page: https://duck.co/ia/view/timer

### DIFF
--- a/share/spice/timer/timer.js
+++ b/share/spice/timer/timer.js
@@ -410,8 +410,8 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
             timerInterval,
             $addTimerBtn;
 
-        function addTimer() {
-            var timer = new Timer(timers.length + 1);
+        function addTimer(startingTime) {
+            var timer = new Timer(timers.length + 1, startingTime);
             timer.$element.insertBefore($addTimerBtn.parent());
             timers.push(timer);
         }
@@ -426,7 +426,7 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
 
             // have at least one timer when the IA is displayed
             if (timers.length === 0) {
-                addTimer();
+                addTimer(enteredTime);
             }
 
             // every 100 ms, update timers
--- a/share/spice/timer/timer.js
+++ b/share/spice/timer/timer.js
@@ -11,6 +11,7 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
         SOUND_NAME = "alarm-sound",
         soundUrl = DDG.get_asset_path('timer', 'alarm.mp3'),
         soundIsPlaying = false,
+        hasShown = false,
         $lastTimerToFinish,
         Timer,
         cachedPlayer;
@@ -459,6 +460,13 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
         }
         
         function onShow() {
+            // make sure this runs only once
+            if (hasShown) {
+                return;
+            }
+
+            hasShown = true;
+
             var lastUpdate = new Date().getTime(),
                 enteredTime = parseQueryForTime(),
                 $dom = Spice.getDOM("timer"),
@@ -501,11 +509,6 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
             });
         }
 
-        function onHide() {
-            $addTimerBtn.off("click");
-            clearInterval(timerInterval);
-        }
-
         Spice.add({
             id: 'timer',
             name: 'Timer',
@@ -524,8 +527,7 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
 
             //wait for the spice to load before displaying things
             //this makes sure the divs display at the right time so the layout doesn't break
-            onShow: onShow,
-            onHide: onHide
+            onShow: onShow
         });
         
 
--- a/share/spice/timer/timer.js
+++ b/share/spice/timer/timer.js
@@ -10,6 +10,7 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
     var MAX_TIME = 359999, // => 99 hrs 59 mins 59 secs
         SOUND_NAME = "alarm-sound",
         soundUrl = DDG.get_asset_path('timer', 'alarm.mp3'),
+        soundIsPlaying = false,
         Timer,
         cachedPlayer;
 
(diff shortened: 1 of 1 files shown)


share/spice/timer/timer.js
```diff
@@ -73,14 +76,45 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
...
+            setTimeout(playAlarm, 500);
+            return;
         }
-        DDG.require('audio', requirePlayer);
+
+        // start looping sound - single click anywhere on the screen will
+        // stop looping
+        loop();
+        soundIsPlaying = true;
+        $(document).one("click", stopLoop);
```
moollaza: typo? `$document.on(`

moollaza: Aah I see: http://api.jquery.com/one/ -- That's neat!

share/spice/timer/timer.js
```diff
@@ -406,19 +440,31 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
...
+            timerInterval,
+            $addTimerBtn;
+
+        function addTimer(startingTime) {
+            var timer = new Timer(timers.length + 1, startingTime);
+            timer.$element.insertBefore($addTimerBtn.parent());
+            timers.push(timer);
+        }
         
         function onShow() {
```
moollaza: I think really we should set onShow to fire once be default. We've been adding logic to several IA's now to explicitly check if the event has fired and prevent any further calls.

Alternatively we should make a new `init()` or `build()` (like we have for Goodies) that only fires once because that's almost always what we want out of the `onShow()` event

russellholt: @moollaza point me to examples where you check that it's only doing it once?

moollaza: @russellholt I thought I had seen it in more Spices, but it looks like right now it's only Amazon: 

https://github.com/duckduckgo/zeroclickinfo-spice/blob/4e967c6672c9aeaadb54652502270dd1ce159c57/share/spice/amazon/amazon.js#L50

Almost every Spice that defines an `onShow` really only needs it to run once though, and right now they're running every time we switch back to the tab. In many cases we're using it to add classes to divs, define handlers for events, or load more data -- all of which should only be done once.

For example, [movies](https://duckduckgo.com/?q=movies&t=ffab&ia=movies) is loading the images for each tile from a 3rd party in `onItemShown` -- I just realized it's running each time we switch to the tab, causing a lot of unnecessary requests. 

---

For Goodies, we're using a check to make sure it only runs once for CheatSheet.js (https://github.com/duckduckgo/zeroclickinfo-goodies/blob/2173d26caf4a33480ea7518d3c18c666e6de5267/share/goodie/cheat_sheets/cheat_sheets.js#L91) and we'll need it for the 2048 game (still in-progress), to bind our event-handlers to DOM elements

moollaza:
Should be live on Beta shortly: https://beta.duckduckgo.com/?q=timer+4+seconds

andrey-p:
@abeyang Added repeated timer shaking.

@moollaza Sweet, completely forgot about the beta server! Does it update instantly or is there a bit of a delay?

share/spice/timer/timer.js
russellholt: @moollaza 
the purpose of `onShow` is to run whenever the tab is shown. it should never not be used to load more data, but doing something only once isn't exactly an engineering feat. If that's a common pattern, we can control for it. but `onShow` will always fire when it's time to, for example, restart your animation or resume playing a video.

Care to add anything to that @bsstoner ?

bsstoner: :+1: We implemented it this way to provide a lower level of control over things from Spice. We could add abstractions on top, like `onFirstShow` or something.

abeyang:
@moollaza can we get this on beta too? Thanks!

moollaza:
@andrey-p @abeyang deploying again to beta with the new updates...should be live in a few minutes

share/spice/timer/timer.js
moollaza: >We could add abstractions on top, like onFirstShow or something.

:+1: That would be great!

jagtalon:
@andrey-p @abeyang I'd like to add one more update to this one. The "+" button doesn't seem to be working with the dark theme very well:

https://vid.me/7qyp

jagtalon:
@andrey-p I think you should stop propagation on keydown as well. So when I focus on the input box and click the right or left keys, it switches the IA. It makes editing a bit annoying. :)

https://vid.me/c52f

jagtalon:
@andrey-p setInterval in `onShow` is getting stopped when we switch tabs. It should probably be outside  of the `onShow` function.

https://vid.me/Yabg

share/spice/timer/timer.js
```diff
@@ -73,14 +77,45 @@ License: CC BY-NC 3.0 http://creativecommons.org/licenses/by-nc/3.0/
...
             });
+
+            return;
+        }
+
+        // if a sound is already playing, stop for a moment
+        // and then start again
+        if (soundIsPlaying) {
+            stopLoop();
+            setTimeout(playAlarm, 500);
```
jagtalon: `playAlarm` doesn't seem to exist?

moollaza:
>@andrey-p @abeyang I'd like to add one more update to this one. The "+" button doesn't seem to be working with the dark theme very well:

@andrey-p @jagtalon I believe I've fixed this in the live version

share/spice/timer/timer.js
andrey-p: Gah. Thanks for spotting that.

andrey-p:
> @andrey-p setInterval in onShow is getting stopped when we switch tabs. It should probably be outside of the onShow function.

@jagtalon Yeah, I was wondering about this - my original thinking was that the timer shouldn't run if you switch to a different IA, but then again, it continues running when you switch to a different browser tab.

I just tried adding that locally and it felt a bit weird, having a sound coming from nowhere without any visual feedback. Then again, the sound stops playing on your first click so maybe it's not a massive problem?

russellholt:
i think its fine if it continues in the background, i think thats whats expected

jagtalon:
@andrey-p that's true. Although running in the background is what's is expected and the same thing happens for the SoundCloud IA. Maybe we could have the design team make up an icon of sorts to show what tab sound is coming from?

andrey-p:
Fair enough - I've changed it so it runs in the background.

andrey-p:
> Maybe we could have the design team make up an icon of sorts to show what tab sound is coming from?

I thought it would be useful to discuss this further, I've opened up an internal task for it.

jagtalon:
@andrey-p thanks! ok doing some final checks before merging this!

jagtalon:
:+1:

PR #2150, opened 2015-09-02
sort news by date

### LINKED ISSUE #2129: News: Results should be sorted by age
I just realized our news results aren't sorted by age! We should definitely show the most recent results first!

Let's add a `sort_default` block to News.js and sort these results by recency.

------
IA Page:  http://duck.co/ia/view/news

marianosimone:
This should fix #2129

-----
IA Page: https://duck.co/ia/view/news

### DIFF
--- a/share/spice/news/news.js
+++ b/share/spice/news/news.js
@@ -55,6 +55,7 @@
             story.title = story.title.replace(/<b>|<\/b>|:/g, "");
             if (DDG.isRelevant(story.title, skip)) {
                 setSourceOnStory(story);
+                story.sortableDate = parseInt(story.date || 0);
                 goodStories.push(story);
             }
 
@@ -97,7 +98,13 @@
                 },
                 templates: {
                     item: 'news_item'
-                }
+                },
+                sort_fields: {
+                    date: function(a, b) {
+                        return b.sortableDate - a.sortableDate;
+                    }
+                },
+                sort_default: 'date'
             });
         }
     }


share/spice/news/news.js
```diff
@@ -55,6 +55,7 @@
             story.title = story.title.replace(/<b>|<\/b>|:/g, "");
             if (DDG.isRelevant(story.title, skip)) {
                 setSourceOnStory(story);
+                story.sortableDate = parseInt(story.date);
```
marianosimone: trying to do this only when necessary, and only once for each element instead of when comparing

tommytommytommy:
This idea sounds good to me! The only change that I would make is to guard against the date possibly being undefined (perhaps just assign those articles a date of "0"?). It's possible for some stories to come back with no dates (shouldn't happen often, but the probability is nonzero).

marianosimone:
good idea, @tommytommytommy. I fixed that!

tommytommytommy:
@marianosimone Thanks!

@moollaza The code looks good to me. I'm not sure what the process is for testing/merging because it's a fork; would you mind pinging me when you put this on a test server so that I can follow along for future PRs? Handing back to you to complete.

moollaza:
@marianosimone it's live on beta: https://beta.duckduckgo.com/?q=news+america&ia=news

everything LGTM  :+1: 

Thanks a ton!

PR #2152, opened 2015-09-03
GoWatchIt: Small updat, remove experimental smartmatch operator

moollaza:
/cc @iambibhas

### DIFF
--- a/lib/DDG/Spice/GoWatchIt.pm
+++ b/lib/DDG/Spice/GoWatchIt.pm
@@ -19,7 +19,7 @@ attribution github  => ['https://github.com/plexusent', 'GoWatchIt.com'],
 
 my @triggers = ('watch', 'stream', 'watch online', 'on demand', 'watch now', 'stream online', 'buy movie', 'rent movie','movie');
 my @ignorewords = ('movie', 'show', 'tv', 'online', 'stream');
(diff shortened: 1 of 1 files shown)


lib/DDG/Spice/GoWatchIt.pm
```diff
@@ -31,7 +31,8 @@ spice wrap_jsonp_callback => 1;
 handle remainder_lc => sub {
 
   return unless $_; # Guard against "no answer"
-  return if ($req->query_lc ~~ @stopwords); # Guard against stop words "apple watch"
+  return if grep {$req->query_lc eq $_} @stopwords;
```
iambibhas: @moollaza can you explain a little how the grep is working? I thought the block was something that we searched for in the array(`@stopwords` in this case). But the block is a conditional in this case?

moollaza: @iambibhas it's pretty odd, but grep acts like `map` when you give it a block expression. The difference is that map returns the value of the block, whereas grep returns the current element, if the block evals to true.

So for each item we check if the current item matches our query, and if so we return the item. This of course produces an array as our final value, and if we get an array we want to bail in this case.

Does that make sense?

PR #2153, opened 2015-09-03
News Images

bsstoner:
Requires internal changes, don't merge yet.


share/spice/news/news.js
```diff
@@ -93,10 +93,27 @@
...
                     count: goodStories.length,
                     searchTerm: searchTerm,
-                    itemType: 'News articles'
+                    itemType: 'News articles',
+                    rerender: [
+                        'image'
+                    ]
                 },
                 templates: {
                     item: 'news_item'
```
moollaza: Should we switch to the new-er `Media` template? We use it for [Dogo News](https://duckduckgo.com/?q=dogo+news&t=ffab&ia=kidsnews).

bsstoner: Oh, that's interesting, maybe. I'll defer to @chrismorast and @abeyang, maybe there's a way to consolidate the two.

bsstoner:
:+1: internal PR merged

PR #2155, opened 2015-09-04
Modify Spice.add and Spice.failed ID for maps

tommytommytommy:
Requires internal coordination before merge/release!


b2ddg:
@moollaza I tested this one, and it's ready to merge. Needs coordination with https://dub.duckduckgo.com/core/ddg/pull/4832

Issue #2156, opened 2015-09-07
CitiBike: Need to normalize dates to standard format and UTC Timezone


moollaza:
I realized that the date string we get from the CitiBike API (https://www.citibikenyc.com/stations/json) is a non-standard date format which the `new Date()` constructor cannot parse. Also, it's given in EDT instead of UTC, but the string doesn't indicate the timezone, causing further problems.

@marianosimone do you know if we can get CitiBike to give us a proper, UTC formatted string? As well, can you update the code to replace the `-` with `/` then convert the time to UTC (you'll probably need to set timezone for the date to EDT first)

This way when dealing with other timezones everything will be normalized :+1: 

/cc @xuv

-----
IA Page: https://duck.co/ia/view/citi_bike_nyc

xuv:
@marianosimone: Here's the line of code that worked for me to replace the `-` with `/` (on line 90):  
```lastCommunication: moment(new Date( (item.lastCommunicationTime).replace(/\-/g,'/') ) ).fromNow()```

marianosimone:
Quick update: The people behind the data feeds for CitiBike NYC (and many other bike sharing services) released a new format called [GBFS](https://github.com/NABSA/gbfs), and they won't be adding features/fixing the legacy ones anymore.

- The good news is that the new format handles dates better (using Posix timestamps)
- The bad news is that they splitted the feed in two parts (actually, many more, some of them interesting for us, but not directly related to this), so the station information (like location, name and address) is separated from the status (how many available bikes there are)... So we can't migrate to it without significantly changing how we do it (as I understand, spices can only hit one url to get the data)
- The hopeful news is that I've been talking with one of the people from the project, and he encouraged me to open a [new ticket](https://github.com/NABSA/gbfs/issues/21) asking for an aggregate feed that would allow us to keep the current implementation

PR #2157, opened 2015-09-07
set timezone to lastCommunicationDate string before parsing as date

marianosimone:
Investigating #2156, I discovered that there's no problem with the date format, except for the fact that the timezone is NOT specified, so it depends on the user's timezone.

Adding EDT to the `lastCommunicationTime` should be enough to ensure consistent results

### DIFF
--- a/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
+++ b/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
@@ -87,7 +87,7 @@
                         title: item.stationName,
                         availableDocks: item.availableDocks,
                         availableBikes: item.availableBikes,
-                        lastCommunication: moment(new Date(item.lastCommunicationTime)).fromNow()
+                        lastCommunication: moment(new Date(item.lastCommunicationTime.replace(/\-/g,'/') + " EDT")).fromNow()
                     };
                 }
             });


marianosimone:
@moollaza @xuv I don't see any problems with parsing the `-`separators in the date

xuv:
@marianosimone There is a problem in Firefox and Chrome, at least. 
Ex in Firefox:
```
new Date('2015-09-07 02:20:53 PM')
Invalid Date
new Date('2015/09/07 02:20:53 PM')
Date 2015-09-07T18:20:53.000Z
```

And should we add Momentjs Timezone lib? http://momentjs.com/timezone/docs/

marianosimone:
@xuv wired... I'm running Chrome 45.0.2454.85 (64-bit) on MacOS and get:

    > new Date('2015-09-07 02:20:53 PM')
    < Mon Sep 07 2015 14:20:53 GMT-0400 (EDT)

Anyway, if there IS a problem with other browsers/OS, I'm adding the fix.

As per Moment Timezone, I'm not sure if we need it. @moollaza?

marianosimone:
@moollaza @xuv, oh... shoot. Actually, appending EDT won't work, as when we go back to EST it will be off by an hour.

Let me try again with the people at CitiBike to see if I can get a better solution for this

xuv:
@marianosimone My bad. You're right. `new Date('2015-09-07 02:20:53 PM')` works in Chrome. But not in Firefox 40.0.3 on Linux.

And for the EDT/EST thing, that's why I was suggesting Moment Timezone. Maybe that would help solve this. Not sure.

moollaza:
@marianosimone if CitiBike can adjust their date stamp to be in UTC, or at least specify the timezone offset that would be great. I'd like to avoid adding additional overhead if possible, but worst case I guess we'll have to use MomentTimezone

marianosimone:
@moollaza agreed. I'm in contact with them now, and expect to have an answer pretty soon. I'll keep you guys updated

moollaza:
@marianosimone any updates on this?

marianosimone:
Unfortunately, I haven't heard back from the people behind the feeds yet... I'm trying again now! Will keep you posted

marianosimone:
This is not a good solution, and I'm following up with the people from CitiBIke for a better one. Will follow up in the referenced bug report

PR #2164, opened 2015-09-09
pass full query_lc to local

b2ddg:
Changing handle remainder to handle query_lc so the full query is passed to local

@bsstoner @nilnilnil @tommytommytommy


tommytommytommy:
Change looks okay to me, but the tests need to be updated:
https://travis-ci.org/duckduckgo/zeroclickinfo-spice/builds/79480925#L2843

tommytommytommy:
Actually, now that we're handling the entire query instead of the remainder, the false positives that we were stopping will get through unless we modify the check, not the test.

tommytommytommy:
:+1:

PR #2167, opened 2015-09-10
Increase maps relevancy cutoff

b2ddg:
Changing the relevancy threshold to a stricter value of 0.9

@tommytommytommy @nilnilnil


tommytommytommy:
:+1:

PR #2171, opened 2015-09-11
Add vendor prefixes to the transform CSS property

jagtalon:
I added 

```css
-webkit-transform: translate(-50%, 0);
-ms-transform: translate(-50%, 0);
```

So that this IA will work on IE 9 and Safari


jagtalon:
For @moollaza

moollaza:
LGTM :+1:

PR #2172, opened 2015-09-11
currency - added more names to thb and bdt

iambibhas:
Today I was trying to convert some baht to rupees and got stumped. So added baht to the list. And also added taka to Bangladesh Taka, as it's more commonly called only taka.


moollaza:
LGTM :+1:

PR #2173, opened 2015-09-11
Moving chains_re to places triggering

b2ddg:
Moving the chains regex to places to unify IA triggering. Matches internal 4898

@tommytommytommy @nilnilnil


tommytommytommy:
:+1:

PR #2176, opened 2015-09-13
first implementation of bike sharing for the Bay Area (CA)

marianosimone:
After NYC and PHI, now comes the Bay Area!

bike_share_bay_area

https://duck.co/ia/view/bay_area_bike_share_ca

### DIFF
--- a/lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm
+++ b/lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm
@@ -1,0 +1,39 @@
+package DDG::Spice::BikeSharing::BayAreaBikeShareCA;
+
+use strict;
+use DDG::Spice;
+
+name 'Bay Area Bike ShareCA';
+source 'Bay Area Bike Share CA';
+icon_url 'http://www.bayareabikeshare.com/assets/images/bayarea/icons/favicon.ico';
+description 'Search Bay Area Bike Share stations in the Bay Area, CA';
+primary_example_queries 'bike share bay area', 'bike share san francisco';
+secondary_example_queries 'bike share stations near san jose';
+category 'geography';
+topics 'everyday', 'travel', 'entertainment', 'geography';
+code_url 'https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm';
+attribution github => 'marianosimone',
+            web  => ['http://www.marianosimone.com',  'Mariano Simone'];
+
+triggers any => 'bike sharing', 'bike share', 'bikeshare';
+
+spice to => 'http://www.bayareabikeshare.com/stations/json';
+spice wrap_jsonp_callback => 1;
+spice is_cached => 0;
+spice proxy_cache_valid => '200 304 15m';
+
+my @places = ("bay area", "san francisco", "sfo", "sf", "san jose", "mountain view", "palo alto", "redwood", "redwood city");
+my $place_re = join "|", @places;
+
+handle remainder => sub {
+    my $place = $_ || undef;
+
+    unless ($place) {
+        $place = $loc->city if $loc && $loc->city;
+    }
+
+    return unless $place && $place =~ m/\b($place_re)\b/;
+    return $1;
+};
+
+1;
--- a/share/spice/bike_sharing/bay_area_bike_share_ca/bike_sharing_bay_area_bike_share_ca.js
+++ b/share/spice/bike_sharing/bay_area_bike_share_ca/bike_sharing_bay_area_bike_share_ca.js
@@ -1,0 +1,114 @@
+(function(env) {
+    "use strict";
+
+    var references = {
+        'san francisco': {
+            lat: 37.7577,
+            lon: -122.4376
+        },
+        'redwood city': {
+            lat: 37.5081359,
+            lon: -122.2139269
+        },
+        'palo alto': {
+            lat: 37.42565,
+            lon: -122.13535
+        },
+        'mountain view': {
+            lat: 37.4133235,
+            lon: -122.081267
+        },
+        'san jose': {
+            lat: 37.2970155,
+            lon: -121.817413
+        }
+    };
+
+    references['sf'] = references['sfo'] = references['san francisco'];
+    references['redwood'] = references['redwood city'];
+
+    env.ddg_spice_bike_sharing_bay_area_bike_share_ca = function(api_result) {
+        if (!api_result || !api_result.stationBeanList) {
+            return Spice.failed('bay_area_bike_share_ca');
+        }
+
+        var script = $('[src*="/js/spice/bike_sharing/bay_area_bike_share_ca/"]')[0],
+            source = $(script).attr("src"),
+            query = source.match(/bay_area_bike_share_ca\/([^\/]+)/)[1];
+
+        if (query) {
+            query = decodeURIComponent(query);
+        }
+
+        DDG.require('moment.js', function() {
+        DDG.require('maps', function() {
+            moment.locale('en', {
+                relativeTime : {
+                    past: "%s ago",
+                    s:    "1 sec",
+                    ss:   "%d sec",
+                    m:    "1 min",
+                    mm:   "%d min",
+                    h:    "1 hour",
+                    hh:   "%d hours",
+                    d:    "1 day",
+                    dd:   "%d days",
+                    M:    "1 month",
+                    MM:   "%d months",
+                    y:    "1 year",
+                    yy:   "%d years"
+                }
+            });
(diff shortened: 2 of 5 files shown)


share/spice/bike_sharing/bay_area_bike_share_ca/bike_sharing_bay_area_bike_share_ca.js
```diff
@@ -0,0 +1,111 @@
...
+                },
+                sort_fields: {
+                    distance: function(a, b) {
+                        return a.distanceToReference - b.distanceToReference;
+                    }
+                },
+                sort_default: 'distance',
+                data: api_result.stationBeanList,
+                normalize: function(item) {
+                    if (item.testStation || item.statusKey != 1) {
```
marianosimone: another interesting feature of this resultset is that it has a `city` property, that could be used to filter out by "san francisco", "redwood city", "palo alto", etc. I'm not sure if we want to do that and, if so, how to handle synonyms (maybe with a dictionary? If so, in js or in perl, before sending the data here?

jagtalon:
@marianosimone awesome! Checking it out. :)

share/spice/bike_sharing/bay_area_bike_share_ca/bike_sharing_bay_area_bike_share_ca.js
```diff
@@ -0,0 +1,111 @@
...
+                    MM:   "%d months",
+                    y:    "1 year",
+                    yy:   "%d years"
+                }
+            });
+            Spice.add({
+                id: 'bay_area_bike_share_ca',
+                name: 'Bike Sharing',
+                meta: {
+                    sourceName: 'Bay Area Bike Share CA',
```
jagtalon: We probably don't need the "CA" at the end of the source name since:

- I don't think people will be confused with which Bay Area we're referring to since they either made it specific in the query / they're in the bay area.

- The source doesn't have "CA" at the end http://www.bayareabikeshare.com/stations

share/spice/bike_sharing/bay_area_bike_share_ca/bike_sharing_bay_area_bike_share_ca.js
jagtalon: I think we'd like to do that and we're currently working on the filtering functionality. It's not out yet, but it would be a good addition in the future.

jagtalon:
@marianosimone I think we need to reduce the height of the titles a little since a part of the text in the third line is being shown:

screen shot 2015-09-14 at 12 14 38 pm

screen shot 2015-09-14 at 12 14 48 pm

lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm
```diff
@@ -0,0 +1,39 @@
...
+spice proxy_cache_valid => '200 304 15m';
+
+my @places = ("bay area", "san francisco", "sfo", "sf", "san jose", "mountain view", "palo alto", "redwood", "redwood city");
+my $place_re = join "|", @places;
+
+handle remainder => sub {
+    my $place = $_ || undef;
+
+    unless ($place) {
+        $place = $loc->city if $loc && $loc->city;
```
jagtalon: It's possible for `$loc` to be `undef` so check that as well before accessing `$loc->city`.

lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    return unless $place && $place =~ m/\b($place_re)\b/;
```
jagtalon: This `unless` is a bit hard to read. Can we flip it to `return if ...` instead?

lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm
marianosimone: mhhh.. i was just copying from @moollaza's PR: https://github.com/duckduckgo/zeroclickinfo-spice/pull/2166

marianosimone:
@moollaza @AcriCAA do you know what's going on with that 3rd line in the title? It doesn't seem to be a problem for NYC and PHI. Any ideas?

lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm
moollaza: We are checking `$loc` before `$loc->city` :)

AcriCAA:
@marianosimone I have been looking and can't find anything in the code causing it.  I have been trying to research whether the parentheticals in the name could perhaps be throwing off the sizing? I need to clone this project and mess around a bit when I get a chance.

lib/DDG/Spice/BikeSharing/BayAreaBikeShareCA.pm
jagtalon: @moollaza bah I'm blind. Thanks!

jagtalon:
IA Page for this: https://duck.co/ia/view/bay_area_bike_share_ca

share/spice/bike_sharing/bay_area_bike_share_ca/bike_sharing_bay_area_bike_share_ca.js
marianosimone: Agreed, changed ;)

marianosimone:
@jagtalon can you confirm the third line problem? (it seems like you are running Chrome in a Mac, right?)

I'm not seeing the issue on Firefox in a Linux box, so it *might* be browser related.

line

cc @moollaza @AcriCAA

moollaza:
@marianosimone @AcriCAA I'm deploying this to beta.duckduckgo.com. Should up soon. I'll take a look...

AcriCAA:
@moollaza sweet!  You knocked another out of the park @marianosimone!

jagtalon:
@marianosimone sorry for the late reply. Checking it out to see what's up

jagtalon:
@marianosimone It definitely shows up on Webkit on Windows and Mac (Chrome and Safari is what I tested it on) but doesn't show up on Firefox. This also happens for NYC and Philly bike maps.

Adjusting the height from 2.5em to 2.4em fixes it.

jagtalon:
@marianosimone since this doesn't have any footer like NYC and Philly Bike Share, you'd have to adjust the tile view to something like this:

bike sharing - v3e no updates

marianosimone:
@jagtalon: what about this?
three-lines

using `3.5em` allows three lines to be there, and still leaves some padding below the numbers. If you like it, we could easily port to PHI

moollaza:
@marianosimone if this is the same as PHL in terms of implementation then I'm happy to merge this. We can deal with the 3-line titles by switching this to the Text template (which will also remove the extra white-space at the bottom) as I've done in #2187

marianosimone:
@moollaza I applied the same changes you did in #2187 (and the `tileTitle` property as well).

@jagtalon can you confirm there's no problem with an extra line showing up? (I'm testing on a Mac now, and it looks good to me)

moollaza:
@marianosimone thanks this is looking great. I see that there's a small "bug" in the template that the 3-lines causes which is that the bottom padding on the title is actually being ignored because the footer is absolutely positioned. We'll need to fix this ourselves though.

However @marianosimone, one weird thing is that it seems we're getting results for all of California, instead of just the bay area. Can you fix that?

After that, this should be good!

moollaza:
bike_sharing_san_francisco_at_duckduckgo

marianosimone:
@moollaza all those cities are, indeed, a part of the [Bay Area](https://en.wikipedia.org/wiki/San_Francisco_Bay_Area), and they are all part of the same bike sharing system.

I applied the styling changes to match your PR for Philly

moollaza:
>all those cities are, indeed, a part of the Bay Area, and they are all part of the same bike sharing system.

I didn't realize the Bay Area included all of that, but what I actually meant to point out as the bug, is that for my query in the above picture, I specified "sf" and we're showing results for all of the Bay Area, which isn't good.

Can we ensure we only show results for the specified region?

marianosimone:
@moollaza see comment here: https://github.com/duckduckgo/zeroclickinfo-spice/pull/2176#discussion_r39348666

(BTW: the same happens with the CitiBike NYC IA, but that one doesn't return the data that might help us filter out)

jagtalon:
@marianosimone @moollaza tested this all the way down to IE 9 and it looks good

jagtalon:
:+1:

moollaza:
@marianosimone congrats! It's live: https://duckduckgo.com/?q=bike+share+bay+area&t=ffcm&ia=bikesharing

PR #2182, opened 2015-09-15
Twitter: Fixes #2158

### LINKED ISSUE #2158: Twitter: Title link should point to result's Twitter profile
We need to define a `url` property in the `normalize()` function that should be identical to the `sourceURL` we create. That way clicking the title link will take you to the Twitter profile for the result we're showing, instead of linking back to the current page!

------
IA Page:  http://duck.co/ia/view/twitter

samskeller:
The title link now points to the result's Twitter page, as described in the issue.

screen shot 2015-09-14 at 9 00 52 pm

### DIFF
--- a/share/spice/twitter/twitter.js
+++ b/share/spice/twitter/twitter.js
@@ -45,6 +45,7 @@
                     title: item.name,
                     subtitle: subtitle,
                     altSubtitle: getURL(item),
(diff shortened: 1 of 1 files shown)


jagtalon:
@samskeller NICE! :+1:

PR #2188, opened 2015-09-18
Set browser location attribute in places.pm

b2ddg:
Allow local to use lat,lon coordinate data 

Needs internal PR:4940


b2ddg:
I think this needs an updated DDG/Meta/ZeroClickInfoSpice.pm to pass the build tests.

PR #2189, opened 2015-09-18
Spice wgha: Reduce cache time and add user location to query

bigcurl:
Hi,

This pull request reduces the cache time from 60m to 5m to deliver fresher results for the spice wgha. 

Before the spice would trigger for every request from the ''DE" region. Now only fires when there is a useful city in the query including the users location. This saves some cpu cycles as the same match is done in the API. Tests have been updated accordingly.

Regards
Samuel


lib/DDG/Spice/WGHA.pm
```diff
@@ -20,13 +20,12 @@ my @triggers = ("was geht", "wasgeht", "aus gehen", "ausgehen", "nachtleben", "p
...
+spice proxy_cache_valid => "200 5m";
 
 handle query_lc => sub {
-  return $_ if $loc->country_code eq 'DE';
-
-  if ($_ =~ /aschaffenburg|augsburg|berlin|bremen|darmstadt|dortmund|dresden|düsseldorf|duesseldorf|essen|frankfurt|ffm|freiburg|hamburg|hannover|heidelberg|kassel|köln|koeln|leipzig|mainz|mannheim|münchen|muenchen|münster|muenster|nürnberg|nuernberg|stuttgart|wiesbaden/) {
-    return $_;
+  $a = $_ . ' ' . $loc->city;
+  if ($a =~ /aschaffenburg|augsburg|berlin|berlin-biesdorf|berlin-friedrichshain|berlin-karow|berlin-köpenick|berlin-lichtenberg|berlin-marzahn|berlin-mitte|berlin-oberschöneweide|berlin-pankow|berlin-treptow|bremen|cologne|darmstadt|dieburg|dortmund|dreieich|dresden|dresden-altstadt|dresden-neustadt|düsseldorf|essen|ffm|frankfurt|frankfurt am main|freiburg|freiburg im breisgau|friedrichshain|griesheim|halle|hamburg|hamburg-eimsbüttel|hamburg-mitte|hannover|hanover|heidelberg|kassel|koeln|germany|kreuzberg|köln|leipzig|leipzig-connewitz|leipzig-leutzsch|leipzig-plagwitz|leipzig-stötteritz|mainz|mannheim|munich|mühlheim|münchen|münchen|germany|münster|neu isenburg|nuremberg|nürnberg|offenbach|prenzlauer berg|sankt pauli|sersheim|stuttgart|wiesbaden/) {
```
moollaza: In the interest of saving more CPU cycles :), I think we should put this list into a JSON/YAML file so we can load it into a hash and do a hash lookup instead of using a regex. Alternatively, we could build the regex outside the `handle` function so we don't need to re-build the hash each time the function is executed?

bigcurl:
@moollaza Can you please point me to a spice that uses the json/yaml file mechanism?

moollaza:
> @moollaza Can you please point me to a spice that uses the json/yaml file mechanism?

Sure! One example is the SunRiseSet Spice: https://github.com/duckduckgo/zeroclickinfo-spice/blob/cd4488ce605e1c32f967c268175d86cfdff09595/lib/DDG/Spice/SunRiseSet.pm#L24

But there are several others: https://github.com/duckduckgo/zeroclickinfo-spice/search?utf8=%E2%9C%93&q=loadFile

bigcurl:
@moollaza I changed the direction for this pull request and updated the title accordingly. For now lets just reduce the cache size and add the users location to a query. This helps sorting out querys like "was geht heute" which would become "was geht heute frankfurt" if the user is from Frankfurt. Let me know what you think.

moollaza:
@bigcurl my concern with these changes is that now it will trigger for everyone else in the world, where it's far less likely to be relevant to the user.

For example, North American users' who search "party" would now trigger this IA which tells other IAs to wait for it (because we expect to provide a result) but in these cases (users outside Germany), I think we're pretty sure that we're not going to return a relevant result, so we'd be waiting for nothing.

Does that make sense? I think the previous direction was better.

moollaza:
@bigcurl I just tested this locally and in DuckPAN my browser is telling me that the result is malformed (I think because it's empty). I think you need to ensure you're returning valid JSON for queries that have no results

moollaza:
This is the query I used: http://ddg.wasgehtheuteab.de/duckduckgo/events.json?q=parties%20Phoenixville&callback=ddg_spice_wgha&api_key=<api_key>

bigcurl:
ok, I will provide a new pull request which will take the other direction again.

moollaza:
@bigcurl apologies, I didn't mean to close the PR!

moollaza:
@bigcurl Ping! I'm just checking in to see if you're still interested in finishing this one up? We'd love to see this go live!

I think we still need to find a good balance with the triggering. In it's current state this will simply trigger far too often for irrelevant queries.The old approach, triggering for all queries that come from Germany, or queries with a German city in it makes the most sense. Otherwise anyone in the world who searches for a party will be making needless API calls.

bigcurl:
Pong. Still on it but will take some time.

bigcurl:
Hi @moollaza,

changed everything back: 
* will only trigger for request from DE
* cities regex is created outside of handler

Please have a look at the changes so far and I wish you a merry christmas

moollaza:
@bigcurl apologies for the delays here -- it looks like the branch has conflicts now, can you resolve them so I can merge this in?

bigcurl:
@moollaza Fixed the conflict.

moollaza:
Thanks! This LGTM :+1:

PR #2194, opened 2015-09-21
Quixey fix & cleanup

moollaza:
Update Quixey to get it back online
- Needed to fix url handling for images (no more parsing magic)
- Also some code cleanup/re-org

### DIFF
--- a/share/spice/quixey/quixey.js
+++ b/share/spice/quixey/quixey.js
@@ -1,5 +1,5 @@
 (function(env) {
-    var quixey_image_domain = "d1z22zla46lb9g.cloudfront.net";
+    "use strict";
 
     // spice callback function
     env.ddg_spice_quixey = function(api_result) {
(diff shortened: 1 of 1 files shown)


share/spice/quixey/quixey.js
```diff
@@ -244,50 +237,29 @@
...
-        // and it redirects to HTTPS. What we want is an HTTP link (for speed).
-        if (item.platforms && item.platforms.length > 0 && item.platforms[0].icon_url) {
-            domain = item.platforms[0].icon_url.match(/https?:\/\/([^\/]+)/)[1];
+            if (item.editions){
+                $.each(item.editions, function(index, edition) {
+                    if (edition.icon_url){
+                        icon_url = item.editions.icon_url;
+                        return false
+                    }
+                });
```
bsstoner: FYI--you can use `item.editions.forEach(function(edition, index){ ... }` now that we dropped IE8 support :)

bsstoner:
:+1: Tested and works!

PR #2195, opened 2015-09-22
Bradley/require async

bbraithwaite:
Call the DDG.require function with updated API to call multiple/async. NB there is a dependency on internal code.

### DIFF
--- a/share/spice/sports/mlb/games/sports_mlb_games.js
+++ b/share/spice/sports/mlb/games/sports_mlb_games.js
@@ -15,7 +15,7 @@
             return Spice.failed('mlb_games');
         }
         
-        DDG.require('sports', function(){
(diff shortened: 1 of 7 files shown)


share/spice/sports/mlb/games/sports_mlb_games.js
```diff
@@ -15,7 +15,7 @@
             return Spice.failed('mlb_games');
         }
         
-        DDG.require('sports', function(){
+        DDG.require(['moment.js', 'sports'], function(){
```
bbraithwaite: The call to Games.init(MLBInit); in this function callback also requires moment.js, so including that here to speed things up.

moollaza:
This is awesome :+1:

bsstoner:
:+1:

PR #2198, opened 2015-09-23
Coderwall: remove image and switch to text template

moollaza:
@abeyang I removed the image, but the icon template is obviously meant to use an icon so there was padding on the title/subtitles.

I assume we should be using the text template instead? Here's the result:

coderwall_jpcamara_at_duckduckgo

Here's the icon template, without the image: 

coderwall_jpcamara_at_duckduckgo

**Update**: Just realized the `text_detail` template doesn't support an `altSubtitle` I guess we'll have to use the `Icon` template and some CSS. Alternatively, we could add an altSubtitle to the text template?

### DIFF
--- a/share/spice/coderwall/coderwall.js
+++ b/share/spice/coderwall/coderwall.js
@@ -58,7 +58,7 @@
             },
 
             templates: {
-                group: 'text'
+                group: 'icon'
             }
         });
     };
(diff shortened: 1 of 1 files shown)


jagtalon:
Setting to internal because @sdougbrown is working on a fix for the icon template.

moollaza:
Merging. Internal fix to template padding is merged and will be released:

coderwall_jpcamara_at_duckduckgo

PR #2199, opened 2015-09-23
Currency Tiles - Fix Border

sdougbrown:
Uses the pre-built `disc` class to handle the styling for the flags in these currency tiles.  They now have an inset border. 

to @jagtalon 
cc @abeyang @moollaza @chrismorast 

---

IA Page: https://duck.co/ia/view/currency


jagtalon:
@sdougbrown nice!

Issue #2068, opened 2015-08-04
Airlines: "Showing 2 Flight" Typo


hawkrives:
When searching for a flight that happens once a day, at least (for instance, "DL 580"), the header area that describes how many results were found says "Showing 2 Flight"—that is, it's not pluralized properly.

I've spelunked around this repo a bit, but I'm not sure where the stuff for that header is located. I have confirmed that it does not happen with the Images tab, so it seems to be type-specific, but I can't seem to find where that's defined.

Anyway, I *love* DuckDuckGo, and being able to file issues like this is amazing!

------
IA Page:  http://duck.co/ia/view/airlines

moollaza:
@hawkrives thanks a lot for reporting this!

The issue comes from here: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/airlines/airlines.js#L275

We're using a new approach to gather flights from multiple API calls, and so the check is actually true because the API call we're considering only had 1 result. We're not taking into account the additional flight we grabbed in another call. We need to check the length of our finalized array to make sure it's correct.

Also we have a fancy new pluralization helper fn, that can be used here: `DDG.pluralize(num, 'single form', "plural form")`. When there's no plural form, we return `"single form" + "s"` when necessary.

yash14123:
Is this fixed or still open?

tagawa:
Thanks for the ping @theaverageguy. Looks like the wording on this has changed so this is no longer an issue: https://beta.duckduckgo.com/?q=ua78&iai=1&ia=flights

PR #2202, opened 2015-09-26
Flight(s) typo

rmad17:
I received the duckduckhack weekly update
where this was listed under 5 minutes fixes. It said, It should say "Flights" when we present multiple results. Otherwise, just say "Flight".
Submitting the patch for the same.

Fixes https://github.com/duckduckgo/zeroclickinfo-spice/issues/2068

### DIFF
--- a/share/spice/airlines/airlines.js
+++ b/share/spice/airlines/airlines.js
@@ -272,7 +272,7 @@
                 sourceUrl: "http://www.flightstats.com/go/FlightStatus/flightStatusByFlight.do?"
                     + "airlineCode=" + results[0].airlineCode 
                     + "&flightNumber=" + results[0].flightNumber,
-                itemType: results.length === 1 ? "flight" : "flights by departure time" 
+                itemType: results.length === 1 ? "Flight" : "Flights" 
             },
             normalize: function(item) {
 


moollaza:
@rmad17 thanks a lot for trying to fix this! We really appreciate it :+1: 

Unfortunately though, this doesn't actually solve the problem. Please take a look at #2068 and read the comment I made explaining the root of the problem. Some extra logic needs to be implemented to account for things. You've only adjusted the casing of the words, but it will still say `"Showing 2 Flight"`.

aa102_at_duckduckgo_and_1____projects_zeroclickinfo-spice__bash_

rmad17:
@moollaza  Thanks for clarifying. It was my mistake I didn't actually search and realize the bug. If I find some time I'll try to fix this by tomorrow. I'm a newbie at contributing so please bear with my mistakes. :)

rmad17:
@moollaza I did try to take a look around the repo to find out where the logic would be, I was not successful.  
However, when I actually searched the plurarized form did come out correctly.
selection_002

moollaza:
@rmad17 I believe you'll need to define a variable inside the closure that you modify inside the `display` function.

We're actually calling the `ddg_spice_flights_route` function twice, because we add another script tag to the page, which makes an additional API call and calls the callback function separately.

So, we need to give the callback function access to a private variable that can be read and updated each time the callback is called so that it can keep a total count of the results we're showing and then display "Flight" or "Flights" accordingly.

moollaza:
If you'd like to team up with someone on figuring this out, I'd suggest your join our Slack channel and ask for someone's help on there.

If you'd like to our Slack channel please send an email to quackslack@duckduckgo.com.

moollaza:
I think we may have fixed this one with some recent updated. This branch is behind anyways so I'm closing this. If you'd like to re-investigate it you're welcome to. I can re-open this PR if needed

PR #2203, opened 2015-09-28
Fix movies lo-res fallback image so we don't have empty tiles

bsstoner:
Using the lo-res image as a fallback for movies tiles so we don't have empty white tiles. Not ideal, but imo better than a broken looking tile.

What it currently looks like in production:
screen shot 2015-09-28 at 6 32 59 am

With stretched image (tiles are in slightly different order on my dev box cause of caching):
screen shot 2015-09-28 at 9 39 15 am

@jagtalon 
cc @andrey-p @moollaza

### DIFF
--- a/share/spice/in_theaters/in_theaters.js
+++ b/share/spice/in_theaters/in_theaters.js
@@ -1,20 +1,6 @@
 (function(env) {
     "use strict";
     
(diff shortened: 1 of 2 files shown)


jagtalon:
@bsstoner thank you! Checking

jagtalon:
@bsstoner works well! :+1:

PR #2204, opened 2015-09-28
Cryptocurrency error handling

bsstoner:
I can't reproduce this anymore, but a few weeks ago we were seeing cases that this error was being thrown:
```
TypeError: rows[0] is undefined
```

I'm assuming it was a case where `rows` was just an empty array. But in these cases where it threw the error it was not calling Spice.failed(), which would cause a bad fallback to images. I don't think there's any harm in adding these additional checks, but please let me know if you think we can't do this.


jagtalon:
@bsstoner thank you! Checking

jagtalon:
@bsstoner works well! :+1:

PR #2206, opened 2015-09-28
Fix octopart console errors

bsstoner:
screen shot 2015-09-28 at 1 12 04 pm

@jagtalon
cc @moollaza

### DIFF
--- a/share/spice/octopart/buy.handlebars
+++ b/share/spice/octopart/buy.handlebars
@@ -1,4 +1,6 @@
 <div class="detail__callout--pr">
-	<a href="{{{datasheet}}}" class="btn btn__datasheet">Datasheet</a>
+    {{#if datasheet}}
+	    <a href="{{{datasheet}}}" class="btn btn__datasheet">Datasheet</a>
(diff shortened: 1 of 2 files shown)


jagtalon:
@bsstoner checking!

jagtalon:
@bsstoner it's a bit weird that the datasheet button just links to the current page because there's no URL in the a tag. 

screen shot 2015-09-28 at 2 27 49 pm

What if we link to https://octopart.com/contact/send_file?ppid=24675717&type=datasheet&continue_to=/721007-carlisle-24675717 instead and the button saying "Send a Datasheet"? This might be a design question though, but would love to hear what you think!

bsstoner:
Why if we add a conditional to the template and hide the button? https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/octopart/buy.handlebars

jagtalon:
@bsstoner that might be the best solution so go for it! :)

jagtalon:
@bsstoner nice! :+1:

Issue #2218, opened 2015-10-08
Octopart: JS error


bsstoner:
Seeing this error get thrown periodically, need to guard against it:
```
TypeError: null is not an object (evaluating 'item.avg_price_v2[0].toFixed')
```

------
IA Page:  http://duck.co/ia/view/octopart

bsstoner:
Example: https://duckduckgo.com/?q=specs+for+torque+c146

![screen shot 2015-10-08 at 4 53 08 pm](https://cloud.githubusercontent.com/assets/126358/10379589/156c4cd2-6ddd-11e5-8c3a-c595c1b62090.png)

SibuStephen:
Please review `#2226`

PR #2226, opened 2015-10-13
A Javascript bug because of Octopart: JS Error

SibuStephen:
Removed `toFixed()` function  as the use for that function was inappropriate for that particular section, along with that Initialized a variable with zero, to avoid future Null exceptions.

### DIFF
--- a/share/spice/octopart/octopart.js
+++ b/share/spice/octopart/octopart.js
@@ -43,24 +43,25 @@
                 sourceName: 'Octopart',
                 sourceUrl : 'http://octopart.com/search?q=' + api_result.request.q
             },
+         
             normalize: function(item) {
+                var items = 0;
+                items = item.item; // our item object is wrapped in an "item" property
 
-                item = item.item; // our item object is wrapped in an "item" property
-
-                var images = get_images(item.imagesets),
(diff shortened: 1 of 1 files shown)


daxtheduck:
@SibuStephen Hey!
        
Thanks!

moollaza:
@SibuStephen thanks for submitting this. Please be sure to mention something like "fixes #2218" in your description so we know what this is about. Also, it automatically closes the issue when this is merged :+1: 

Regarding your fix, I don't see why `item` was renamed to `items` or why it needs to be initialized to `0`?

It seems like the problem is that we're not checking if `item.avg_prive_v2` is actually defined?

We can more easily fix this bug with a solution like this:

```js
var item = item.item;
if (!tem.avg_price_v2[0]){
    return null;
}
```

SibuStephen:
@moollaza  i will check with it and soon will submit a PR for this.

moollaza:
@SibuStephen I see you already submitted a new PR. FWIW, there's really no need, you can just add more commits to your branch and push them here. If you really want you can `git squash` you commits so all the changes are reduced to a single commit. We don't really care about the commit log being messy though, as long as the final result is good.

PR #2249, opened 2015-10-27
(Sports) Remove special boxscore re-sorting for mobile

sdougbrown:
This removes the out-of-order game that was being inserted in mobile portrait mode.

to @bsstoner 
cc @b1ake @chrismorast


bsstoner:
:+1:

PR #2254, opened 2015-10-31
Fixing Issue #2253

### LINKED ISSUE #2253: Random Words: Title + Template should be different for single word results
- The title reads `Random Words` instead of `Random Word`
- We don't need to use the list template for a single word. The Text template should work

------
IA Page:  http://duck.co/ia/view/rand_word

codenirvana:
I hope this will fix #2253

---
IA Page: http://duck.co/ia/view/rand_word

### DIFF
--- a/share/spice/rand_word/rand_word.js
+++ b/share/spice/rand_word/rand_word.js
@@ -12,45 +12,39 @@
             list_of_list.push(api_result.splice(0, 4));
         }
 
+        var spiceObj =  {
+            id: "rand_word",
+            data: {
+                title: 'Random Words',
+                list: list_of_list
+            },
+            name: "Answer",
+            meta: {
+                sourceUrl: 'https://wordnik.com/randoml',
+                sourceName: 'Wordnik',
+                sourceIcon: true
+            },
+            templates: {
+                group: 'list',
+                options: {
+                    list_content: Spice.rand_word.content,
+                    moreAt: true
+                }
+            }
+        };
+        
         if ( list_of_list.length == 1 ) {
-            Spice.add({
-                id: "rand_word",
-                data: list_of_list[0],
-                name: "Answer",
-                meta: {
-                    sourceUrl: 'https://wordnik.com/randoml',
-                    sourceName: 'Wordnik',
-                    sourceIcon: true
-                },
-                templates: {
-                    group: 'text',
-                    options: {
-                        content: Spice.rand_word.single_word,
-                        moreAt: true
-                    }
+            spiceObj.data = list_of_list[0];
+            spiceObj.templates = {
+                group: 'text',
+                options: {
+                    content: Spice.rand_word.single_word,
+                    moreAt: true
                 }
-            });
-        } else {
-            Spice.add({
-                id: "rand_word",
-                data: {
-                    title: 'Random Words',
-                    list: list_of_list
-                },
-                name: "Answer",
-                meta: {
-                    sourceUrl: 'https://wordnik.com/randoml',
-                    sourceName: 'Wordnik',
-                    sourceIcon: true
-                },
-                templates: {
-                    group: 'list',
-                    options: {
-                        list_content: Spice.rand_word.content,
-                        moreAt: true
-                    }
-                }
-            });
+            };
         }
+        
+        Spice.add( spiceObj );
+        
     }
 }(this));
--- a/share/spice/rand_word/rand_word.js
+++ b/share/spice/rand_word/rand_word.js
@@ -11,11 +11,13 @@
         while (api_result.length > 0) {
             list_of_list.push(api_result.splice(0, 4));
         }
-
+        
+        var title = (api_result.length === 1 ? 'Random Word' : 'Random Words');
+        
         Spice.add({
             id: "rand_word",
             data: {
-                title: 'Random Words',
+                title: title,
                 list: list_of_list
             },
             name: "Answer",
--- a/share/spice/rand_word/rand_word.js
+++ b/share/spice/rand_word/rand_word.js
@@ -35,7 +35,7 @@
             }
         };
         
-        if ( list_of_list.length == 1 ) {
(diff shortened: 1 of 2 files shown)


MrChrisW:
@codenirvana Thanks for the fix. There's two components to issue #2253

>The title reads Random Words instead of Random Word
> We don't need to use the list template for a single word. The Text template should work

You've fixed the Random Word(s) one but not switched the [template to Text](https://github.com/codenirvana/zeroclickinfo-spice/blob/random-word-fix/share/spice/rand_word/rand_word.js#L29-L34)

codenirvana:
@MrChrisW switching to text template for one word now.
Not sure its the right way or not, I am new to spice!
Also when I checked using ```duckpan``` it show this error:
```
API key not found. Using DuckDuckGo's endpoint:
"https://ddh10.duckduckgo.com/js/spice/rand_word/1-100-1"
```

share/spice/rand_word/rand_word.js
```diff
@@ -11,11 +11,19 @@
...
             list_of_list.push(api_result.splice(0, 4));
         }
-
+        
+        var title = 'Random Words';
+        var group = 'list';
+        
+        if ( api_result.length === 1 ) {
+            title = 'Random Word';
+            group ='text';
```
moollaza: I was wrong before, it'll be easier if you keep with the `list` group, but for the single would you need to specify `content: Spice.rand_word.single_word` (for example) Where `single_word.handlebars` is a new sub-template that create a link for a single word.

You'll need to play around with it to get it working. I think this will require you to get your own Wordnik API key

codenirvana:
@moollaza Requested for Wordnik API key 2 days ago.... still waiting for their response.

codenirvana:
@moollaza Problem with title is fixed but don't know how to adjust the template for single word.
Tried some tweaks but nothing seems to work.

share/spice/rand_word/rand_word.js
```diff
@@ -15,7 +15,7 @@
         Spice.add({
             id: "rand_word",
             data: {
-                title: 'Random Words',
+                title: ( list_of_list.length ==1 ) ? 'Random Word' : 'Random Words',
```
moollaza: I just remembered we have `DDG.pluralize()`. Let's use it here:

```
title: DDG.pluralize(list_of_lists.length, "Random Word"),
```

moollaza:
>Problem with title is fixed but don't know how to adjust the template for single word.
Tried some tweaks but nothing seems to work.

@codenirvana what did you try? For the purpose of seeing it work, I would suggest you use an `if` statement that calls `Spice.add()`. This way you'll be able to ensure it works in each case. Then from there you can simplify the code to only use 1 `Spice.add()` call and modify the object you pass to the call as needed.

E.g.

```js

if (isSingleWord) {
    Spice.add({
        // object with group: "text"
    });
} else {
    Spice.add({
        // object with group: "list"
    });
}
```

share/spice/rand_word/rand_word.js
codenirvana: @moollaza ```DDG.pluralize()``` method is not working... do I need to import anything?
```
title: DDG.pluralize(list_of_lists.length, "Random Word","Random Words"),
```

codenirvana:
@moollaza have a look. It works now.
Any other way to do the same more effectively?

moollaza:
>  It works now.

Nice!

> Any other way to do the same more effectively?

@codenirvana absolutely. You can define your object outside of the `Spice.call()` and modify it as needed.

So I would move the `Spice.add()` outside and after the `if` and conditionally define properties of your object:

```js
var spiceOb = {
    id: 'rand_word',
    name: 'Answer',
    meta: { ... },
    templates: {  //assume multiple words
    }
}

// modify what's needed
if (isSingleWord) {
    spiceObj.templates = {
        group: ...,
        ...
    }
}

Spice.add( spiceObj ); // only call once
```

codenirvana:
@moollaza defining object outside works :+1: 
Code looks much cleaner now.

share/spice/rand_word/single_word.handlebars
```diff
@@ -0,0 +1,2 @@
+<h3 class="c-list__title">Random Word</h3>
```
moollaza: The title should still be handled by the Text template, not in the `content`.

Let's remove everything in here except the `<a>` tag

share/spice/rand_word/rand_word.js
```diff
@@ -31,6 +31,20 @@
                     moreAt: true
                 }
             }
-        });
+        };
+        
+        if ( list_of_list.length == 1 ) {
+            spiceObj.data = list_of_list[0];
```
moollaza: Let's capture the first item in `api_result` above, before we start splicing the list, so that we can easily access it here:

```js
var first = api_result[0];
...

if ( list_of_list.length == 1 ) {
    spiceObj.data = {
        title: "Random Word",
        word: first.word
    };
    spiceObj.templates = {
         group: 'text',
        options: {
            content: Spice.rand_word.single_word,
             moreAt: true
        }
    };
}
```

codenirvana:
@moollaza I think the issue is fixed now.

codenirvana:
@moollaza Ping!

share/spice/rand_word/rand_word.js
```diff
@@ -31,6 +33,24 @@
                     moreAt: true
                 }
             }
-        });
+        };
+        
+        if ( list_of_list.length == 1 ) {
```
moollaza: This isn't correct. `list_of_list` is an array of 4 element arrays. So right now, if you query `2/3/4 random words` we're only showing a single word now. We need to check the number of total results we received and make sure it's only 1

codenirvana:
@moollaza ```list_of_list[0].length``` fixed the issue!

codenirvana:
@moollaza Is this issue solved with my last change?

moollaza:
@codenirvana Thanks! This LGTM :+1:

PR #2257, opened 2015-11-01
WordMap, added another words for trigger : Spice

twinword:
another_word_for_love_at_duckduckgo

### DIFF
--- a/lib/DDG/Spice/WordMap.pm
+++ b/lib/DDG/Spice/WordMap.pm
@@ -35,10 +35,14 @@ triggers startend => (
     "word cluster",
     "word graph",
     "twinword",
-    "levelpump"
+    "visual thesaurus"
 );
 
 triggers start => (
+    "another words for",
+    "another word for",
+    "synonyms for",
+    "synonym for",
     "similar to",
     "relate to",
     "related to",


moollaza:
@twinword we already handle Synonyms with a separate Instant Answer which I feel is more accurate as it gives proper synonyms. In this case, we'd really be giving "related" words, which aren't all synonyms.

For example: https://duckduckgo.com/?q=words+like+verb&ia=answer -- only a few of these are actually synonyms.

twinword:
How about deleting Synonym related triggers? However, Another words for trigger looks relevant. If you think this way, I’ll pull request again with the changes. 


Twinword Inc.
https://twinword.com | Twinword 
https://linkedin.com/company/twinword | LinkedIn
https://fb.com/twinword | Facebook


On Nov 3, 2015, at 4:07 AM, Zaahir Moolla <notifications@github.com> wrote:

@twinword <https://github.com/twinword> we already handle Synonyms with a separate Instant Answer which I feel is more accurate as it gives proper synonyms. In this case, we'd really be giving "related" words, which aren't all synonyms.

For example: https://duckduckgo.com/?q=words+like+verb&ia=answer <https://duckduckgo.com/?q=words+like+verb&ia=answer> -- only a few of these are actually synonyms.

PR #2260, opened 2015-11-02
WordMap, added another words, removed synonyms for trigger

twinword:
Added another words, removed synonyms for trigger.

### DIFF
--- a/lib/DDG/Spice/WordMap.pm
+++ b/lib/DDG/Spice/WordMap.pm
@@ -41,8 +41,6 @@ triggers startend => (
 triggers start => (
     "another words for",
     "another word for",
-    "synonyms for",
-    "synonym for",
     "similar to",
     "relate to",
     "related to",
--- a/lib/DDG/Spice/WordMap.pm
+++ b/lib/DDG/Spice/WordMap.pm
@@ -35,10 +35,14 @@ triggers startend => (
     "word cluster",
     "word graph",
     "twinword",
-    "levelpump"
+    "visual thesaurus"
 );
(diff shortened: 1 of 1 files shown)


moollaza:
@twinword I'm not sure this is entirely accurate either, because we're still listing out several "related" words.

@twinword I feel like "another word" or "other word(s)" are simpler terms used to describe a synonym.

For example https://duckduckgo.com/?q=related+to+love&ia=answer -- I wouldn't say that all of those words could be used in place of "love".

Same here: https://duckduckgo.com/?q=related+to+create&ia=answer

I think the current triggers are pretty accurate and give a result that's relevant and correct. I feel it would be more correct to use those triggers for the Synonyms IA. A user who doesn't know the word "synonym" would probably search something like "another word for X". 

Thoguhts?

twinword:
Actually, these wordings are hard to disambiguate. The term, synonym is more like academic. When we try to RECALL relevant words, we do not always expect to have synonyms only. So, I think using synonym for WordMap might mislead, but another words for WordMap is safe.

moollaza:
>The term, synonym is more like academic. When we try to RECALL relevant words, we do not always expect to have synonyms only.

I disagree, I believe that "another word for" is the same as asking for a synonym. I feel that showing antonyms or anything other than a synonym is incorrect.

I don't like comparing our results to others, but both Bing and Google use "another word for X" to show the definition + synonyms. Similarly all the link results are for synonyms, which I feel sends a pretty strong message, that we should only be showing synonyms for those queries.

PR #2262, opened 2015-11-05
New CodeDay Spice

tjhorner:
# CodeDay Spice

This Instant Answer will show [CodeDay](https://codeday.org) events (sorted by closest) when a user searches for "codeday" or a variant of that. Instead of a user having to search "CodeDay" and then click their nearest event, the nearest event will be listed, as well as some other close events. This IA uses the official (and used internally) [CodeDay Clear](http://docs.clear.codeday.org/docs) API to get the events, so the data will always be accurate. Some queries that would trigger this IA are "codeday", "when is codeday?", "codeday.org", and "what is codeday". They will all return a map of all CodeDay events sorted by nearest. This IA is useful for users curious about CodeDay and where they can attend one. The API is totally free, and doesn't require a token (if it ever should, I will send a PR and generate an API application for DDG if needed).

I heard about DuckDuckHack from searching "duckduckgo plugins". Really! I was wondering how DDG could know so much :smile:

Screenshots:

image

image

## Checklist

```
[x] Tested cross-browser compatibility (via BrowserStack)

    - Windows 10
        - [ ] Google Chrome
        - [ ] Firefox
        - [ ] Opera
        - [ ] IE 11
        - [ ] Microsoft edge
        
    - Windows 8
        - [x] Google Chrome (36)
        - [x] Firefox (30)
        - [ ] Opera
        - [x] IE 10
        - [x] IE 11

    - Windows 7
        - [x] Google Chrome (36)
        - [ ] Firefox
        - [ ] Opera
        - [x] IE 8 (it seems DDG itself doesn't even work here)
        - [x] IE 9
        - [ ] IE 10

    - Windows XP
        - [x] Google Chrome (36)
        - [ ] IE 7
        - [x] IE 8 (it seems DDG itself doesn't even work here)

    - Linux (Fedora 22, 4.1.10-200.fc22.x86_64)
        - [x] Google Chrome
        - [x] Firefox
        
    - Mac OSX
        - [ ] Google Chrome
        - [ ] Firefox
        - [x] Opera (12.15)
        - [x] Safari (8)

    - iOS (iPhone)
        - [x] Safari Mobile (iPhone 5S)
        - [ ] Google Chrome
        - [ ] Opera

    - iOS (iPad)
        - [x] Safari Mobile (iPad 4G)
        - [ ] Google Chrome
        - [ ] Opera

    - Android
        - [ ] Firefox
        - [x] Native Browser
        - [x] Google Chrome (N6, N5, latest Chrome)
        - [ ] Opera

```
-----
IA Page: https://duck.co/ia/view/code_day

### DIFF
--- a/lib/DDG/Spice/CodeDay.pm
+++ b/lib/DDG/Spice/CodeDay.pm
@@ -2,31 +2,53 @@ package DDG::Spice::CodeDay;
 # ABSTRACT: Get all CodeDay events, sorted by closest to farthest
 
 use DDG::Spice;
+use Text::Trim;
 
-spice is_cached => 1;
+spice is_cached => 0;
 
 name "CodeDay";
 source "CodeDay.org";
 icon_url "https://codeday.org/assets/img/favicon.ico";
-description "Get the CodeDay event nearest to you and find out more about it.";
-primary_example_queries "codeday near me", "when is codeday?";
-secondary_example_queries "codeday";
+description "Get information about a CodeDay or find the nearest one.";
+primary_example_queries "codeday near me", "when is codeday?", "codeday san diego";
+secondary_example_queries "codeday", "codeday seattle", "what is codeday bay area?";
 category "programming";
 topics "programming", "geek", "computing", "entertainment";
 code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/CodeDay.pm";
 attribution github => ["tjhorner", "TJ Horner"],
             twitter => "tjhorner";
 
-spice to => 'https://clear.codeday.org/api/regions/nearby?lat=$1&lng=$2&limit=50';
-spice from => "(.+)/(.*)";
+spice to => 'https://clear.codeday.org/api/regions/$1?term=$2&lat=$3&lng=$4&limit=50';
+spice from => "(.+)/(.*)/(.*)/(.*)";
 
 spice wrap_jsonp_callback => 1;
 
 triggers any => "codeday", "code day"; # yes, some people spell it like "code day"
 
-handle query => sub {
-    # TODO: Maybe add handling for "codeday [city name]" or "[city name] codeday"?
-    return $loc->latitude,$loc->longitude;
+# adapted from GetEvents
+my $pattern = '';
+my @triggers = ('when((\'|)s| is) (codeday|code day)', 'where((\'|)s| is) codeday', 'what((\'|)s| is) (codeday|code day)', '(codeday|code day)');
+
+handle query_lc => sub {
+    my $query = $_;
+    
+    if((lc $_ eq "codeday") || (index(lc $_, "near me") != -1)){
+        return unless $loc->latitude && $loc->longitude && ($loc->country_code eq "US");
+        return "nearby", " ", $loc->latitude, $loc->longitude;
+    } else {
+        # get the city name
+        foreach $pattern (@triggers) {
+             s/$pattern//gi;
+        }
+        
+        my $query = trim($_);
+        
+        return "search", $query, " ", " " if $query ne "";
+        
+        # they weren't searching for a specific event, give them the nearest one
+        return unless $loc->latitude && $loc->longitude && ($loc->country_code eq "US");
+        return "nearby", " ", $loc->latitude, $loc->longitude;
+    }
 };
 
 1;
--- a/lib/DDG/Spice/CodeDay.pm
+++ b/lib/DDG/Spice/CodeDay.pm
@@ -32,6 +32,8 @@ my @triggers = ('when((\'|)s| is) (codeday|code day)', 'where((\'|)s| is) codeda
 handle query_lc => sub {
     my $query = $_;
     
+    s/\?//gi;
+    
     if((lc $_ eq "codeday") || (index(lc $_, "near me") != -1)){
         return unless $loc->latitude && $loc->longitude && ($loc->country_code eq "US");
         return "nearby", " ", $loc->latitude, $loc->longitude;
--- a/lib/DDG/Spice/CodeDay.pm
+++ b/lib/DDG/Spice/CodeDay.pm
@@ -29,7 +29,7 @@ triggers any => "codeday", "code day"; # yes, some people spell it like "code da
 my $pattern = '';
 my @triggers = ('when((\'|)s| is) (codeday|code day)', 'where((\'|)s| is) codeday', 'what((\'|)s| is) (codeday|code day)', '(codeday|code day)');
 
-handle query_lc => sub {
+handle query => sub {
     my $query = $_;
     
     s/\?//gi;
--- a/lib/DDG/Spice/CodeDay.pm
+++ b/lib/DDG/Spice/CodeDay.pm
@@ -1,0 +1,32 @@
+package DDG::Spice::CodeDay;
+# ABSTRACT: Get all CodeDay events, sorted by closest to farthest
+
+use DDG::Spice;
+
+spice is_cached => 1;
+
+name "CodeDay";
+source "CodeDay.org";
+icon_url "https://codeday.org/assets/img/favicon.ico";
+description "Get the CodeDay event nearest to you and find out more about it.";
+primary_example_queries "codeday near me", "when is codeday?";
+secondary_example_queries "codeday";
+category "programming";
+topics "programming", "geek", "computing", "entertainment";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/CodeDay.pm";
+attribution github => ["tjhorner", "TJ Horner"],
+            twitter => "tjhorner";
+
+spice to => 'https://clear.codeday.org/api/regions/nearby?lat=$1&lng=$2&limit=50';
+spice from => "(.+)/(.*)";
+
+spice wrap_jsonp_callback => 1;
+
+triggers any => "codeday", "code day"; # yes, some people spell it like "code day"
+
+handle query => sub {
+    # TODO: Maybe add handling for "codeday [city name]" or "[city name] codeday"?
+    return $loc->latitude,$loc->longitude;
+};
+
+1;
--- a/share/spice/code_day/code_day.js
+++ b/share/spice/code_day/code_day.js
@@ -23,14 +23,18 @@
                         data_front: {
                             showPin: true,
                             title: "CodeDay " + event.name,
-                            altSubtitle: startDate.format("MMMM Do") + " - " + endDate.format("Do") + ", noon - noon",
+                            altSubtitle: startDate.format("MMM Do") + " - " + endDate.format("Do") + ", noon - noon",
                             titleClass: 'tile__title--3 tx--16 tx--bold mg--none',
-                            altSubClass: 'tx--13 tx-clr--grey'
+                            altSubClass: 'tx--13 tx-clr--grey',
+                            city: venue.address.city,
+                            state: venue.address.state,
+                            
+                            footer_content: Spice.code_day.footer_front
                         },
                         data_back: {
                             title: "CodeDay " + event.name,
                             url: event.current_event.urls.home,
-                            description: venue.address.city + ", " + venue.address.state,
+                            description: venue.address.line_1 + "; " + venue.address.city + ", " + venue.address.state,
                             
                             footer_content: Spice.code_day.footer_back,
 
@@ -38,6 +42,7 @@
                         },
                         address: venue.address.line_1,
                         city: venue.address.city,
+                        state: venue.address.state,
                         lat: parseFloat(event.location.lat),
                         lon: parseFloat(event.location.lng)
                     };
--- a/share/spice/code_day/code_day.js
+++ b/share/spice/code_day/code_day.js
@@ -1,0 +1,65 @@
+(function (env) {
+    "use strict";
+
+    env.ddg_spice_code_day = function(api_result){
+        if (!api_result || api_result.length === 0) {
+            return Spice.failed('code_day');
+        }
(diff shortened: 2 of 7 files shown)


daxtheduck:
@tjhorner Hey!
        
Thanks!

tjhorner:
Hm, let me check and see why that build failed.

moollaza:
@tjhorner that's not your fault, that's something new we released. It doesn't know what Instant Answer your files are connected to because there is not yet an Instant Answer page for your IA.

I'll make one now.

/cc @zachthompson Can we prevent Travis from running the IAChangelog code?

lib/DDG/Spice/CodeDay.pm
```diff
@@ -0,0 +1,32 @@
...
+
+spice to => 'https://clear.codeday.org/api/regions/nearby?lat=$1&lng=$2&limit=50';
+spice from => "(.+)/(.*)";
+
+spice wrap_jsonp_callback => 1;
+
+triggers any => "codeday", "code day"; # yes, some people spell it like "code day"
+
+handle query => sub {
+    # TODO: Maybe add handling for "codeday [city name]" or "[city name] codeday"?
```
moollaza: Right now this triggering is very broad. Any and all queries with "code\s?day" in them will trigger this IA and return a result.

Lets add in a regex here to ensure the query matches some expected pattern:

```perl
handle query_lc => sub {
    return unless m/^when (is|will be) code\s?day$/; #just a rough example
    return unless exists $loc->latitude && exists $loc->longitude; #ensure we have a lat/lon for the user
    return $loc->latitude, $loc->longitude;
}
```

tjhorner: Sure thing, what's a good syntax to use here for multiple regexes?

tjhorner:
So how would I make it query a different API if I wanted?

tjhorner:
Alright, added some checks and added support for looking up a specific event. Will add tests later.

zachthompson:
@moollaza I'm not sure we want to disable it.  Travis is testing whether the repo is releasable or not with the change.  It's a helpful diagnostic because it wouldn't be releasable without an IA page and it really should be created by this point.

tjhorner:
@moollaza any other problems you can notice?

tjhorner:
Moment of realization. I should have made a separate branch for this, I want to develop another Spice right now but since I pushed to `master` I can't lol

lib/DDG/Spice/CodeDay.pm
```diff
@@ -0,0 +1,56 @@
...
+# adapted from GetEvents
+my $pattern = '';
+my @triggers = ('when((\'|)s| is) (codeday|code day)', 'where((\'|)s| is) codeday', 'what((\'|)s| is) (codeday|code day)', '(codeday|code day)');
+
+handle query_lc => sub {
+    my $query = $_;
+    
+    s/\?//gi;
+    
+    if((lc $_ eq "codeday") || (index(lc $_, "near me") != -1)){
```
moollaza: no need for the `lc`. We've already done that for you via the `handle query_lc` :smile:

moollaza:
>Moment of realization. I should have made a separate branch for this, I want to develop another Spice right now but since I pushed to master I can't lol

@tjhorner whoops, that’s no problem, you can still develop another IA on another branch, but you'll want to do something like `git reset --hard upstream/master` to make the new branch match our current master branch.

>So how would I make it query a different API endpoint if I wanted?
>Right now I'm just sending in the endpoint I should call and the params.

That's the simplest way to use multiple endpoints right now. The other option is a separate Spice Perl package, with it's own `spice to`.

share/spice/code_day/code_day.js
```diff
@@ -0,0 +1,63 @@
...
+                    var venue = event.current_event.venue,
+                        startDate = moment(event.current_event.starts_at * 1000), // convert from unix epoch
+                        endDate = moment(event.current_event.ends_at * 1000);
+                    
+                    return {
+                        id: event.id,
+                        data_front: {
+                            showPin: true,
+                            title: "CodeDay " + event.name,
+                            altSubtitle: startDate.format("MMMM Do") + " - " + endDate.format("Do") + ", noon - noon",
```
moollaza: Lets use the abbreviated month name, I don't think this will fit on mobile screens.

moollaza: Actually, let's mimic the GetEvents design: `Abbreviated month, start time-end time`

share/spice/code_day/code_day.js
```diff
@@ -0,0 +1,63 @@
...
+                            titleClass: 'tile__title--3 tx--16 tx--bold mg--none',
+                            altSubClass: 'tx--13 tx-clr--grey'
+                        },
+                        data_back: {
+                            title: "CodeDay " + event.name,
+                            url: event.current_event.urls.home,
+                            description: venue.address.city + ", " + venue.address.state,
```
moollaza: Let's put this (City + State) in the footer of the front tile and the full address on the back of the tile

share/spice/code_day/code_day.js
```diff
@@ -0,0 +1,63 @@
...
+                        id: event.id,
+                        data_front: {
+                            showPin: true,
+                            title: "CodeDay " + event.name,
+                            altSubtitle: startDate.format("MMMM Do") + " - " + endDate.format("Do") + ", noon - noon",
+                            titleClass: 'tile__title--3 tx--16 tx--bold mg--none',
+                            altSubClass: 'tx--13 tx-clr--grey'
+                        },
+                        data_back: {
+                            title: "CodeDay " + event.name,
```
moollaza: Let's put the venue name here -- this way the back of the tile will describe the venue in more detail

share/spice/code_day/code_day.js
moollaza: We should probably also indicate the # of tickets remaining -- that seems like fairly important info? Similarly the price. This info can go in the front footer I think. 

This way the front gives you info about the event, and the back gives you info about the location

share/spice/code_day/code_day.js
```diff
@@ -0,0 +1,63 @@
...
+                    if(!event.current_event) return null;
+                    var venue = event.current_event.venue,
+                        startDate = moment(event.current_event.starts_at * 1000), // convert from unix epoch
+                        endDate = moment(event.current_event.ends_at * 1000);
+                    
+                    return {
+                        id: event.id,
+                        data_front: {
+                            showPin: true,
+                            title: "CodeDay " + event.name,
```
moollaza: Seems live the `event.current_event.name` is ideal?

tjhorner: I didn't use that because it says the season name as well, which is very redundant and takes up more space.

share/spice/code_day/code_day.js
tjhorner: > We should probably also indicate the # of tickets remaining

How about we only show it when it's less than 20? I was considering this earlier but someone might search it, see "87 Tickets Remaining" and think "oh I have time I'll register later" and then forget about it.

tjhorner: > Let's put this (City + State) in the footer of the front tile and the full address on the back of the tile

Address doesn't fit most of the time, it overlaps the footer. Maybe some CSS to make the tile bigger?

share/spice/code_day/code_day.js
tjhorner: It does fit on mobile screens, it's not the best. Abbreviated month does sound better.

lib/DDG/Spice/CodeDay.pm
tjhorner: :+1: will change in the next commit

tjhorner:
Changes in the diff comments have been committed.

lib/DDG/Spice/CodeDay.pm
```diff
@@ -0,0 +1,56 @@
...
+spice to => 'https://clear.codeday.org/api/regions/$1?term=$2&lat=$3&lng=$4&limit=50';
+spice from => "(.+)/(.*)/(.*)/(.*)";
+
+spice wrap_jsonp_callback => 1;
+
+triggers any => "codeday", "code day"; # yes, some people spell it like "code day"
+
+# adapted from GetEvents
+my $pattern = '';
+my @triggers = ('when((\'|)s| is) (codeday|code day)', 'where((\'|)s| is) codeday', 'what((\'|)s| is) (codeday|code day)', '(codeday|code day)');
```
moollaza: We can simplify this: `((when|where|what)((\'| i|)s) )?(code\s?day)`

> https://www.debuggex.com/r/VIQB_vXrQV8qKkLD

moollaza:
@tjhorner when I tested this PR the other day I was getting 500 Errors from DuckPAN when making requests to the CodeDay API. Today I'm in our office and I'm still seeing the 500 errors. Are you seeing these as well?

When I curl the API it works, but it doesn't seem to like DuckPAN. I'll have to deploy this to Beta to see how its going to behave in a production environment.

lib/DDG/Spice/CodeDay.pm
```diff
@@ -0,0 +1,56 @@
...
+description "Get information about a CodeDay or find the nearest one.";
+primary_example_queries "codeday near me", "when is codeday?", "codeday san diego";
+secondary_example_queries "codeday", "codeday seattle", "what is codeday bay area?";
+category "programming";
+topics "programming", "geek", "computing", "entertainment";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/CodeDay.pm";
+attribution github => ["tjhorner", "TJ Horner"],
+            twitter => "tjhorner";
+
+spice to => 'https://clear.codeday.org/api/regions/$1?term=$2&lat=$3&lng=$4&limit=50';
```
moollaza: @tjhorner this has now been deployed to beta, but it's not working properly. I suspect it's these calls API calls.

I think what we'll have to do, is always supply a lat long, by looking up the lat/long for the specified City. We have plans to expose a method for this in the future, but right now we don't have an easy way to do this.

I did some digging and came across http://damog.net/files/misc/zipcodes-csv-10-Aug-2004.zip which is a CSV containing the lat/lon for all US Zip Codes. This is overkill, but if we slim it down to the cities where there will be Code Day events, or maybe just the capitals for each state, we can easily load the file and lookup the lat long for it and make an API call.

How does that sound?

lib/DDG/Spice/CodeDay.pm
tjhorner: I tried this, it only seems to work with the separate regexes. I haven't tried this exact one though, I'll do it later.

lib/DDG/Spice/CodeDay.pm
tjhorner: >but it's not working properly

I noticed that on beta... it works perfectly with the DuckPAN instance on my end, weird.

>or maybe just the capitals for each state, we can easily load the file and lookup the lat long for it and make an API call.

I can add a search by ZIP on our end, I still think lat/long would be more accurate though.

tjhorner:
>Today I'm in our office and I'm still seeing the 500 errors. Are you seeing these as well?

Nope, works here... I can see what's happening on my end that causes the 500, mention me on the DDH slack and I can see what's going on in the logs.

lib/DDG/Spice/CodeDay.pm
moollaza: @tjhorner sorry to clarify I mean we could store a mapping of state names to lat/long here. So then we could stick to only using the one API endpoint. That way `codeday near me` would use lat long from `$loc` and `codeday san diego` would lookup the lat/long in our local textfile, and use that with the search endpoint. This way we don't need to fiddle with using two different API endpoints, or returning blank parameters.

The third solution is to adjust your API and this code so that we can always just return an array of defined data, and don't need to worry about passing a blank space as a value.

e.g.

```perl
spice from => "(.+)/(.+);
spice to => "https://clear.codeday.org/api/lookup?type=$1&term=$2

...

handle query_lc => sub {
    ...
    return "region", $city_name if isCitySearch;
    return "latlong", "$lat|$long" if isLocalSearch;
```

This would require the API to handle things slightly differently, but enables our `return` to be normalized a little and then we don't need to worry about NGINX mishandling empty array values.

What do you think?

tjhorner: > The third solution is to adjust your API and this code so that we can always just return an array of defined data, and don't need to worry about passing a blank space as a value.

This sounds like the best solution. I'll add an endpoint for this later today.

moollaza: :+1:

moollaza:
@tjhorner sorry for the delays here -- I can't remember where we left off. Did we get it working? I ran into SSL issues yesterday with our NGINX and a 3rd party (IsItUp.org). I'm having the team look into that...

tjhorner:
I don't think we ever did... let me check.

Nope, still doesn't work on beta. Let me find my ssh key for the server, and then get the logfile to see what's happening.

So Clear isn't getting any requests when I search "codeday" on beta. Are you still using `unsafe.clear.codeday.org`?

moollaza:
@tjhorner I've just re-deployed the current code to Beta. It's not currently using unsafe.clear.codeday.org so it's not working.

If you update the `spice to` to use `http://unsafe.clear.codeday.org` it should work. There seems to be a problem with Perl's LWP and certain SSL sites. I think this is the third site that's given us problems so I'll have our ops team investigate. Using `http://unsafe.clear.codeday.org` I'm able to get this working in DuckPAN.

jdorweiler:
This pull request is being closed due to inactivity but, if anyone would like to continue it, please @ mention 'duckduckhack-contributors' for help and http://docs.duckduckhack.com/submitting/submitting-overview.html for reference.

PR #2264, opened 2015-11-05
Add AnnounceRelease plugin

zachthompson:
Add new dzil plugin to send a message about new releases and their changes.


moollaza:
@zachthompson Thanks! This LGTM :+1:

PR #2273, opened 2015-11-10
Displaying latest news articles for navigational queries to news sites

brianrisk:
@moollaza @b1ake @nilnilnil 

Demoed this to Blake for breaking news.  He recommended getting to PR state.  Thanks Zaahir and Caine for heads-up on the endpoint.

Triggers include:

### news sites
* msnbc
* nyt
* ny times
* nytimes
* the times
* new york times
* bbc
* google news
* cbc
* cbc news
* rt
* sky news
* abc news
* e online
* eonline
* today
* today news

### regions
* boston news
* ny daily news
* ny post
* new york post
* the post
* chicago news
* chicago tribune

### generic queries
* celebrity
* celebrity news

### topic sites
* engadget
* hacker news
* boingboing
* 538
* fivethirtyeight

### other
* clickhole

### people
* nate silver


lib/DDG/Spice/Articles.pm
```diff
@@ -0,0 +1,51 @@
...
+spice to => 'https://duckduckgo.com/x.js?u=$1';
+spice wrap_jsonp_callback => 1;
+spice proxy_cache_valid => "418 1d";
+
+handle sub {
+
+    my $query = lc $_;
+    
+    # find which blog name was used
+    return unless $query =~ m/\b($site_qr)\b/;
```
MrChrisW: Should only return a result if the **exact trigger** is matched to avoid searches with additional words.

**google news** = Result
**google news is cool** = No result

MrChrisW:
This looks great @brianrisk, looks like there's a slight issue with the phrasing of HTML, the endpoint does return HTML entities and these are not handled well by the front end.

**Example**
2015-11-23-04-google news at duckduckgo

@moollaza Suggestions?

brianrisk:
@MrChrisW thanks for catching that.  I added a function to decode entities.  I've also made the query have to exactly match the trigger word.

@moollaza I'd love for the tiles to be larger so as to be more accommodating for titles and descriptions.  Suggestions?  Also, looks like tests not related to this IA are failing.

moollaza:
@brianrisk if you use the `media` group it will allow for taller tiles. Take a look at the DogoNews Spice.

There's also a `news_item` template that could be used: https://github.com/duckduckgo/duckduckgo-answerbar-templates/blob/master/item/news_item.handlebars

share/spice/articles/articles.js
```diff
@@ -0,0 +1,79 @@
...
+    var decodeEntities = (function() {
+      // this prevents any overhead from creating the object each time
+      var element = document.createElement('div');
+
+      function decodeHTMLEntities (str) {
+        if(str && typeof str === 'string') {
+          // strip script/html tags
+          str = str.replace(/<script[^>]*>([\S\s]*?)<\/script>/gmi, '');
+          str = str.replace(/<\/?\w(?:[^"'>]|"[^"]*"|'[^']*')*>/gmi, '');
+          element.innerHTML = str;
```
moollaza: This should be avoided -- we had used this approach before but ran into issue which you can read about here:

http://stackoverflow.com/questions/1147359/how-to-decode-html-entities-using-jquery#comment6018122_2419664

The safest approach is to do it manually like http://underscorejs.org/docs/underscore.html#section-161 -- ideally we should port this functionality to a new DDG helper in https://github.com/duckduckgo/duckduckgo-utils/blob/master/util.js

nilnilnil:
@brianrisk thanks for this. I think the best way to integrate this is to take the idea and integrate into our current news service. That way we can make sure we don't end up with race conditions and a weird experience.

moollaza:
Closing this for now as we'll be tackling internally.

PR #2274, opened 2015-11-11
Missing Kids

brianrisk:
@moollaza @nilnilnil @chrismorast @yegg 

I've converted the Missing Kids goodie to a spice so that it can better be open sourced.

Zaahir, I'm having trouble getting description to display.  It's a basic_image_item, so it should have one.  What's going wrong?

Chris: thanks for your encouragement to open source.

-----
IA Page: https://duck.co/ia/view/missing_kids

### DIFF
--- a/lib/DDG/Spice/MissingKids.pm
+++ b/lib/DDG/Spice/MissingKids.pm
@@ -1,0 +1,37 @@
+package DDG::Spice::MissingKids;
+
+use DDG::Spice;
+use Text::Trim;
+use YAML::XS 'LoadFile';
+
+# meta data                                                                                                                                           
+primary_example_queries "missing kids";
+secondary_example_queries "missing children";
+description "Shows missing children from your state if you are located in the US.";
+name "Missing Kids";
+code_url "https://duckduckgo.com";
+topics "social";
+category "location_aware";
+attribution github => ['https://github.com/brianrisk','Brian Risk'],
+            email => ['brian@geneffects.com','Brian Risk'],
+            web => ["https://www.geneffects.com", "Geneffects"],
+            twitter => "brianrisk";
+
+# defining our triggers                                                                                                                               
+my @triggers = ('missing child', 'missing kid');
+triggers start => @triggers;
+
+# set spice parameters
+spice to => 'https://duckduckgo.com/x.js?u=http://www.missingkids.com/missingkids/servlet/XmlServlet?act=rss&LanguageCountry=en_US&orgPrefix=NCMC&state=$1';
+spice wrap_jsonp_callback => 1;
+spice proxy_cache_valid => "418 1d";
+
+handle sub {
+    # return two-letter state
+    if ($loc->country_code eq 'US' && $loc->region) {
+    	return $loc->region;
+    }
+    return;
+};
+
+1;
--- a/lib/DDG/Spice/MissingKids.pm
+++ b/lib/DDG/Spice/MissingKids.pm
@@ -16,7 +16,7 @@ attribution github => ['https://github.com/brianrisk','Brian Risk'],
             twitter => "brianrisk";
 
 # defining our triggers                                                                                                                               
-triggers query_lc => qr/missing (?:kids?|child(ren)?)/;
+triggers start => "missing kid", "missing kids", "missing child", "missing children";
 
 # set spice parameters
 spice to => 'https://duckduckgo.com/x.js?u=http://www.missingkids.com/missingkids/servlet/XmlServlet?act=rss&LanguageCountry=en_US&orgPrefix=NCMC&state=$1';
--- a/lib/DDG/Spice/MissingKids.pm
+++ b/lib/DDG/Spice/MissingKids.pm
@@ -24,11 +24,12 @@ spice wrap_jsonp_callback => 1;
 spice proxy_cache_valid => "418 1d";
 
 handle sub {
+    return unless defined $loc->country_code 
+        && $loc->country_code eq 'US'
+        && defined $loc->region;
+        
     # return two-letter state
-    if ($loc->country_code eq 'US' && $loc->region) {
-    	return $loc->region;
-    }
-    return;
+    return $loc->region;
 };
 
 1;
--- a/lib/DDG/Spice/MissingKids.pm
+++ b/lib/DDG/Spice/MissingKids.pm
@@ -1,15 +1,13 @@
 package DDG::Spice::MissingKids;
 
 use DDG::Spice;
-use Text::Trim;
-use YAML::XS 'LoadFile';
 
 # meta data                                                                                                                                           
 primary_example_queries "missing kids";
 secondary_example_queries "missing children";
 description "Shows missing children from your state if you are located in the US.";
 name "Missing Kids";
-code_url "https://duckduckgo.com";
+code_url "https://github.com/brianrisk";
 topics "social";
 category "location_aware";
 attribution github => ['https://github.com/brianrisk','Brian Risk'],
--- a/lib/DDG/Spice/MissingKids.pm
+++ b/lib/DDG/Spice/MissingKids.pm
@@ -18,8 +18,7 @@ attribution github => ['https://github.com/brianrisk','Brian Risk'],
             twitter => "brianrisk";
 
 # defining our triggers                                                                                                                               
-my @triggers = ('missing child', 'missing kid');
-triggers start => @triggers;
+triggers query_lc => qr/missing (?:kids?|child(ren)?)/;
 
 # set spice parameters
 spice to => 'https://duckduckgo.com/x.js?u=http://www.missingkids.com/missingkids/servlet/XmlServlet?act=rss&LanguageCountry=en_US&orgPrefix=NCMC&state=$1';
--- a/share/spice/missing_kids/missing_kids.js
+++ b/share/spice/missing_kids/missing_kids.js
@@ -42,6 +42,7 @@
                 title = capitalizeWords(title);
                 
                 var description = '';
+                var subtitle = '';
                 if (typeof item.description !== 'undefined') {
                     description = item.description.text;
                     description = DDG.strip_html(description);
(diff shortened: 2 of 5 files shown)


share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
...
+                sourceUrl: 'http://www.missingkids.com',
+                count: articles.length,
+                itemType: 'Missing Kids'
+            },
+            templates: {
+                item: 'basic_image_item',
+                detail: false,
+                options: {
+                    rating: false,
+                    content: true,
```
moollaza: @brianrisk this is overriding the `description`. `content` is meant to be a sub-template, and the basic_image_item template has a check that if `content` is defined, we use it to produce the body instead of the `description`.

I would suggest you use:

```js
templates: {
    group: 'media',
    detail: false,
    item_detail: false
},
```

This uses the `media_item` template which is ideal for a picture, title and description.

The result looks like: 

missing_kids_alabama_at_duckduckgo_and_1____projects_zeroclickinfo-spice__perl_

moollaza: I think this IA would benefit from using an `item_detail` template. Its unfortunate that they don't have a proper API where the description is broken up. Would you be interested in reaching out to them to see if they can make an API for us? I can only imagine it would benefit their cause...

Alternatively I think we should parse the description text and use it with a `record` template to produce a table with the info more clearly laid out:
<table>
    <tr>
        <td> Age now: </td>
        <td> 12 </td>
    </tr>

    <tr>
        <td> Missing: </td>
        <td> 11/07/2015 </td>
    </tr>

    <tr>
        <td> Missing From: </td>
        <td> Melbourne, FL </td>
    </tr>

    <tr>
        <td> Contact: </td>
        <td> Melbourne Police Department (Florida) 1-321-608-6731 </td>
    </tr>
</table>

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
...
+            },
+            normalize: function(item) {
+                var title = item.title.text;
+                // removing "Missing: ";
+                title = title.replace(/.*\:/,"");
+                title = title.replace(/.*\:/,"");
+                var description = '';
+                if (typeof item.description !== 'undefined') {
+                    description = item.description.text;
+                    description = description.replace(/(<([^>]+)>)/ig,"");
```
moollaza: you can use `DDG.strip_html()` for this.

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
...
+                options: {
+                    rating: false,
+                    content: true,
+                },
+            },
+            normalize: function(item) {
+                var title = item.title.text;
+                // removing "Missing: ";
+                title = title.replace(/.*\:/,"");
+                title = title.replace(/.*\:/,"");
```
moollaza: why the duplicate?

We can simplify to: `var title = item.title.text.replace(/.*\:/,"");`

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_missing_kids = function (api_result) {
+        
+        if (!api_result) Spice.failed('missing_kids');
+        if (api_result.error) Spice.failed('missing_kids');
```
moollaza: We can combine these:

```js
if (!api_result || api_result.error) {
    Spice.failed('missing_kids');
}
```

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
...
+        
+        var articles = api_result.rss.channel.item;
+        if (articles.length == 0) Spice.failed('missing_kids');
+        if (articles.length > 20) articles = articles.slice(0,20);
+        
+        var script = $('[src*="/js/spice/missing_kids/"]')[0],
```
moollaza: Might be easier to use `DDG.get_query_encoded().toLowerCase().replace(/missing (kids|children)/, "");

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
...
+        Spice.add({
+            id: 'missing_kids',
+            name: 'Missing Kids',
+            data: articles,
+            meta: {
+                primaryText: 'Missing children in ' + decodedQuery,
+                sourceName: 'National Center for Missing & Exploited Children',
+                sourceUrl: 'http://www.missingkids.com',
+                count: articles.length,
+                itemType: 'Missing Kids'
```
moollaza: I believe this isn't needed because you have `primaryText` defined

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
...
+
+        Spice.add({
+            id: 'missing_kids',
+            name: 'Missing Kids',
+            data: articles,
+            meta: {
+                primaryText: 'Missing children in ' + decodedQuery,
+                sourceName: 'National Center for Missing & Exploited Children',
+                sourceUrl: 'http://www.missingkids.com',
+                count: articles.length,
```
moollaza: don't think this is needed either

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,60 @@
...
+                itemType: 'Missing Kids'
+            },
+            templates: {
+                item: 'basic_image_item',
+                detail: false,
+                options: {
+                    rating: false,
+                    content: true,
+                },
+            },
```
moollaza: We need to remove trailing commas because they can break IE

lib/DDG/Spice/MissingKids.pm
```diff
@@ -0,0 +1,34 @@
...
+triggers query_lc => qr/missing (?:kids?|child(ren)?)/;
+
+# set spice parameters
+spice to => 'https://duckduckgo.com/x.js?u=http://www.missingkids.com/missingkids/servlet/XmlServlet?act=rss&LanguageCountry=en_US&orgPrefix=NCMC&state=$1';
+spice wrap_jsonp_callback => 1;
+spice proxy_cache_valid => "418 1d";
+
+handle sub {
+    # return two-letter state
+    if ($loc->country_code eq 'US' && $loc->region) {
```
moollaza: We need to check that `$loc->country_code` and `$loc->region` are actually defined, because they're not guarenteed to be defined for all locations:

```perl
return unless defined $loc->country_code && $loc->country_code eq 'US';
...
```

lib/DDG/Spice/MissingKids.pm
```diff
@@ -0,0 +1,34 @@
...
+code_url "https://github.com/brianrisk";
+topics "social";
+category "location_aware";
+attribution github => ['https://github.com/brianrisk','Brian Risk'],
+            email => ['brian@geneffects.com','Brian Risk'],
+            web => ["https://www.geneffects.com", "Geneffects"],
+            twitter => "brianrisk";
+
+# defining our triggers                                                                                                                               
+triggers query_lc => qr/missing (?:kids?|child(ren)?)/;
```
moollaza: Let's use trigger words here. I know the regex is simpler, but the trigger words are more efficient.

You *could* use `missing` as a start trigger and use the regex in your `handle`, but I think the best way is to just list out all the phrases.

share/spice/missing_kids/missing_kids.js
brianrisk: Can we combine those?  if api_result is a null object then attempting to access a state variable will throw an error, correct?  So line 5 ensures it is not null and 6 safely checks the error field.

share/spice/missing_kids/missing_kids.js
brianrisk: What you have returns the encoded query.  What I'm doing here is getting the return value from handle sub, which is the two-letter state abbreviation.

brianrisk:
@moollaza thanks for the review; I've put most of those fixes in.  The media template works great.  The one chainge I'd like to not do is item_detail.  If a user thinks they recognize a missing child from that list, then they should click through to the site to see the additional photos and information.

share/spice/missing_kids/missing_kids.js
moollaza: Yup we can. It'll short-circuit when `!api_result === true`

share/spice/missing_kids/missing_kids.js
moollaza: My mistake :+1:

share/spice/missing_kids/missing_kids.js
brianrisk: KK, change is in.

moollaza:
>If a user thinks they recognize a missing child from that list, then they should click through to the site to see the additional photos and information.

@brianrisk I agree with that, but I'd like to make sure the description is easy to read/understand.

How about we pull the the age and location out of the description and put them in the footer, each on their own line. We could put the missing date as the subtitle.

**Note**: If we don't use the description, you need to set `item.description = null` otherwise our template sees that `item.description` is defined (from the JSON) and it causes problems, because it tries to use the obj as an array...

The result could look like this:

missing_kids_at_duckduckgo

share/spice/missing_kids/missing_kids.js
```diff
@@ -0,0 +1,58 @@
...
+                    description = item.description.text;
+                    description = DDG.strip_html(description);
+                    description = description.replace(/&quot;/ig,"\"");
+                    description = description.replace(/&#039;/ig,"'");
+                    // removing PERSON'S NAME, 
+                    description = description.replace(/^.+?,/,"");
+                }
+                return {
+                    title: title,
+                    url: item.link.text,
```
moollaza: I made an awesome discovery!

swap this for `image: item.enclosure.url.replace("t.jpg", ".jpg"),` and we get the full-size images. We'll need to add CSS to increase the tile height, because now the images are a bit too big :+1:

moollaza: If we add some CSS to increase `tile_media` height to `20em` and remove the `padding-bottom` on `tile__title`

It looks like this:

missing_kids_at_duckduckgo

moollaza: We could also put the date string into `Moment.js` and get a pretty date out of it, so its easier to read

brianrisk:
@moollaza added larger image, better parsed info with footer subtemplate

lib/DDG/Spice/MissingKids.pm
```diff
@@ -0,0 +1,35 @@
...
+attribution github => ['https://github.com/brianrisk','Brian Risk'],
+            email => ['brian@geneffects.com','Brian Risk'],
+            web => ["https://www.geneffects.com", "Geneffects"],
+            twitter => "brianrisk";
+
+# defining our triggers                                                                                                                               
+triggers start => "missing kid", "missing kids", "missing child", "missing children";
+
+# set spice parameters
+spice to => 'https://duckduckgo.com/x.js?u=http://www.missingkids.com/missingkids/servlet/XmlServlet?act=rss&LanguageCountry=en_US&orgPrefix=NCMC&state=$1';
```
moollaza: @brianrisk I figured out the problem -- we need to URL encode the content after `&u=` -- I will submit a PR to continue this...

moollaza:
Continuing in #2333

PR #2276, opened 2015-11-11
Remove metadata from template

zachthompson:
* Remove (most) metadata from module template
* Simplify syntax


moollaza:
Thanks! This LGTM :+1:

PR #2281, opened 2015-11-13
On hover, show a loading symbol while the tile gif loads

moollaza:
- This adds a new "loader" png to `/share/riffsy` which displays once the mouse hovers on a tile
- Once the gif for the tile has loaded, the loader is hidden again

- This also removes the `gif` and `gifs` triggers from Gifs.pm as the Riffsy Spice will be handling those triggers now.

- These updates will eventually be ported into the Images template itself.

Demo: https://moollaza.duckduckgo.com/?q=GOAT+GIFS&ia=gifs

/cc @jagtalon @abeyang @bsstoner


jagtalon:
@moollaza checking!

jagtalon:
@moollaza a few questions and comments:

- We can simplify the keyframe to:

```css
@keyframes loader-animate {
    0%, from {
        background-position: 1024px;
    }
    100%, to {
        background-position: 0px;
    }
}
```

to 

```css
@keyframes loader-animate {
    0% {
        background-position: 1024px;
    }
    100% {
        background-position: 0px;
    }
}
```

- On mobile you'd have to click on the image twice to get to the detail view which is different from our normal images. I wonder if we want to investigate that UX interaction.

- This works perfectly on IE 9, but it doesn't animate (CSS animation is IE 10 and up) and only shows three dots. I wonder if we're ok with that or if we want a gif fallback.

moollaza:
More updated -- reverted to a different look for this IA. We'll implement the loader embedded in the details overlay into the Images template later.

moollaza:
Looks good on BrowserStack :+1:

jagtalon:
Awesome. Animation falls back nicely on IE 9. A few nitpicky things though:

- `background-position` should be in CSS
- Simplify to `scale(sx)` instead of `scale(sx, sy)`
- `from, 0%` in the keyframes should just be `0%`

Thanks! I'm excited for this!

PR #2267, opened 2015-11-06
New Pwned Spice

b2ddg:
Needs coordination with https://github.com/duckduckgo/duckduckgo/pull/157

**What does your Instant Answer do?**
This IA searches haveibeenpwned.com for email accounts exposed in data breaches

**What problem does your Instant Answer solve (Why is it better than organic links)?**
This information isn't readily apparent from organic links. The text dumps have been nicely parsed and made available through this api.

**What is the data source for your Instant Answer? (Provide a link if possible)**
haveibeenpwned.com

**Why did you choose this data source?**
It's a definitive source and well made api.

**Are there any other alternative (better) data sources?**
no

**What are some example queries that trigger this Instant Answer?**
pwned test@hotmail.com
have i been pwned test@hotmail.com

**Which communities will this Instant Answer be especially useful for? (gamers, book lovers, etc)**
people on the internet

**Is this Instant Answer connected to a DuckDuckHack [Instant Answer idea](https://duck.co/ideas)?**
https://duck.co/ideas/idea/4799/check-if-an-e-mail-has-been-compromised-in-a-hack

**Which existing Instant Answers will this one supersede/overlap with?**
none

**Are you having any problems? Do you need our help with anything?**
no

**What are the terms of use for the API? Will DuckDuckGo need specific authorization (e.g. an API key)? Are there any costs associated with API usage?**
no

**What does the Instant Answer look like? (Provide a screenshot for new or updated Instant Answers)**
screen shot 2015-11-06 at 3 55 29 pm

## Checklist
Please place an 'X' where appropriate.

```
[x] Added metadata and attribution information
[] Wrote test file and added to t/ directory
[x] Verified that instant answer adheres to design guidelines (https://duck.co/duckduckhack/design_styleguide)
[x] Verified that instant answer adheres to code styleguide (https://duck.co/duckduckhack/code_styleguide)
[x] Tested cross-browser compatibility

 Please let us know which browsers/devices you've tested on:
        
    - Mac OSX
        - [x] Google Chrome
        - [x] Firefox

    - iOS (iPhone)
        - [x] Safari Mobile

```

-----
IA Page: https://duck.co/ia/view/pwned

PR #2283, opened 2015-11-13
Pwned Spice: Update design with larger snippet, footer and subtitle

moollaza:
Continuing #2267...

Made some adjustments to this IA. The result looks like this:

have_i_been_pwned_test_hotmail_com_at_duckduckgo_and_1____projects_zeroclickinfo-spice__bash_

I tried using the Clearbit [company logo API](http://blog.clearbit.com/logo) to produce icons for each company and it works pretty well, but in the cases where they have no logo, we end up show ugly `alt` text:

have_i_been_pwned_test_hotmail_com_at_duckduckgo

---

I think we need some padding between the footer and snippet, but that should be handled at the template level, because it will affect all IAs -- /cc @abeyang 

/cc @b2ddg

### DIFF
--- a/lib/DDG/Spice/Pwned.pm
+++ b/lib/DDG/Spice/Pwned.pm
@@ -14,7 +14,7 @@ spice wrap_jsonp_callback => 1;
 spice error_fallback => 'Email address not found';
 
 handle remainder_lc => sub {
-    return $_ if $_ =~ /.+\@.+/;
+    return $_ if $_ =~ /\b([a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,})\b/;
     return;
 };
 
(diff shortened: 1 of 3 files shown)


b2ddg:
@moollaza this looks so much better with the icons. Do you think we should display which information was exposed, like email address, passwords, names, etc, or should we leave it like you have here. I don't have a preference either way. 
Thanks for giving this one a UI pass, it's looking much better

moollaza:
>Do you think we should display which information was exposed, like email address, passwords, names, etc, or should we leave it like you have here. I don't have a preference either way. 

@b2ddg I'm not sure, I was going to have them display in the detail pane, but seeing as you can click the MoreAt link and it shows the info for the specified account it may not be necessary

We could possibly fit it into the `subtitle` or `footer`.

Either way, merge this PR (so your other PR will update) and we can finish it in your original PR.

I'd like to add the icons, but I think we'll want a way for the templates to disable the Alt text, which isn't currently possible (unless we use JS to strip it which is kinda hacky)

b2ddg:
:+1:

PR #2284, opened 2015-11-13
New Pubmed Spice

b2ddg:
This IA searches for PubMed abstracts

It's focused on PubMed which mainly provides medical journal articles. 

https://www.ncbi.nlm.nih.gov/pubmed/

It's the definitive source for medical journal abstracts.

nothing better

pubmed crispr
pubmed optogenetics

bioscience researchers

https://duck.co/ideas/idea/292/pubmed-central-open-access-science-repository

it can potentially overlap with plos

I'm curious if there is an existing template that makes the detail title a link to a specified url.
Also, I'm curious about the use of ajax calls, is this ok practice for spices?

There's no initial authorization, but the source may need to be notified if the api is used often.

screen shot 2015-11-13 at 9 51 53 am

## Checklist

```
[] Wrote test file and added to t/ directory

    - Mac OSX
        - [x] Google Chrome
        - [x] Firefox
        - [x] Safari

    - iOS (iPhone)
        - [x] Safari Mobile

```

-----
IA Page: https://duck.co/ia/view/pubmed


daxtheduck:
Automated screenshot from https://beta.duckduckgo.com

b2ddg:
That screenshot tool is pretty sweet.

share/spice/pubmed/pubmed.js
```diff
@@ -0,0 +1,97 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_pubmed = function(api_result){
+        if (!api_result || api_result.error) {
+            return Spice.failed('pubmed');
+        }
+        api_result = api_result.esearchresult.idlist;
+        var url = 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=pubmed&retmode=json&rettype=abstract&retmax=10&id=';
```
moollaza: @b2ddg this is considered a privacy leak, because the client makes a direct request to the 3rd party.

The solution is to create a "dummy" spice, `DDG::Spice::Pubmed::Data` with a `spice to`, but no callback (or `wrap_jsonp`)

Then you hit the nginx endpoint for that package from here:

```js
$.getJSON("/js/spice/pubmed_data/foo/bar", function(data){
  item.set("description", data.description);
}
```

Make sense?

XKCD does this (and so do a few other IAs): https://github.com/duckduckgo/zeroclickinfo-spice/search?utf8=%E2%9C%93&q=getJSON

b2ddg: Thanks @moollaza, that makes sense.

share/spice/pubmed/pubmed.js
```diff
@@ -0,0 +1,97 @@
...
+            id: "pubmed",
+            name: "Pubmed",
+            data: api_result,
+            meta: {
+                sourceName: "Pubmed",
+                sourceUrl: 'http://www.ncbi.nlm.nih.gov/pubmed/',
+                rerender: [ 'description' ]
+            },
+            normalize: function(item) {
+                var boxData = [{heading: 'Article'}];
```
moollaza: You don't need this :)

It's only if you want to create an InfoBox like we use in Drinks, for example: https://duckduckgo.com/?q=pokemon+bulbasaur&ia=pokedex

share/spice/pubmed/pubmed.js
```diff
@@ -0,0 +1,97 @@
...
+                        item.set('description', abstract_text);
+                    }
+                });
+
+                item.loadedAbstract = 1;
+            },
+            templates: {
+                group: 'text',
+                variants: {
+                    tileSnippet: "large"
```
moollaza: let's try `tileTitle: "3line-small" or "2line" (whichever accomodates the data best)

lib/DDG/Spice/Pubmed.pm
```diff
@@ -0,0 +1,20 @@
...
+name 'Pubmed';
+
+triggers startend => ('pubmed');
+
+spice to => 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi?db=pubmed&term=$1&retmode=json';
+
+spice wrap_jsonp_callback => 1;
+
+handle remainder_lc => sub {
+    return $_;
```
moollaza: should check that `$_` is defined:

```perl
handle remainder_lc => sub {
    return unless $_;

    # potentially validate remainder here
    return unless m/\w+/;
    return $_;
}
```

share/spice/pubmed/pubmed.js
```diff
@@ -0,0 +1,97 @@
...
+            success: function(r) {                                                                                             
+                delete r.result.uids;
+                api_result = r.result;
+                var articles = Object.keys(api_result).map(function(key){return api_result[key]});
+                api_result = articles;
+            }
+        });    
+        Spice.add({
+            id: "pubmed",
+            name: "Pubmed",
```
moollaza: Should be an IA Pages topic. `Reference` or `Science` for example

share/spice/pubmed/pubmed.js
```diff
@@ -0,0 +1,97 @@
...
+            },
+            onItemShown: function(item) {
+                var url = 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml&rettype=abstract&id='+item.uid;
+                if (item.loadedAbstract) { return; }
+
+                $.ajax({
+                    url: url,
+                    dataType: 'xml',
+                    success: function(r) {
+                        var abstract_text = r.getElementsByTagName('AbstractText')[0].firstChild.data;
```
moollaza: we should have a falback text when this string is null/undefined, e.g. "No description given"

share/spice/pubmed/pubmed.js
```diff
@@ -0,0 +1,97 @@
...
+                }
+                if (item.source) {
+                    boxData.push ({
+                        label: "Source",
+                        value: item.source,
+                    });
+                }
+                var tmp_url = 'http://www.ncbi.nlm.nih.gov/pubmed/' + item.uid;
+                return {
+                    hrefTitle: item.title,
```
moollaza: don't believe we need `hrefTtitle` or `href` - the Info template doesn't use them for anything

daxtheduck:
Automated screenshot from https://beta.duckduckgo.com

daxtheduck:
Automated screenshot from https://beta.duckduckgo.com

share/spice/pubmed/search/pubmed_search.js
```diff
@@ -0,0 +1,71 @@
...
+                }
+            },
+            onItemShown: function(item) {
+                var url = '/js/spice/pubmed/abstract/'+item.uid;
+                if (item.loadedAbstract) { return; }
+                $.ajax({
+                    url: url,
+                    dataType: 'xml',
+                    success: function(r) {
+                        var abstract_text = r.getElementsByTagName('AbstractText')[0].firstChild.data;
```
b2ddg: I'm not sure what placeholder text to add here. It doesn't look terrible without the text, what do you think?

b2ddg:
@moollaza I think I've got the dummy spices added correctly here. Do I need to do anything to handle the xml api?

moollaza:
@b2ddg sorry for the delay -- yes this needs some updates now that the `alt_to` changes are live. You can ditch the 2 extra Perl packages and use `alt_to` in `Pubmed::Search`.

For the XML support, you need to use x.js and pass it along a uri-escaped URL.

```perl
spice to => 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi?db=pubmed&term=$1&retmode=json'; 

spice alt_to => {
    abstract => {
        # handle XML
        to => 'http://duckduckgo.com/x.js?u=' . uri_escape("http://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml&rettype=abstract&id=") . '$1'
    },
    ids => {
        to => 'http://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=pubmed&retmode=json&rettype=abstract&retmax=10&id=$1'
    }
}
```

lib/DDG/Spice/Pubmed/Search.pm
```diff
@@ -0,0 +1,21 @@
+package DDG::Spice::Pubmed::Search;
```
moollaza: We can actually ditch the `::Search` here now that we don't need the other two packages anymore

share/spice/pubmed/search/pubmed_search.js
moollaza: I think something like "No abstract available" works.

share/spice/pubmed/search/pubmed_search.js
```diff
@@ -0,0 +1,71 @@
...
+                if (item.authors) {
+                    for (var i in item.authors) {
+                        authors.push(item.authors[i].name);
+                    }
+                }
+                var tmp_url = 'http://www.ncbi.nlm.nih.gov/pubmed/' + item.uid;
+                return {
+                    title: item.title,
+                    url: tmp_url,
+                    subtitle: authors.join(", "),
```
jagtalon: Should be something like `authors.join(", ") || ' '` just in case it's missing so that it will take up space in the tile:

screen shot 2016-01-17 at 5 17 30 pm

b2ddg: Thanks @jagtalon, good catch

lib/DDG/Spice/Pubmed/Search.pm
b2ddg: When I remove the ::Search, to make this to DDG::Spice::Pubmed, it seems to be messing up the calls that use /js/spice/pubmed/abstract and /js/spice/pubmed/ids. Is it ok leaving this, or should I investigate why it isn't working as expected?

moollaza: @b2ddg sorry, when the package name is 4+ levels deep, we take the 3rd+ portions and prefix the `alt_to` endpoints with that, creating `pubmed_abstract` and `pubmed_ids`.

If the package is just `DDG::Spice::Pubmed` we don't currently prefix the `alt_to`'s with anything (though maybe we should)?

As a fix, you can manually set the `alt_to` names to `pubmed_abstract` and `pubmed_ids`

moollaza: The AJAX calls would need to be changed to `js/spice/pubmed_abstract` as well, instead of `pubmed/abstract`

moollaza: On second thought, we might as well leave the package as-is and keep things simple :+1:

b2ddg: That makes sense about changing the alt_to and ajax calls. I'll leave as is, but thanks for taking the time to explain it out.

b2ddg: Related to this, I think the tests are failing for this PR because there isn't an IA page for this yet. Is there any specific nomenclature I should use when creating that, or can it be called "Pubmed"?

moollaza:
@b2ddg I've made some updates:

- 4line title
- wider tile
- "More at" links to search query

b2ddg:
@moollaza sweet! Thanks man!

Do you think we should worry about that failing soundcloud test? or is it safe to ignore?

moollaza:
> Do you think we should worry about that failing soundcloud test? or is it safe to ignore?

Safe to ignore, it was a bug that got caught in your branch. The merge test passes so the end result will be OK.

moollaza:
I can't deploy to beta because of the failing test, but I'm pretty confident this will be fine.

b2ddg:
@moollaza does this need a change to the IA page to go live?

moollaza:
@b2ddg yes I haven't gotten around to testing the new IAs to be deployed

Issue #2217, opened 2015-10-07
Fix positioning of rating icon in Movies detail pane


chrismorast:
Can we reposition this icon so that it's centered within the line of text? In addition, can we add a bit of padding to give it a little more breathing room?

![screen shot 2015-10-07 at 1 21 48 pm](https://cloud.githubusercontent.com/assets/7061920/10350112/ff2b25d8-6cf6-11e5-8bae-cb0ad5db7d8c.png)

Note: Required changes are highlighted in the image

------
IA Page:  http://duck.co/ia/view/in_theaters

PR #2285, opened 2015-11-13
Fixes issue #2217

cinlloc:
Fixes issue #2217 using an old CSS class which looks abandoned.


moollaza:
@cinlloc Thanks! This LGTM :+1:

PR #2287, opened 2015-11-17
Skip host veriication with LWP SSL

zachthompson:
Travis breaks when trying to connect to HTTPS links on S3.  This disables the host verification checks for tests.


moollaza:
Thanks! This LGTM :+1:

Issue #1706, opened 2015-03-30
MetaCpan: Switch to Text or Info template group, add Title to result


moollaza:
This IA currently uses the `Base` group with a record template. It should use the `List` template. That will allow us to also provide a `title` and `subtitle` so the result look a bit more like our other IAs.

https://duckduckgo.com/?q=cpanm+Dist%3A%3AZilla&ia=software

We still need to convert this to use the new API and should then be able to show more useful data.
More Details: #898

altern8tif:
Since the new API allows actual searching of MetaCPan packages, was wondering if it would make more sense to use the Text template group instead.

So that it'll look like somewhat like the Rubygems IA:

![screen shot 2015-11-01 at 9 29 33 pm](https://cloud.githubusercontent.com/assets/5793364/10869139/b8b00c66-80df-11e5-941e-9c166b82832c.png)

altern8tif:
On that note, would it make sense to convert all the package manager IAs (ie. npm, homebrew, pypi) into a proper search-based IA as opposed to requiring the exact-matched query as it currently does?

<img width="1214" alt="screen shot 2015-11-02 at 3 34 03 pm" src="https://cloud.githubusercontent.com/assets/5793364/10876389/3cfd44ce-8177-11e5-9ab1-458c38d50a18.png">


@moollaza any thoughts?

moollaza:
>On that note, would it make sense to convert all the package manager IAs (ie. npm, homebrew, pypi) into a proper search-based IA as opposed to requiring the exact-matched query as it currently does?

The MetaCPAN IA had to function that way because of the API, but I'd love to be able to do a general MetaCPAN search and not have to match the casing perfectly when I search a module name.

For NPM, I think it would be OK to move it to a general search as well. I've grown very familiar to the current implementation though and I may miss it :)

/cc @MrChrisW -- I think you asked this question before?

MrChrisW:
@moollaza I remember mentioning this, it would be great to add the functionality to show multiple results provided the API supports it. @altern8tif :+1: 

Another idea was to allow switching of the template on an exact-match query, where only a single result is returned.... however I'm not sure this is currently feasible.

PR #2293, opened 2015-11-18
Issue #1706 - MetaCpan: Switch to Text or Info template group, add Ti…

Mailkov:
Fix Issue #1706 - MetaCpan: Switch to Text or Info template group, add Ti…


moollaza:
Thanks! This LGTM :+1:

PR #2295, opened 2015-11-18
Added templates.yml for generic template support for Spice IAs

srvsh:
These are backward compatible changes, as in these can be merged before the changes to the DuckPAN App are merged.

(part of work related to [Issue 268 in p5-app-duckpan](https://github.com/duckduckgo/p5-app-duckpan/issues/268))


template/templates.yml
```diff
@@ -0,0 +1,30 @@
...
+        output: lib/DDG/Spice/<:$ia_package_name:>.pm
+        group:  code
+
+    t/Example.t:
+        output: t/<:$ia_package_name:>.t
+        group:  test
+
+    share/spice/example/example.handlebars:
+        output:   share/spice/<:$lia_name:>.handlebars
+        group:    handlebars
```
zachthompson: optional: 1

template/templates.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    share/spice/example/example.js:
+        output:   share/spice/<:$lia_name:>.js
```
moollaza: This should be in a sub-directory, and needs to be able to handle further subdirectories -- I can't remember if that's builtin already?

e.g:

- Spice::Movie => /share/spice/movie/movie.js
- Spice::Movie::Showtimes => /share/spice/movie/showtimes/movies_showtimes.js

The same applies to all files in share directories

template/templates.yml
```diff
@@ -0,0 +1,30 @@
...
+#           output: output_file # This is a Text::Xslate template itself
+#           group:  group_name  # Human readable group that this file belongs to
+#           optional: 1         # boolean; indicates if this file is optional
+#   
+#   template_name2:
+#       ...
+
+default:
+    lib/DDG/Spice/Example.pm:
+        output: lib/DDG/Spice/<:$ia_package_name:>.pm
```
moollaza: Similarly, this also needs to account for subdirectories

template/templates.yml
zachthompson: It does subdirectories with the `name`, e.g. New::Spice or Newer/Spice.

srvsh:
@zachthompson thanks, I've added a bare minimal CSS file and set both the CSS and handlebars to be optional.

@moollaza thanks, I fixed the ouput paths and used `ia_path` to get the path. I think there was a very minor bug with `ia_path` if the IA name was a path with a lowercase first letter. I've fixed that [here](https://github.com/srvsh/p5-app-duckpan/commit/4ca6147fb2f48d5bf8bd5cc8f92972ecb01043c0).

template/templates.yml
```diff
@@ -0,0 +1,51 @@
...
+    pm:
+        label:  Perl Module
+        input:  lib/DDG/Spice/Example.pm
+        output: lib/DDG/Spice/<:$ia_path:>.pm
+        needs_restart: true
+
+    test:
+        label:  Perl Module Test
+        input:  t/Example.t
+        output: t/<:$ia_path:>.t
```
moollaza: Hmm although I don't think this is bad, we don't currently follow this. Right now all IA's are just `t/<package_name>.t` -- I see only 1 Spice that is nested.

Let's stick with this. I'll have the community organize the tests :+1:

moollaza:
@srvsh this is all looking pretty good. I have one more piece of feedback and then I think this is all ready for merge + release.

1____projects_zeroclickinfo-goodies__bash_

For Cheat Sheets we should change that last line to something more like `"Successfully created new Cheat Sheet: MyCheatSheet"`

template/templates.yml
srvsh: @moollaza I'm sorry, I'm not very clear about what I should do... Should I leave it as `t/<:$ia_path:>.t`?

srvsh:
@moollaza Thanks for your suggestions! Would it be better to fail with an error message if a user passes a name with path separators for cheat sheets? That way, a user will never see `::` in a cheat sheet IA name. I think we can do this with a new field in the `templates.yml` file.

Do you think we can replace the 'success' message with just `Success!` and not print the IA type and IA name? My reasoning is that we already show a message at the start such as `Creating a new Cheat Sheet Instant Answer...` to keep the user informed about what is happening. Also, the paths of the created files are shown, so the name of the IA may not be very important to the user.

template/templates.yml
moollaza: Ya we can leave this as is. I think nesting the testes like the packages makes sense

moollaza:
>Thanks for your suggestions! Would it be better to fail with an error message if a user passes a name with path separators for cheat sheets? That way, a user will never see :: in a cheat sheet IA name. I think we can do this with a new field in the templates.yml file.

Sure, we should probably limit Cheat Sheet names to `/a-zA-Z0-9\s/`.

>Do you think we can replace the 'success' message with just Success! and not print the IA type and IA name?

Sure, let's try that and see how it goes :)

srvsh:
@moollaza Done, I have addressed your comments [here](https://github.com/duckduckgo/p5-app-duckpan/pull/279#issuecomment-163756573) and [here](https://github.com/duckduckgo/zeroclickinfo-goodies/pull/1778#issuecomment-163756958). Do check it out and let me know if it is as expected!

moollaza:
Sorry for the delay, I'll review this soon, hopefully tomorrow!

moollaza:
@srvsh Thanks! This LGTM :+1:

PR #2302, opened 2015-11-24
Fix people in space metabar text

bsstoner:
@andrey-p 
cc @moollaza

### DIFF
--- a/share/spice/people_in_space/people_in_space.js
+++ b/share/spice/people_in_space/people_in_space.js
@@ -40,6 +40,7 @@
             signal: "high",
             data: api_result,
             meta: {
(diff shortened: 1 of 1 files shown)


moollaza:
This LGTM :+1:

share/spice/people_in_space/people_in_space.js
```diff
@@ -40,6 +40,7 @@
             signal: "high",
             data: api_result,
             meta: {
+                primaryText: api_result.number + (api_result.number == 1 ? " Person" : " People") + ' in Space',
```
moollaza: I guess we could use `DDG.plural(api_result.number, "Person", "People")` but this is fine.

PR #2309, opened 2015-11-26
Modify news callback to accept a hash

tommytommytommy:
This PR makes news work more like images and other services by accepting a hash (instead of just an array of stories). Requires simultaneous release with an internal PR.


b2ddg:
:+1:

PR #2310, opened 2015-11-27
spice_anime - ignoring "currently airing"

iambibhas:
To avoid triggering this IA when searching for `currently airing anime`.


jagtalon:
@iambibhas makes sense! :+1:

PR #2329, opened 2015-12-09
Github: Check repository size in relevancy block

### LINKED ISSUE #2318: GitHub: result links to an empty repo
https://duckduckgo.com/?q=github+api+contributor+stats&ia=software
links here https://github.com/ukasik123/Testscript-3

------
IA Page:  http://duck.co/ia/view/github

MrChrisW:
Fixing #2318 

@moollaza I'm attempting to use a relevancy block to match any number that is not 0 for the [size](https://gist.github.com/MrChrisW/23dabf7d791d98cd2d61#file-gistfile1-txt-L98) value returned from the API, this should fix #2318. 

However this doesn't seem to work as expected and I'm receiving a console error `TypeError: h.match is not a function` referring to the function used to check relevancy. 

I can just check the size in the `normalize` block and return nothing if 0, however I felt this was the more "correct" way to do it. Let me know if I'm missing something... :frowning:

### DIFF
--- a/share/spice/github/github.js
+++ b/share/spice/github/github.js
@@ -49,6 +49,9 @@
                         }
                     },
                 normalize: function(item) {
+                    if (item.size === 0) {
+                        return;
+                    }
                     return {
                         title: item.name,
                         subtitle: item.owner.login + "/" + item.name,
--- a/share/spice/github/github.js
+++ b/share/spice/github/github.js
@@ -58,8 +58,7 @@
                 },
                 relevancy: {
                     primary: [
-                        { key: 'description', match: /.+/, strict: false }, // Reject things without a description.
-                        { key: 'size', match: /^[1-9][0-9]*$/, strict: false }
+                        { key: 'description', match: /.+/, strict: false } // Reject things without a description.
                     ]
                 },
                 sort_fields: {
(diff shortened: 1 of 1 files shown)


jagtalon:
@MrChrisW ah I think I know what's going on here. `match` is a method in the type `String` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match, and I think `size` is not a string:

screen shot 2015-12-09 at 11 15 29 am

So what's happening is something like `(1234).match(/^someregex/)` which produces this error.

screen shot 2015-12-09 at 11 17 38 am

jagtalon:
So internally I think it should check or at least call `toString` for this to be generalized.

moollaza:
@MrChrisW I think for now let's just fix by checking the value in `normalize` that way we can get this fix soon.

@jagtalon can you make a task for platform team to ensure we convert the candidate to a string. I'd say its low priority. We may just want a better way of handling numbers in the relevancy block, or a way to provide a callback to be run to allow for custom checks.

MrChrisW:
@jagtalon Thanks for the quick explanation! I see in the internal function it's just using .match without converting toString https://gist.github.com/MrChrisW/7d779b78ff749aea188f#file-gistfile1-js-L32

@moollaza Will update, thank.

MrChrisW:
@moollaza @jagtalon Updated *Github: Ignore items (repositories) with size of 0* - 1d4a447

jagtalon:
:+1:

PR #2330, opened 2015-12-09
SoundCloud: ensure call is encoded; result of fix to DDG lib

moollaza:
Depends on: https://github.com/duckduckgo/duckduckgo/pull/168

/cc @zachthompson @b2ddg


moollaza:
This LGTM :+1:

PR #2339, opened 2015-12-15
Fix #2316: Problem with loading animation

### LINKED ISSUE #2316: Riffsy: Loader Animation is not smooth
https://youtu.be/6BdRYSbaAG0

Looks like the problem is that there are two keyframes named `loader-animate` and they're both doing different things.

screen_shot_2015-12-02_at_10_27_20_pm

I suggest renaming the keyframe in Riffsy to something else. This doesn't affect anything else except for Riffsy so this is a localized problem not to worry.

\+ @bsstoner @moollaza

jagtalon:
A keyframe with the same name has conflicted with Riffsy's keyframe. A better solution for this is to have gif detection integrated into the global image template (thinking of working on this soon!), but for now, we don't want it broken so here's a patch that renames the keyframe so that it doesn't conflict.

https://v.usetapes.com/g3M3Cu1VWN

-----
IA Page: https://duck.co/ia/view/riffsy

### DIFF
--- a/share/spice/riffsy/riffsy.css
+++ b/share/spice/riffsy/riffsy.css
@@ -13,12 +13,12 @@
     -ms-transform: scale(0.5, 0.5);
     -webkit-transform: scale(0.5, 0.5);
     transform: scale(0.5, 0.5);
-    -moz-animation: loader-animate 0.5s steps(16) infinite;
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata post coming soon

moollaza:
@jagtalon This LGTM :+1:

PR #2361, opened 2015-12-28
use avg_rating in quixey results when available

### LINKED ISSUE #2342: Quixey: Should default to custom `rating_avg` for more accuracy
The IA for "wolfram alpha mobile app" shows the wrong rating for the WA app.  It shows the apps are rated .5/5, when it should show 4/5.

------
IA Page:  http://duck.co/ia/view/apps

### LINKED ISSUE #2364: Quixey: Incorrect rating
Showing incorrect app ratings.  searching for "Sproggiwood" shows 1/2 ratings, instead of 4 1/2 rating.

https://duckduckgo.com/?q=Sproggiwood&t=ffab&ia=apps

------
IA Page:  http://duck.co/ia/view/apps

marianosimone:
fixes #2342 and fixes #2364

-----
IA Page: https://duck.co/ia/view/apps

### DIFF
--- a/share/spice/quixey/quixey.js
+++ b/share/spice/quixey/quixey.js
@@ -123,7 +123,7 @@
                     img_m:         screenshot,
                     title:         item.name,
                     heading:       item.name,
-                    rating:        item.rating,
+                    rating:        features.rating_avg || item.rating,
                     reviewCount:   features.allversions_rating_count || features.rating_count || null,
                     url_review:    item.dir_url,
                     price:         pricerange(item),


daxtheduck:
Metadata post coming soon

share/spice/quixey/quixey.js
```diff
@@ -123,7 +123,7 @@
                     img_m:         screenshot,
                     title:         item.name,
                     heading:       item.name,
-                    rating:        item.rating,
+                    rating:        features.rating_avg || item.rating,
```
moollaza: @marianosimone actually we should try and use `custom.avg_rating` when it's defined. It looked like it was accurate, I'm not sure about the `rating_avg` property.

marianosimone: I couldn't find any documentation on the API, but I did find that this attribute has the expected value for all the cases I tried (instead of the 0.x sometimes found in `item.rating`. Any particular reason to prefer `custom.avg_rating` instead?

moollaza:
>I couldn't find any documentation on the API

Sorry about that, it unfortunately requires a login to their partner portal 

> but I did find that this attribute has the expected value for all the cases I tried (instead of the 0.x sometimes found in item.rating.

Ah well that's good to know, this may work then!

>Any particular reason to prefer custom.avg_rating instead?

When I initially looked it had the correct value, I don't recall seeing the `rating_avg` but it may very well work!. The `custom` data, I believe, is directly from the iOS or Play store -- frankly their API even confuses me!

I'll deploy this to beta for testing!

marianosimone:
Actually, `feature` seems like a shortcut for `custom.feature`... It IS very confusing

jagtalon:
@moollaza @marianosimone looks like their API is currently down:

screen shot 2016-01-02 at 3 41 59 pm

Will test it as soon as it goes back up.

moollaza:
@marianosimone Thanks! This LGTM :+1:

moollaza:
Gave a quick visual comparison of beta and production for "flight tracking apps" and the ratings look much more correct. Many currently have no ratings or very low which is definitely wrong

PR #2379, opened 2016-01-06
Aqi - Added term 'local' instead of having to use zip codes

davidofferman:
Hello,
In this update to Aqi (Air Quality Index), I added a new option to provide the word "local" instead of a zip code. This could be very useful if you are traveling to a different city and don't know the local zip code.

**A few questions...**
I'm not sure if I validate the location property correctly (I'm not a Perl developer). Is it sufficient enough to check if the `$loc` and `$loc->{postal_code}` return `true`, and to assume a valid postal code is present? Is there a more preferred way?

In the test file, I'm not sure if the `DDG::Request->new` method can be used along side the method of testing strings by themselves. In all the test files I've seen, they either use one or the other.

I was considering having this Instant Answer trigger without any additional content for the more specific trigger words, such as "air quality" and "air quality index". These queries would then request the postal code from the location variable. I would leave out the trigger "aqi" since that could end up over triggering. What are your opinions on this type of trigger to be added also?

---
Aqi Instant Answer
https://duck.co/ia/view/aqi

### DIFF
--- a/lib/DDG/Spice/Aqi.pm
+++ b/lib/DDG/Spice/Aqi.pm
@@ -20,9 +20,15 @@ spice to => 'http://www.airnowapi.org/aq/observation/zipCode/current/?format=app
 spice wrap_jsonp_callback => 1;
 
 handle remainder => sub {
-    my $query = shift;
+    my $query = lc(shift);
 
-    if($query eq 'local') {
+    if($query eq '' || $query eq 'local' || $query eq 'current') {
+
+        if($query eq '') {
+            # We don't want to trigger on empty strings with AQI as the
+            # trigger.
+            return if $req->{query_raw} =~ /\baqi\b/i;
+        }
 
         # Set $query to failed return value. This way we can keep `else`
         # statements limited.
--- a/lib/DDG/Spice/Aqi.pm
+++ b/lib/DDG/Spice/Aqi.pm
@@ -23,8 +23,21 @@ handle remainder => sub {
     my $query = shift;
 
     if($query eq 'local') {
-        return unless ($loc && $loc->{postal_code});
-        $query = $loc->{postal_code};
+
+        # Set $query to failed return value. This way we can keep `else`
+        # statements limited.
+        $query = '';
+
+        if(defined($loc) && exists($loc->{postal_code}) && exists($loc->{country_code})) {
+            # The current API only supports USA postal codes.
+            if($loc->{country_code} eq 'US') {
+                $query = $loc->{postal_code};
+            }
+        }
+
+        # At this point $query can either be an empty string that we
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Spice/Aqi.pm
```diff
@@ -20,9 +20,18 @@ spice to => 'http://www.airnowapi.org/aq/observation/zipCode/current/?format=app
 spice wrap_jsonp_callback => 1;
 
 handle remainder => sub {
-     my $query = shift;
-     return unless $query =~ qr/^\d{5}$/; #check if the entire remainder string is a 5 digits number;
-     return $query; # we know it exists, it's a number and it has 5 digits;
+    my $query = shift;
+
+    if($query eq 'local') {
+        return unless ($loc && $loc->{postal_code});
```
moollaza: This is probably the most robust approach to verifying the `$loc` object is defined and has the data we need:

https://github.com/duckduckgo/zeroclickinfo-spice/blob/a64e52ea83e7153c1dc33eaf68b2f96c3905f457/lib/DDG/Spice/OpenNIC.pm#L63

I would suggest you also verify that the country is USA and otherwise bail out as this API only supports US zipcodes

davidofferman:
I updated the checks for the `$loc` variables along with making sure the country is US. I also added a few additional comments to hopefully make future changes a breeze. 

I'm good with this commit if everyone else agrees.

lib/DDG/Spice/Aqi.pm
```diff
@@ -20,9 +20,31 @@ spice to => 'http://www.airnowapi.org/aq/observation/zipCode/current/?format=app
 spice wrap_jsonp_callback => 1;
 
 handle remainder => sub {
-     my $query = shift;
-     return unless $query =~ qr/^\d{5}$/; #check if the entire remainder string is a 5 digits number;
-     return $query; # we know it exists, it's a number and it has 5 digits;
+    my $query = shift;
+
+    if($query eq 'local') {
```
moollaza: @davidofferman now that I think about this a little more, I think it's fair to trigger this when there is no remainder as long as the user is in the US.

That way queries like "air quality" would show this, similar to how "weather" shows the local weather.

moollaza: I would also add `current` as an option: Something like `if $query eq "" || query m/^(local|current)$/ {` perhaps?

moollaza:
@davidofferman this is now live on beta! https://beta.duckduckgo.com/?q=aqi%2090210

jagtalon:
@davidofferman this is perfect! :+1:

PR #2377, opened 2016-01-06
People In Space: Use cloudinary platform to resize bio images on-the-fly

MrChrisW:
The large image size delivered by the howmanypeopleinspace.com website has caused slow loading of the astronaut images for the IA. (sorry mobile users :grimacing: )

The astronaut images are only displayed with a size of 100x130 however they're delivered at 640x800 (see [example image](http://www.howmanypeopleareinspacerightnow.com/app/biophotos/mikhail-kornienko.jpg)) using the cloudinary.com free service we can reduce page loading considerably by resizing the images on-the-fly.

The free account offers the following - I've signed up and am using the account with the url name **duckduckgo**

2016-01-06-09-cloudinary management console - dashboard

This would allow all 6 astronaut images to be pulled around 40 times a day within one month - I'm not sure what the traffic to this IA is, so this may not be feasible. 

**Speed improvements** - 70% load improvement.

#### After

cc// @moollaza @jagtalon

PR #2384, opened 2016-01-07
People In Space: Use local images, failback to API images

MrChrisW:
Implementation of @moollaza suggestion in #2377

>One thought is that we could simply store smaller versions of the current images in the repo and have it fallback to the API's image if it's different from the one we're using. The astronauts don't change too often it seems.

This will use the local images and if they don't exist failback to the API images. 

@moollaza Please let me know if there's a better way to implement checking of the local images as currently we're loading them twice. This is where the magic happens [people_in_space.js#L21-L42](https://github.com/MrChrisW/zeroclickinfo-spice/blob/8648a966018ecd602cff1f3152943c78bcd00ce8/share/spice/people_in_space/people_in_space.js#L21-L42). Let me know if the resolution of the images is ok.

------
IA Page:  http://duck.co/ia/view/people_in_space


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

davidofferman:
Saving the images as JPEG with high quality can result in the local images being around 20kb in size instead of 50kb (resolution of 150px by 188px). I'm not sure if that amount in size really matters, since you already reduced images from 500kb to 50kb by storing them locally.

MrChrisW:
Thanks for the suggestion @davidofferman ! I'll update to use JPEG, I think this should be fine for the photos and will reduce size even futher.

share/spice/people_in_space/people_in_space.js
```diff
@@ -17,6 +17,29 @@
...
+            
+            obj.onItemShown = function(item) {
+                // Get filename from biophoto image url. Replace file extention jpg to png
+                // Use DDG.get_asset_path to get full path to images under spice directory
+                var imageFilename = item.biophoto.replace(/^.*[\\\/]/, ''),
+                    localImage = DDG.get_asset_path('people_in_space', "images/" + imageFilename)
+                
+                // Perform a HTTP GET request to retrieve local images using filename from API 
+                // This is used to determine if a local image exists 
+                $.ajax({
```
moollaza: This might be slightly simpler? I haven't tested it though...

```js
$.get(localImage, function(data, statusText, xhr) {
  item.set('image', localImage);
}, "image/jpeg")
.error(function() {
  item.set('image', item.biophoto);
});
```

moollaza:
@MrChrisW this looks pretty good! I'm wondering if we may also need a 2x images for retina screen. I suspect we will..

share/spice/people_in_space/people_in_space.js
MrChrisW: Using [`$.get`](http://api.jquery.com/jQuery.get/) is simply a shorthand method for the [`$.ajax`](http://api.jquery.com/jQuery.ajax/) method.

The reason I don't fire on the `error` callback is because DuckDuckGo production will not return a non 200 HTTP status code, essentially meaning it's a success. Example https://duckduckgo.com/share/spice/people_in_space/1177/image.png - [see status code returned](https://gist.githubusercontent.com/MrChrisW/4d941d83bd9fee542990/raw/5487baa7d70c47c0cf1bf7f69d10c66336a1817e/example.sh). 

Also... I've tested using the dataType parameter at the end of the $.get function like in your example above and am unable to get it to function correctly.

MrChrisW:
@moollaza Updated with retina screen support 4eb2b5a

share/spice/people_in_space/people_in_space.js
moollaza: :+1: :+1: :+1:

moollaza:
Deploying to beta...

MrChrisW:
@moollaza Don't think this got deployed to beta correctly... I'm seeing https://beta.duckduckgo.com/share/spice/people_in_space/1181/people_in_space.spice.js - which doesn't contain these changes

moollaza:
No it didn't, I'm not sure why though -- was trying to figure that out locally...

MrChrisW:
@moollaza Any update on this?

share/spice/people_in_space/people_in_space.js
```diff
@@ -42,14 +68,17 @@
...
+                    'image' // rerender images, required when using item.set in onItemShown
+                ]
             },
             templates: {
                 options: {
                     moreAt: true,
                     content: Spice.people_in_space.content
                 }
-            }
+            },
```
moollaza: :scream: Trailing comma!

moollaza:
@MrChrisW Thanks! This LGTM :+1:

share/spice/people_in_space/people_in_space.js
moollaza: Fixed.

MrChrisW: :tired_face: Sorry! Thx

PR #2407, opened 2016-01-13
Digital Object Identifier lookup

jbarrett:
TBD:
- [ ] Select appropriate template and fill out text
- [x] Verify and fill out tests
- [ ] @nomeata's original version also optionally linked to BibTeX - can we do that?
- [x] How much pre-processing of the data can I do in Perl?
- [x] Category: Reference or Science?
- [x] Trigger space collision - [numeric DOIs trigger Calculator](https://duckduckgo.com/?q=10.1063%2F1.3701566&ia=answer) currently
- [ ] Disambiguation - including "DOI" in query [asks "Did you mean DO I...?"](https://duckduckgo.com/?q=DOI%3A10.1063%2F1.3701566)
- [ ] Requires "internal" redirect support.

Ref: #1535 #119 

IA: https://duck.co/ia/view/doi


daxtheduck:
@jbarrett Hey!
        
Name:
Street Address:
City:
State / Province:
Country:
Phone (Non-US Orders only): 
T-shirt size:
GitHub username:
Link to this Pull Request:

Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Spice/DoiBibtex.pm
```diff
@@ -0,0 +1,25 @@
...
+use DDG::Spice;
+
+description "Look up a digital object identifier (bibtex code)";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/DoiBibtex.pm";
+attribution github => ["https://github.com/nomeata", "Joachim Breitner"],
+            web => ["http://www.joachim-breitner.de", "Joachim Breitner"],
+            email => ['mail@joachim-breitner.de', "Joachim Breitner"];
+status "enabled";
+
+triggers start => 'This spice should never match any request, as it is just a helper for the doi spice.';
```
moollaza: @jbarrett we now have a much simpler way for your to create this now!

http://docs.duckduckhack.com/backend-reference/api-reference.html (checkout the "Multiple API Endpoints" section) -- you'll want to use the `spice alt_to` feature

moollaza:
>  @nomeata's original version also optionally linked to BibTeX - can we do that?

Yup. http://docs.duckduckhack.com/frontend-reference/display-reference.html#moretext-string-or-object-or-array-optional. [Example](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/rain/rain.js#L31)

>  How much pre-processing of the data can I do in Perl?

Of the JSON Data? None. Cleanup/normalization has to be done in the JS.

>   Category: Reference or Science?

Both! You can choose 2 categories!

>   Trigger space collision - numeric DOIs trigger Calculator currently

Calculator is currently being refactored, if the DOI is valid math though, we can either set `signal: "high"` in our JS callback to ensure our tab shows, or add some logic to Calculator to ignore queries that match the DOI regex.

>    Disambiguation - including "DOI" in query asks "Did you mean DO I...?"

We can add logic in the Perl (in `handle`) to check the query against a regex and only return our candidate DOI if the query looks appropriate

lib/DDG/Spice/DoiBibtex.pm
jbarrett: Can you have different `accept_header`s for each `alt_to`?

jbarrett:
Thanks for that @moollaza! 

When I say category, I really mean tab name :)

jbarrett:
So this looks kind of funkadelic right now, so I'm thinking the info template might not be suitable:



Though it is showing much the same info as the original effort:



I'm not sure what the expectation is for "More at" links. In this case you are taken to dx.doi.org, but this essentially functions as a redirect proxy for the referenced content so you'll ultimately end up at, for example, the journal in which the article was published.

Retrieving BibTeX is a second API call - I'm not sure how to handle it. The original effort had custom markup to unhide a dev and populate it with the TeX. If we are presenting it, it should be `<pre>` in any case.

moollaza:
@jbarrett which parts of the data are most useful? What's the use case for searching a doi?

I found crossref.org and used their DOI search tool and it presents: http://search.crossref.org/?q=10.1063%2F1.3701566

Is that the kind of data you'd expect from us?

- title
- publication date
- authors

If so we can probably keep the title of the result as the article title. We could probably put the publication data and set the type as the subtitle (if we pass an array to subtitle we can put more content, they'll be separated with `|`)

We could build a `record_data` object of key-value pairs to point out the rest of the useful info in a table.

It could look something like:


> **My Awesome Article Title**
> Journal Article | Topic: Biomicrofluidics
> ```
> Authors           | A, B, C, D, E |
> Publication Date: | MM/DD/YYYY    |
> Other Stuff       | Foo Bar Baz   |
> ```

moollaza:
> I'm not sure what the expectation is for "More at" links. In this case you are taken to dx.doi.org, but this essentially functions as a redirect proxy for the referenced content so you'll ultimately end up at, for example, the journal in which the article was published.

That sounds fine to me, the idea is we give them most of the important useful info here, and the `More at` link should send them to more detailed info, if available.

>Retrieving BibTeX is a second API call - I'm not sure how to handle it. The original effort had custom markup to unhide a dev and populate it with the TeX. If we are presenting it, it should be <pre> in any case.

What would this info contain? I would think you'd want to grab this extra data in an `onItemShown()` event, and use it to adjust/add to the data, forcing it to re-render with the new info.

We do this for Amazon as the tiles are displayed: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/amazon/amazon.js#L39

Note the `$.getJSON` (where we make our AJAX request) followed by `item.set( <obj> )` to update our data. `meta.rerender` is set above, forcing the item to be re-rendered when `reviewCount` is `set`

jbarrett:
@moollaza This is what I'm not sure about - is the data most useful as a list / table of info or as something copyable as a reference?

The BibTeX is a common format used when referencing in scientific papers, e.g.

```
@article{Mao_2012,
	doi = {10.1063/1.3701566},
	url = {http://dx.doi.org/10.1063/1.3701566},
	year = 2012,
	publisher = {{AIP} Publishing},
	volume = {6},
	number = {2},
	pages = {024113},
	author = {Xiaole Mao and Ahmad Ahsan Nawaz and Sz-Chin Steven Lin and Michael Ian Lapsley and Yanhui Zhao and J. Philip McCoy and Wafik S. El-Deiry and Tony Jun Huang},
	title = {An integrated, multiparametric flow cytometry chip using {\textquotedblleft}microfluidic drifting{\textquotedblright} based three-dimensional hydrodynamic focusing},
	journal = {Biomicrofluidics}
}
```

I haven't used it myself, but scientists I know use it often. Do we have anyone in research who can chime in on this?

jbarrett:
Speaking to my resident scientist @TriploidTree, the clickable title is the most useful thing, authors list is good to have and BibTeX is quite useful (but referencing software can often scrape info themselves using DOIs).

moollaza:
@jbarrett maybe we should make a separate "Bibtex" IA that just pulls the bibtext data into a copy-able div and link to it from this IA?

Alternatively, we could put the publication date, and authors into an `info_box` and put the bibtex content here. Using the `info` template, should suppress the content from being too tall

moollaza:
or rather, if the authors are more useful than the other bibtex info, I'd say put the authors as the `description` and use the infobox to list out all the info in the `bibtext`

@abeyang any thoughts on how we should organize this info?

jbarrett:
The BibTeX is a specific format - we shouldn't present it differently to how we receive it.

I like the idea of having a separate IA and linking to it. Does this approach require new files?

moollaza:
>I like the idea of having a separate IA and linking to it. Does this approach require new files?

If we go that route, you won't need any new files for this IA, but we'll need to make that new IA in a separate PR, with its own files (so the IA pages don't confuse which files belong to which IA)

jbarrett:
Gotcha. The CrossRef presentation is also good for the basic info. I'm thinking there are many circumstances where we don't have, for example, a journal (if pointing to a data set rather than a paper, say) so it will require some small amount of massaging to get it to present consistently.

Seems this is also on beta now, hooray! But the example query only triggers calculator... aww.

moollaza:
> Seems this is also on beta now, hooray! But the example query only triggers calculator... aww.

@jbarrett yup! I deployed it, hoping to demo it at All Hands. Unfortunately it looks like there's an NGINX problem though, because I couldn't get it to return results. The API is being hit, but nothing is coming back: https://beta.duckduckgo.com/js/spice/doi/10.4187%2Frespcare.03984

I believe the issue is that the URL we're hitting redirects, which NGINX typically can't follow. 

e.g. http://dx.doi.org/10.1063/1.3701566 => http://scitation.aip.org/content/aip/journal/bmf/6/2/10.1063/1.3701566

I'm not too sure how we'd solve this one because the final URL seems to be dynamic (meaning we can't simply adjust our `spice to`). We may need to proxy the request to one of our internal endpoints that can follow a redirect...

jbarrett:
I guess it's the `accept_header` parts?

share/spice/doi/doi.js
```diff
@@ -0,0 +1,78 @@
...
+            signal: 'high',
+            meta: {
+                sourceName: "dx.doi.org",
+                sourceUrl: api_result.URL
+            },
+            normalize: function(item) {
+                return {
+                    url: item.URL,
+                    title: item.title,
+                    subtitle: sprintf( '%s, %s, DOI:%s', format_authors(item.author), item.issued.raw, item.DOI ),
```
jagtalon: Whoa, I didn't know we had sprintf defined :))

share/spice/doi/doi.js
```diff
@@ -0,0 +1,78 @@
...
+                        href: "/js/spice/doi_bibtex/" + api_result.DOI,
+                        text: 'BibTeX',
+                    }
+                }
+            }
+        });
+    };
+}(this));
+
+function format_author(author) {
```
jagtalon: Please define these functions inside 

```javascript
(function(env) { ...

}(this));
```

above to limit the scope and prevent them from being available globally.

jagtalon: More information on this: https://bonsaiden.github.io/JavaScript-Garden/#function.scopes

share/spice/doi/doi.js
```diff
@@ -0,0 +1,78 @@
...
+                }
+            },
+            templates: {
+                group: 'info',
+                options: {
+                    moreAt: true,
+                    moreText: {
+                        href: "/js/spice/doi_bibtex/" + api_result.DOI,
+                        text: 'BibTeX',
+                    }
```
jagtalon: Nitpick: please remove trailing commas.

Comma, comma, comma, comma, comma, chameleon

jbarrett: Thanks for feedback, @jagtalon! On this one, I usually leave trailing commas for when I am adding items later (in Perl!) - does this have side effects in js or is it a style thing?

share/spice/doi/doi.js
jbarrett: Haha, don't ask me where it comes from - I was thinking "js has sprintf, right? Oh, it does!"

share/spice/doi/doi.js
jagtalon: @jbarrett no effect! It's a style thing that I get a lot when I contribute to our code because I leave trailing commas as well. Just thought others should share my misery. :))

share/spice/doi/doi.js
jagtalon: @jbarrett apparently someone wrote `sprintf`. 😂

jbarrett:
https://github.com/duckduckgo/duckduckgo/blob/5161375a0beffa5fd12c8b96f67f45ab43a35253/lib/DDG/Rewrite.pm#L142 - code still exists to instruct nginx to pass `Accept` header... unfortunately this is the first IA to require it.

jbarrett:
`Accept` header working, but dx.xoi.org relies on a redirect in some cases to fetch data from, e.g. crossref.

Do we have other instances like this?

edgesince84:
Closing this for now- can be addressed in the future, pending infrastructure changes.  thanks, @jbarrett

PR #2414, opened 2016-01-15
New Grunt Spice

jagtalon:
This plugin gets its data from http://gruntjs.com/plugin-list.json which I admit is a bit big at around 200kb. I think this is ok though since the triggering is limited to "grunt plugins" and "gruntjs plugins". I think anyone searching with those kind of queries have a high signal that they are actually looking for grunt plugins so I think it's worth loading those 200kb (some IAs even load more data--look at recipes. It makes the page 2MB big with the query "cake")

Hope this is ok!

It's modeled after the rubygems spice:

screen shot 2016-01-15 at 3 59 39 am

---

https://duck.co/ia/view/grunt

### DIFF
--- a/lib/DDG/Spice/Grunt.pm
+++ b/lib/DDG/Spice/Grunt.pm
@@ -6,11 +6,11 @@ spice is_cached => 1;
 spice wrap_jsonp_callback => 1;
 spice to => 'http://gruntjs.com/plugin-list.json';
 
-triggers any => 'grunt plugin', 'grunt plugins', 'gruntjs', 'gruntjs plugin', 'gruntjs plugins';
+triggers any => 'grunt plugin', 'grunt plugins', 'gruntjs', 'gruntjs plugin', 'gruntjs plugins', 'grunt.js plugin', 'grunt.js plugins';
 
 handle remainder => sub {
     return $_ if length $_;
-    return $_;
+    return;
 };
 
 1;
--- a/lib/DDG/Spice/Grunt.pm
+++ b/lib/DDG/Spice/Grunt.pm
@@ -6,7 +6,7 @@ spice is_cached => 1;
 spice wrap_jsonp_callback => 1;
 spice to => 'http://gruntjs.com/plugin-list.json';
 
-triggers any => 'grunt plugin', 'grunt plugins', 'gruntjs', 'gruntjs plugin', 'gruntjs plugins', 'grunt.js plugin', 'grunt.js plugins';
+triggers startend => 'grunt plugin', 'grunt plugins', 'gruntjs', 'gruntjs plugin', 'gruntjs plugins', 'grunt.js plugin', 'grunt.js plugins';
 
 handle remainder => sub {
     return $_ if length $_;
--- a/lib/DDG/Spice/Grunt.pm
+++ b/lib/DDG/Spice/Grunt.pm
@@ -1,0 +1,16 @@
+package DDG::Spice::Grunt;
+
+use DDG::Spice;
+
+spice is_cached => 1;
+spice wrap_jsonp_callback => 1;
+spice to => 'http://gruntjs.com/plugin-list.json';
+
+triggers any => 'grunt plugin', 'grunt plugins', 'gruntjs', 'gruntjs plugin', 'gruntjs plugins';
+
+handle remainder => sub {
+    return $_ if length $_;
+    return $_;
+};
+
+1;
--- a/share/spice/grunt/grunt.js
+++ b/share/spice/grunt/grunt.js
@@ -1,0 +1,101 @@
+(function (env) {
+    "use strict";
+    
+    // From http://stackoverflow.com/a/34064434/1546577
+    // This is used to decode entities that is often found in the descriptions.
+    function htmlDecode(input) {
+        var doc = new DOMParser().parseFromString(input, "text/html");
+        return doc.documentElement.textContent;
+    }
+    
+    // Add a star beside the name if it's an official plugin.
+    function officialPlugin(author) {
+        if(author === 'Grunt Team') {
+            return '★';
+        }
+        return '';
+    }
+    
+    env.ddg_spice_grunt = function(api_result) {
+        
+        if (!api_result || !api_result.aaData || !api_result.aaData.length) {
+            return Spice.failed('grunt');
+        }
+        
+        // Get the original query sans the trigger words.
+        var script = $('[src*="/js/spice/grunt/"]')[0],
+            source = $(script).attr("src"),
+            query = source.match(/grunt\/([^\/]*)/)[1];
+        
+        // Return immediately if there is no query.
+        // I don't think we'd like to display all the available plugins.
+        // Although, I could be wrong.
+        if(!query) {
+            return Spice.failed('grunt');
+        }
+        
+        // Generate a regexp out of the query.
+        // We're going to use this to look for relevant items to display.
+        var query_re = new RegExp(query, 'i');
+        
+        // Cache the regexp for removing the "grunt-" at the beginning of all the names.
+        var grunt_re = /^grunt\-/;
+
+        Spice.add({
+            id: "grunt",
+            name: "Grunt Plugins",
+            data: api_result.aaData,
+            meta: {
+                sourceName: 'Grunt',
+                // We can't link to the actual query since filtering is all done on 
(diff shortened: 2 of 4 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=grunt plugin sass)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/grunt_mobile.png)

jagtalon:
I'd like some advice on caching here. The longer the better, but I wouldn't wanna show stale data either.

jagtalon:
Would 30 days be ok?

lib/DDG/Spice/Grunt.pm
```diff
@@ -0,0 +1,16 @@
+package DDG::Spice::Grunt;
+
+use DDG::Spice;
+
+spice is_cached => 1;
+spice wrap_jsonp_callback => 1;
+spice to => 'http://gruntjs.com/plugin-list.json';
+
+triggers any => 'grunt plugin', 'grunt plugins', 'gruntjs', 'gruntjs plugin', 'gruntjs plugins';
```
moollaza: I would use `startend` to be safe

lib/DDG/Spice/Grunt.pm
```diff
@@ -0,0 +1,16 @@
...
+
+handle remainder => sub {
+    return $_ if length $_;
+    return $_;
```
moollaza: I think this should be `return`?

share/spice/grunt/grunt.js
```diff
@@ -0,0 +1,101 @@
+(function (env) {
+    "use strict";
+    
+    // From http://stackoverflow.com/a/34064434/1546577
+    // This is used to decode entities that is often found in the descriptions.
+    function htmlDecode(input) {
+        var doc = new DOMParser().parseFromString(input, "text/html");
```
moollaza: Looks like this isn't supported by IE9 :frowning: 

What entities are expected? Can we do it manually?

We need a standard function for this, even something as brute force as http://underscorejs.org/docs/underscore.html#section-160

share/spice/grunt/grunt.js
```diff
@@ -0,0 +1,101 @@
...
+        // Generate a regexp out of the query.
+        // We're going to use this to look for relevant items to display.
+        var query_re = new RegExp(query, 'i');
+        
+        // Cache the regexp for removing the "grunt-" at the beginning of all the names.
+        var grunt_re = /^grunt\-/;
+
+        Spice.add({
+            id: "grunt",
+            name: "Grunt Plugins",
```
moollaza: I think "Software" is more appropriate?

share/spice/grunt/grunt.js
```diff
@@ -0,0 +1,101 @@
...
+                sourceName: 'Grunt',
+                // We can't link to the actual query since filtering is all done on 
+                // the front-end as well.
+                sourceUrl: 'http://gruntjs.com/plugins'
+            },
+            normalize: function(item) {
+                // Check if the name or description has the query.
+                // We do this because we get all of the plugins available,
+                // and we have to filter the things we need for ourselves.
+                if(!query_re.test(item.name) && !query_re.test(item.ds)) {
```
moollaza: Any reason you did it this way instead of a relevancy block?

share/spice/grunt/grunt.js
```diff
@@ -0,0 +1,101 @@
...
+                    download_count: DDG.commifyNumber(item.dl),
+                    url: 'https://www.npmjs.com/package/' + item.name
+                };
+            },
+            templates: {
+                group: 'text',
+                detail: false,
+                item_detail: false,
+                variants: {
+                    tile: 'basic1',
```
moollaza: No need to use this if you're specifying the title & snippet below :+1:

share/spice/grunt/grunt.js
```diff
@@ -0,0 +1,101 @@
...
+                    tileTitle: '1line',
+                    tileFooter: '2line',
+                    tileSnippet: 'large'
```
moollaza: Looks like there's a lot of white-space in the pictures -- what about switching to `small`?

share/spice/grunt/grunt.js
jagtalon: Software is more appropes, you're right

share/spice/grunt/grunt.js
jagtalon: Oo, yeah, lemme thing of a way. Basically, anything could be in here because the description in NPM is a free country. But maybe converting a small subset is good enough.

share/spice/grunt/grunt.js
jagtalon: Mmmmmm I was lazy? Relevancy block might be better haha. :)

share/spice/grunt/grunt.js
jagtalon: Ah that's perfect.

share/spice/grunt/grunt.js
```diff
@@ -0,0 +1,101 @@
...
+    // This is used to decode entities that is often found in the descriptions.
+    function htmlDecode(input) {
+        var doc = new DOMParser().parseFromString(input, "text/html");
+        return doc.documentElement.textContent;
+    }
+    
+    // Add a star beside the name if it's an official plugin.
+    function officialPlugin(author) {
+        if(author === 'Grunt Team') {
+            return '★';
```
moollaza: Hmm, I think this may look better on the left, or as an `altSubtitle`like "Official Plugin".

I was going to suggest the icon template, but I think it's been removed...

jagtalon:
@moollaza I opted to use `document.implementation.createHTMLDocument` instead from https://developer.mozilla.org/en-US/Add-ons/Code_snippets/HTML_to_DOM. This should prevent XSS attacks, and it's more elegant than doing a regexp like in underscore because it takes advantage of the browser wherein underscore needs to run on other environments like nodejs.

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=grunt plugin sass)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/grunt_mobile.png)

share/spice/grunt/grunt.js
jagtalon: Yup, I like it on the left as well!

jagtalon:
@moollaza so I tried the relevancy block, but I don't think it would work. So I want something that either matches the first OR the second. But doing this:

```
                    { key: 'ds', match: query_re },
                    { key: 'name', match: query_re }
```

Would try to match the first one AND the second one and see if it's true. Not sure if there's a way around that, unless there's a feature of the relevancy block I'm missing. :)

jagtalon:
Ready to review otherwise!

share/spice/grunt/grunt.js
moollaza: @jagtalon if we're gonna stick with the regex, we need to add `\b` around it to prevent substring matches.

e.g. grunt plugin perl returns results containing "pro**perl**y".

https://beta.duckduckgo.com/?q=grunt+plugin+perl&ia=software&iai=0

moollaza: or for now, we can just use a relevancy block and check the title. Chances are something with only a description match might not be as relevant anyways

jagtalon:
@moollaza boop! :)

lib/DDG/Spice/Grunt.pm
```diff
@@ -0,0 +1,16 @@
+package DDG::Spice::Grunt;
+
+use DDG::Spice;
+
+spice is_cached => 1;
+spice wrap_jsonp_callback => 1;
+spice to => 'http://gruntjs.com/plugin-list.json';
+
+triggers any => 'grunt plugin', 'grunt plugins', 'gruntjs', 'gruntjs plugin', 'gruntjs plugins', 'grunt.js plugin', 'grunt.js plugins';
```
moollaza: should probably be `startend`

jagtalon:
@moollaza good point! Done!

moollaza:
Thanks! This LGTM :+1:

PR #2421, opened 2016-01-19
New BoardGameGeek IA

andrey-p:
A Spice for searching game reviews on BoardGameGeek.com

----
IA Page: https://duck.co/ia/view/board_game_geek


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

andrey-p:
Is it normal for Travis to fail for a new IA? It's complaining about missing metadata - not sure if it's something I did or not.

Also, this IA doesn't really look amazing at the moment:

screenshot_2016-01-19_14-49-13

The problem is, the search endpoint only gives you a name, year, and an ID. Then you use the ID to make another API call to grab more useful data like ratings, images, etc. Is this a problem we've dealt with before?

Maybe this would be better done as a Fathead?

GuiltyDolphin:
@andrey-p Yeah - it can take a bit of time for the metadata to be correctly linked (its happened to me!).

And hell yeah - great idea!

moollaza:
>The problem is, the search endpoint only gives you a name, year, and an ID. Then you use the ID to make another API call to grab more useful data like ratings, images, etc. Is this a problem we've dealt with before?

@andrey-p definitely! I'm loading extra info is relatively easy: http://docs.duckduckhack.com/backend-reference/api-reference.html

basically you'll want to create an `spice alt_to` in the Perl. Then use the `onItemShown(item)` to make a request to our 2nd endpoint with each item's ID. We can then use `item.set({ property: value})` to update the tiles on-the-fly

Checkout https://github.com/duckduckgo/zeroclickinfo-spice/blob/43f10037079daa69a0eb455820620a2955a528a9/share/spice/amazon/amazon.js#L48 for a good example

moollaza:
> Maybe this would be better done as a Fathead?

@andrey-p that's also an option, if this info doesn't change often, but the Spice route should work if they have more data available

andrey-p:
> basically you'll want to create an spice alt_to in the Perl. Then use the onItemShown(item) to make a request to our 2nd endpoint with each item's ID. We can then use item.set({ property: value}) to update the tiles on-the-fly

This is awesome! I'll go for this option.

andrey-p:
@moollaza This worked!

Now, I ran into a few problems with the API.

There's a number of bugs with it, one of it means that expansions are included even if you haven't specifically requested them. This means that searching for Ticket to Ride [gives you ~200 results](http://www.boardgamegeek.com/xmlapi2/search?type=boardgame&query=ticket+to+ride), most of which are fan-created expansions and which we don't need.

We can only figure out whether a game is an expansion at the second API call, at which point the tile will have already rendered.

There's two things I could do here:

- Sort results by year and only render the first ~10 ones (expansions will always be made after the original)
- Add `exact=1` in the URL, which limits the number of results greatly. It mostly means we'll only ever get 1 result, and only if the search is an exact text match.

I went for the second option here:

screenshot_2016-01-26_14-49-08

andrey-p:
The other problem is that the API is slow as. But I reckon that'll be less noticeable once we start caching the responses.

jdorweiler:
Would setting a fallback timeout on the IA page work?

moollaza:
>  Sort results by year and only render the first ~10 ones (expansions will always be made after the original)

This could probably work? With a relevancy block we should be able to get the original game assuming the query is accurate enough.

> Add exact=1 in the URL, which limits the number of results greatly. It mostly means we'll only ever get 1 result, and only if the search is an exact text match.

This is obviously the safest bet. Let's see how the order-by-year approach works and if relevancy is an issue we can go with this for now. Maybe we can show the live or demo IA to boardgamegeek and see if they'll improve the API for us?

andrey-p:
@moollaza I've tried the ordering by year thing, and I'm actually pretty happy with the result:

screenshot_2016-02-05_13-47-26

This looked fine for the queries I tried, but I was thinking it could be useful to get it up on the Beta server and get some geeky boardgame members from the community to see if relevancy seems good :)

One other thing I might need to do is to ellipsize the description text. Not sure if it's currently possible without overwriting the template entirely though.

> Maybe we can show the live or demo IA to boardgamegeek and see if they'll improve the API for us?

From the BGG forums I gathered that API support isn't really a high priority for their devs. We might be able to sway them if they're DuckDuckGo users though.

> Would setting a fallback timeout on the IA page work?

@jdorweiler You mean to help with the slowness?

jdorweiler:
@andrey-p I think if the api takes too long to return then the IA doesn't show.  I could be wrong though.  I'm trying to figure out why it doesn't work on beta.

t/BoardGameGeek.t
```diff
@@ -0,0 +1,33 @@
...
+
+use strict;
+use warnings;
+use Test::More;
+use DDG::Test::Spice;
+
+spice is_cached => 1;
+
+ddg_spice_test(
+    [qw( DDG::Spice::BoardGameGeek)],
```
moollaza: This needs to be updated to match the new Perl Module. Then I can deploy to beta!

This is the code needed:

```perl
#!/usr/bin/env perl

use strict;
use warnings;
use Test::More;
use DDG::Test::Spice;

spice is_cached => 1;

ddg_spice_test(
    [qw(DDG::Spice::BoardGameGeek::Search)],
    'carcasonne boardgame' => test_spice(
        '/js/spice/board_game_geek/search/carcasonne',
        call_type => 'include',
        caller => 'DDG::Spice::BoardGameGeek::Search'
    ),
    'catan board game' => test_spice(
        '/js/spice/board_game_geek/search/catan',
        call_type => 'include',
        caller => 'DDG::Spice::BoardGameGeek::Search'
    ),
    'munchkin card game' => test_spice(
        '/js/spice/board_game_geek/search/munchkin',
        call_type => 'include',
        caller => 'DDG::Spice::BoardGameGeek::Search'
    ),
    'where to buy board games' => undef,
    'board games for kids' => undef,
    'board game geek' => undef,
);

done_testing;
```

andrey-p:
All right @moollaza, I've *finally* just updated the tests.

I also bumped up the triggers a bit. Arguably we could try putting the boardgamegeek trigger at the beginning of the query, but I'd like to see how the IA performs when it goes live first.

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=carcassonne boardgame)
                
                [Mobile screenshot](https://images.duckduckgo.com/iu/?u=https://ia-screenshots.s3.amazonaws.com/board_game_geek_mobile.png?ksGule9Xk9)

moollaza:
@andrey-p so the IA is now deployed and triggering...however the images have broken 😞 

Are you free to investigate the cause? Perhaps the API has changed?

share/spice/board_game_geek/search/board_game_geek_search.js
```diff
@@ -0,0 +1,136 @@
...
+                    var responseItem = response.items.item,
+                        rating = DDG.getProperty(responseItem, "statistics.ratings.average.value");
+
+                    // the BGG rating is out of 10 so divide to get a five star rating
+                    if (rating) {
+                        rating /= 2;
+                    }
+
+                    item.set({
+                        img_m: DDG.getProperty(responseItem, "image.text"),
```
moollaza: @andrey-p looks like these URL's all start with `//` now. If we strip that, the images should work again 👍

share/spice/board_game_geek/search/board_game_geek_search.js
```diff
@@ -0,0 +1,136 @@
...
+                        img: DDG.getProperty(responseItem, "thumbnail.text"),
```
moollaza: Same thing here. We need to strip the leading `//`

share/spice/board_game_geek/search/board_game_geek_search.js
moollaza: looks like we may need to swap the `//` for `http://` -- `https://` fails

andrey-p: Odd - I guess it worked in testing because that was on http. I'll update.

andrey-p:
@moollaza Updated, let's hope this works...

To be honest, the BGG API seems pretty flaky so I think we might end up having a lot of problems with it. In any case I really want to see this live :)

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=carcassonne boardgame)
                
                [Mobile screenshot](https://images.duckduckgo.com/iu/?u=https://ia-screenshots.s3.amazonaws.com/board_game_geek_mobile.png?qTy538aWKF)

moollaza:
@andrey-p Thanks! This LGTM :+1:

daxtheduck:
@andrey-p this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

PR #2431, opened 2016-01-21
Arxiv.org abstract retrieval

jbarrett:
https://duck.co/ia/view/arxiv

### DIFF
--- a/lib/DDG/Spice/Arxiv.pm
+++ b/lib/DDG/Spice/Arxiv.pm
@@ -11,11 +11,11 @@ spice wrap_jsonp_callback => 1;
 
 spice to => 'https://duckduckgo.com/x.js?u=http://export.arxiv.org/api/query?id_list=$1';
 
-triggers query_lc => qr/^arxiv[:\s]+?([^\s]+)/;
+triggers start => 'arxiv';
 
-handle matches => sub {
-    return $_[0] if $_[0];
-    return;
+handle query_lc => sub {
+    return unless m/^arxiv[:\s]+?([^\s]+)/;
+    return $1;
 };
 
 1;
--- a/lib/DDG/Spice/Arxiv.pm
+++ b/lib/DDG/Spice/Arxiv.pm
@@ -11,10 +11,11 @@ spice wrap_jsonp_callback => 1;
 
 spice to => 'https://duckduckgo.com/x.js?u=http://export.arxiv.org/api/query?id_list=$1';
 
-triggers any => 'arxiv';
+triggers query_lc => qr/^arxiv[:\s]+?([^\s]+)/;
 
-handle remainder => sub {
-    return $_;
+handle matches => sub {
+    return $_[0] if $_[0];
+    return;
 };
 
 1;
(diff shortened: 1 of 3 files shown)


daxtheduck:
@jbarrett Hey!
        
Name:
Street Address:
City:
State / Province:
Country:
Phone (Non-US Orders only): 
T-shirt size:
GitHub username:
Link to this Pull Request:

Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

jbarrett:
Uses experimental XML support to parse arxiv feeds.

jbarrett:
Getting there, though looks like I may have broken something - not getting a title for my tab:

jbarrett:
Oh, we get it from the live metadata - neato!

lib/DDG/Spice/Arxiv.pm
```diff
@@ -0,0 +1,21 @@
...
+use DDG::Spice;
+
+spice is_cached => 1;
+spice proxy_cache_valid => "200 7d";
+
+spice wrap_jsonp_callback => 1;
+
+spice to => 'https://duckduckgo.com/x.js?u=http://export.arxiv.org/api/query?id_list=$1';
+
+triggers query_lc => qr/^arxiv[:\s]+?([^\s]+)/;
```
moollaza: @zachthompson fwiw this is one of the rare cases where using a word trigger, and splitting words on `-`, `:`, etc. is actually useful. Maybe we should have flag to force matching words on delimiters so the default case is to match on just `\s`?

@jbarrett did you try a word trigger with `arxiv`? I have a feeling it'll work here and then we can do something like:

```perl
handle query_lc => sub {
    return unless m/^arxiv[:\s]+?([^\s]+)/;
    return $1;
}
```

I may be wrong though...

zachthompson: Not sure I follow you.  Why wouldn't you just do a startend and send the remainder to the API?  Can the query to arxiv not contain spaces?

moollaza: Sorry let me clarify:

- @jbarrett is using a regex because the query has the potential to contain no spaces (e.g. "arxiv:1010.2163v1")
- I'm saying because of the way we currently define the "words" in a query, using a word trigger instead of regex should actually work here (this case is pretty rare though, so I just wanted to show you a real example of where this is useful)
- Seeing as a lot of IA's are over-triggering because of the way we currently break up the "words" I still think we should go back to separating on spaces only by default, but we should still provide a way to handle the "words" like we currently do

Does that clarify things?

share/spice/arxiv/arxiv.js
```diff
@@ -0,0 +1,71 @@
...
+                    description: item.feed.entry.summary.text
+                };
+            },
+            templates: {
+                group: 'info',
+                options: get_options( pdf_url )
+            }
+        });
+
+        function get_arxiv_rel_link(links) {
```
moollaza: @jbarrett we should move these out of callback, but keep them inside the closure.

Then they'll be defined outside out callback, but not added to the global namespace, and yet still accessible to our callback function :+1: 

```js
function (env) {
    "use strict";
    
    //private stuff here
    
    env.ddg_spice_arxiv = function (api_result) { ... }
}(this));
```

moollaza:
@jbarrett This LGTM :+1:

Seems to work perfectly on Beta :smile:

jbarrett:
Thanks @moollaza!

PR #2438, opened 2016-01-23
UFC Fighters [WIP]

MrChrisW:
Work in progress....

##### Notes/Questions:
+ CSS is wack and needs fixing - In an attempt to display fighter information within the media tiles I've probably angered the CSS gods :fearful: [ufc.css](https://github.com/MrChrisW/zeroclickinfo-spice/blob/mrchris/ufc_ia/share/spice/ufc/ufc.css)
+ Display fighter information within the tile or simply display the profile picture like the [Google+ IA](https://duckduckgo.com/?q=google%2B+duckduckgo).
+ Titles should probably be narrow.
+ The API returns over 500 fighters, this is currently limited by [splice](https://github.com/MrChrisW/zeroclickinfo-spice/blob/mrchris/ufc_ia/share/spice/ufc/ufc.js#L12) as loading that many images would be a bad idea. It would be nice to have functionality to **load all fighters but lazy load the additional  fighter images on click. I think I've seen this done before but can't find the docs...**
+ Create UFC namespace for additional IAs using the "News" and "Event" endpoints provided. see http://ufc-data-api.ufc.com/

___________________

IA: https://duck.co/ia/view/ufc


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

MrChrisW:
2016-01-25-03-ufc fighters at duckduckgo

share/spice/ufc/footer.handlebars
```diff
@@ -0,0 +1,5 @@
+<div class="info">
+    <div>{{wins}}-{{losses}}-{{draws}} (W-L-D)</div>
+    <div>Nickname: {{nickname}}</div>
```
moollaza: Have you tried putting this as the subtitle? Might make for a more standard UI

moollaza:
@MrChrisW I want to check with the team if we have access to any UFC data through our current provider. This data does look official though which is really nice. I'm more amazed it's available through an open API, which covers so much content. They have real time news feeds, video etc. pretty awesome.

For the images, seeing as we can specify the height, I think in general the design team likes to fill the tile width, it should work well here because these headshots are all consistent.

moollaza:
@MrChrisW also, I'm not a huge fan of the JSON response size. 560kb is quite big. I'm gonna speak to the team about if there's a better way we can handle this. Ideally I'd like the IA to give a detail pane about the fighter which is a better result than showing all the fighters. 

i.e. `ufc demetrius johnson` shows info about just him

share/spice/ufc/footer.handlebars
MrChrisW: :+1:

MrChrisW:
>'m not a huge fan of the JSON response size. 560kb is quite big. I'm gonna speak to the team about if there's a better way we can handle this. Ideally I'd like the IA to give a detail pane about the fighter which is a better result than showing all the fighters.

Completely agree, I did play around with switching templates based on the response as is done in other IA's depending on the search term - to just show details about a single fighter. 

The problem here is you **need** to pull the JSON details of all fighters just to access the information for one, there's no "search" param and accessing using the `/fighter/:id` endpoint just results in a HTML rendered page http://ufc-data-api.ufc.com/api/v3/iphone/fighters/1161 :frowning: 

It would't be hard to internally parse the JSON data and present an endpoint that allows a search param to filter the results, however that's a decision i'll leave to you and the team! 

As always thanks for your feedback :smile:

moollaza:
>there's no "search" param and accessing using the /fighter/:id endpoint just results in a HTML rendered page http://ufc-data-api.ufc.com/api/v3/iphone/fighters/1161 :frowning:

I too noticed the lack of a search endpoint. That's what I'm hoping to discuss with the team because I've seen several IA's depend on a JSON dump + client side search which I want to avoid. I didn't realize the ID specific API result is actually HTML though! :disappointed: that sucks...

MrChrisW:
Hey @moollaza don't worry about this IA at this stage. I'll sort something out with an alternative API and let you know when ready. Cheers

MrChrisW:
@moollaza Hey, wanted your advise on this IA. I've been able to create an API that will return data based on a searched fighter - See example returned data https://gist.github.com/MrChrisW/933c93cc1e607003340a 

My main issue is with the design, I wanted to use the [basic_info_detail](http://docs.duckduckhack.com/assets/diagrams/infobox.png) template. Unfortunately the UFC profile doesn't really contain a decent BIO to fill the template up. See an example:  http://www.ufc.com/fighter/danny-Martinez?id= 

2016-02-07-22-ray borg - official ufc fighter profile

2016-02-07-27testingonly_ufc_danny martinez


The Wiki IA does have some fighter details as well 

2016-02-09-17-danny martinez at duckduckgo


Any suggestions on the design/template to use for this IA?

MrChrisW:
Abandoning due to issues with API.

PR #2445, opened 2016-01-26
MissingKids: Set signal high

moollaza:
This ensures that when the query is "missing kids" if the IA triggers and shows, it's tab will open.

-----
IA Page: https://duck.co/ia/view/missing_kids


GuiltyDolphin:
@moollaza Good thinking :+1:

PR #2447, opened 2016-01-28
Remove background color override for recipes mobile detail panes

bsstoner:
This style seems unnecessary and is breaking the mobile detail pane in dark theme:
screen shot 2016-01-14 at 3 42 04 pm

@moollaza


moollaza:
@bsstoner Thanks! This LGTM :+1:

PR #2450, opened 2016-01-29
Thesaurus: Fixes #412

### LINKED ISSUE #412: Thesaurus missing major meaning
https://duckduckgo.com/?q=thesaurus+friendly

vs

http://words.bighugelabs.com/friendly

sekhavati:
Fixes #412

This PR improves the results displayed by the Thesaurus IA. It also changes the layout. As highlighted in #412 the results displayed were somewhat questionable at times. Below I've listed the changes made to address this and the work that is still outstanding. 

I'd also like to look for a better source of data, perhaps [Wordnik](https://www.wordnik.com/), which is what we use for the Dictionary IA. However that goes beyond the scope of this fix.

### Preview
new-look

### Changes
1. The results returned from the API are now searched and the word type (noun, adjective etc) with the most entries is selected and displayed on the basis that it's likely to be the most common use case for that particular word
2. The results are now displayed using the `list` template group
3. Regardless of the trigger word used (see below) we display all word entries (synonyms, antonyms, related words etc)
3. Deleted redundant Handlebars and CSS files
4. The trigger handling has been simplified to only fire when queries of the following format are submitted:

`<trigger-word> [of|for] <search-word>` or `<search-word> [of|for] <trigger-word>`

where:
`<trigger-word>` is one of `synonym|synonyms|thesaurus|antonym|antonyms`
`<search-word>` is a single word

### Outstanding Issues
1. ~~For reasons unknown to me, unless I return two variables from `Thesaurus.pm` the IA fails to trigger~~
2. ~~`Thesaurus.t` needs to be updated but can't until (1) is addressed~~
3. ~~Is it possible to fix the order of the rows with the `list` template group? (I haven't found a way)~~
4. ~~Lastly, during testing I could only make the _Show More/Less_ button appear by making the width of my window extremely narrow. I'd like it to be present whenever one of the rows has more data than it can display (as is the case above). Is this just a quirk with local servers and it'll be automatically fixed in production or is this something I need to programmatically enforce?~~

show-more

---
IA Page: https://duck.co/ia/view/thesaurus

### DIFF
--- a/lib/DDG/Spice/Thesaurus.pm
+++ b/lib/DDG/Spice/Thesaurus.pm
@@ -21,8 +21,7 @@ handle remainder_lc => sub {
     # Abort if we're left with more than one word to lookup
     return if ($query =~ /\s+/);
   
-    #return $query;
-    return $query, "todo-remove-this-var";
+    return $query;
 };
 
 1;
--- a/lib/DDG/Spice/Thesaurus.pm
+++ b/lib/DDG/Spice/Thesaurus.pm
@@ -3,32 +3,26 @@ package DDG::Spice::Thesaurus;
 
 use strict;
 use DDG::Spice;
+use Text::Trim;
 
-spice from => '([^/]+)/([^/]+)';
 spice to => 'http://words.bighugelabs.com/api/2/{{ENV{DDG_SPICE_BIGHUGE_APIKEY}}}/$1/json?callback={{callback}}';
 
-triggers startend => "synonyms", "synonym", "antonyms", "antonym", "related", "similar", "thesaurus";
+triggers startend => "synonyms", "synonym", "antonyms", "antonym", "thesaurus";
 
-handle query_lc => sub {
-  /^
-      (?:(synonym|antonym|related|similar|thesaurus)s?) \s+
-      (?:(?:terms?|words?)\s+)? (?:(?:of|to|for)\s+)?
-      (\w+) \s*
-      |
-      (\w+) \s+
-      ((synonym|antonym|thesaurus)s?)? \s*
-  $/x;
-
-  my $type = $1 || $4;
-  my $word = $2 || $3;
-
-  return unless $word and $type;
-
-  $type = 'synonym' if $type eq 'thesaurus';
-
-  return $word, $type;
-
-  return;
+handle remainder_lc => sub {
+    my $query = $_;
+    return unless $query;
+    
+    # Remove words that have no bearing on the lookup
+    $query =~ s/\b(of|for)+\b//g;
+    trim $query;    
+    return unless $query;
+    
+    # Abort if we're left with more than one word to lookup
+    return if ($query =~ /\s+/);
+  
+    #return $query;
+    return $query, "todo-remove-this-var";
 };
 
 1;
--- a/share/spice/thesaurus/thesaurus.js
+++ b/share/spice/thesaurus/thesaurus.js
@@ -59,8 +59,8 @@
                 };
                 
                 var result = {
-                    title: query,
-                    subtitle: wordTypeWithMostEntries,                    
+                    title: DDG.capitalize(query),
+                    subtitle: DDG.capitalize(wordTypeWithMostEntries),
                     record_keys : [],
                     record_data : {}
                 };
--- a/share/spice/thesaurus/thesaurus.js
+++ b/share/spice/thesaurus/thesaurus.js
@@ -14,17 +14,17 @@
         
         // Search the results for the word type (noun, adjective etc) with the most entries
         var wordTypeWithMostEntries;
-        var largestWordTypeEntryTotal = 0;
-        
-        Object.keys(api_result).forEach(function(wordType) { // wordType = ajective, noun etc...
-            var wordTypeEntryTotal = 0;
+        var highestEntryCount = 0;
 
-            Object.keys(api_result[wordType]).forEach(function(relationshipType) { // relationshipType = synonym, antonym etc...
-                wordTypeEntryTotal += Object.keys(api_result[wordType][relationshipType]).length;
+        $.each(api_result, function(wordType, words) { // wordType = ajective, noun etc... 
+            var wordTypeEntryCount = 0;
+            
+            $.each(words, function(relationshipType) { // relationshipType = synonym, antonym etc...
+                wordTypeEntryCount += words[relationshipType].length;
             });
-
-            if (wordTypeEntryTotal > largestWordTypeEntryTotal) {
-                largestWordTypeEntryTotal = wordTypeEntryTotal;
+            
+            if (wordTypeEntryCount > highestEntryCount) {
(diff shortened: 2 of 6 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Spice/Thesaurus.pm
```diff
@@ -3,32 +3,26 @@ package DDG::Spice::Thesaurus;
...
+    # Remove words that have no bearing on the lookup
+    $query =~ s/\b(of|for)+\b//g;
+    trim $query;    
+    return unless $query;
+    
+    # Abort if we're left with more than one word to lookup
+    return if ($query =~ /\s+/);
+  
+    #return $query;
+    return $query, "todo-remove-this-var";
```
sekhavati: Outstanding Issue (1) refers to this line of code. I should be able to return what's commented out at line 24, but if I do the IA fails. Does anyone know why?

share/spice/thesaurus/thesaurus.js
```diff
@@ -2,83 +2,90 @@
...
-                            heading : parts_of_speech.charAt(0).toUpperCase() + parts_of_speech.slice(1),
-                            words   : item[parts_of_speech][item.mode].splice(0, 10).join(", ")
-                        });
-                    }
+            data:  api_result[wordTypeWithMostEntries],
+            templates: {
+                group: 'list',
+                options:{
+                    content: 'record', 
+                    rowHighlight : true,
```
moollaza: Wow, I didn't realize this still works. We opted to remove it from all IA's because the design team didn't want it any more. Let's remove that so it looks like all our other "List" IAs

lib/DDG/Spice/Thesaurus.pm
moollaza: I assume you don't have an API key?

If not it's falling back to the live DDG endpoint which incorrectly is using the `spice from` which is expecting 2 values.

If you get your own API key it will work as expected. Once I deploy this to Beta it will be fine :+1:

share/spice/thesaurus/thesaurus.js
```diff
@@ -2,83 +2,90 @@
...
+                    SYNONYMS : "syn",
+                    SIMILAR  : "sim",                    
+                    RELATED  : "rel",
+                    ANTONYMS : "ant"
+                };
+                
+                var result = {
+                    title: query,
+                    subtitle: wordTypeWithMostEntries,                    
+                    record_data : {}
```
moollaza: @sekhavati you can use an array if you'd like, that way the order is preserved.

You can also create a `record_keys` array and the order of that will be used instead.

More details here: https://github.com/duckduckgo/duckduckgo-template-helpers/blob/67f7bb936ad825cf88c66665d0201d07251d707e/template_helpers.js#L851

share/spice/thesaurus/thesaurus.js
```diff
@@ -2,83 +2,90 @@
...
+        var script = $('[src*="/js/spice/thesaurus/"]')[0];
+        var source = $(script).attr("src");
+        var query = source.match(/thesaurus\/([^\/]+)/)[1];
+        query = decodeURIComponent(query);
+        
+        // Search the results for the word type (noun, adjective etc) with the most entries
+        var wordTypeWithMostEntries;
+        var largestWordTypeEntryTotal = 0;
+        
+        Object.keys(api_result).forEach(function(wordType) { // wordType = ajective, noun etc...
```
moollaza: If you want, we have jQuery available, so we typically use `$.each()`. I don't think I've seen anyone use `Object.keys()` before :smile:

share/spice/thesaurus/thesaurus.js
sekhavati: @moollaza Fixed in commit [no.2](https://github.com/sekhavati/zeroclickinfo-spice/commit/21910077d5adf9b793e3403351699be45034de89)

share/spice/thesaurus/thesaurus.js
sekhavati: @moollaza Fixed in commit [no.2](https://github.com/sekhavati/zeroclickinfo-spice/commit/21910077d5adf9b793e3403351699be45034de89)

share/spice/thesaurus/thesaurus.js
sekhavati: @moollaza I've updated this in commit [no.2](https://github.com/sekhavati/zeroclickinfo-spice/commit/21910077d5adf9b793e3403351699be45034de89) as per your suggestion. 

FYI I don't think the section of code you linked to necessarily guarantees the order in which the keys are iterated due to it using for..in (more info [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Array_iteration_and_for...in)).

sekhavati:
@moollaza Can you offer any input for outstanding issue no.4?

lib/DDG/Spice/Thesaurus.pm
sekhavati: @moollaza Thanks, that explained it.

moollaza:
@sekhavati sorry for the delay -- The show more button appears automatically when the height of he content is over a certain threshold. I think by default it forces the height of the rows, so this may be an internal bug. I'd need to see it on beta to determine what's going on. I'll deploy this for testing!

jagtalon:
@moollaza looks like it's not on beta anymore? https://beta.duckduckgo.com/?q=thesaurus+happy&ia=thesaurus

moollaza:
@sekhavati crap I overlooked something -- this will always force the whitespace to be no-wrap even when we _do_ display a show more button.

On mobile this creates issues: 
synonyms_for_beautiful_at_duckduckgo

We may need to apply this CSS via JS, when there's no "Show More"

sekhavati:
@moollaza Are there any examples of doing something similar in other Instant Answers?

moollaza:
> @moollaza Are there any examples of doing something similar in other Instant Answers?

None that come to mind, we've never had to do this before.

@andrey-p any ideas? Can we investigate fixing the chomp instead of employing this workaround?

sekhavati:
@moollaza I guess we could go back to using the old template if `list` is going to be problematic?

moollaza:
> @moollaza I guess we could go back to using the old template if list is going to be problematic? ie: only show synonyms or antonyms etc...

@sekhavati I found a simpler solution for now -- switch the IA from the `list` to the `text` template. That way the IA will not try to chomp the content at all.

It means it will always be full width but I can't find any results that are too long. If necessary we could limit the number of words per type, but it seems fine to me.

We should make a note that we want to switch back to the `list` template though eventually

sekhavati:
@moollaza I applied the suggestion you made, looks good to me. I've also capitalized the title and subtitle.

moollaza:
@sekhavati Thanks! This LGTM :+1:

PR #2455, opened 2016-01-31
Finish support for Holiday IA

marcantonio:
90% of the work was done here:  https://github.com/duckduckgo/zeroclickinfo-spice/pull/1701.

screen shot 2016-01-31 at 8 54 59 pm

IA page: https://duck.co/ia/view/holiday

### DIFF
--- a/lib/DDG/Spice/Holiday.pm
+++ b/lib/DDG/Spice/Holiday.pm
@@ -67,6 +67,13 @@ handle query_lc => sub {
     $q =~ s/\///g;
     $c =~ s/\///g;
 
+    # Translate holidays that timeanddate.com doesn't understand.
+    my %fixups = (
+        "mardi gras" => "shrove tuesday",
+    );
+
+    map { $q =~ s/$_/$fixups{$_}/ } keys %fixups;
+
     return $c, $q, $y;
 };
 
--- a/lib/DDG/Spice/Holiday.pm
+++ b/lib/DDG/Spice/Holiday.pm
@@ -18,16 +18,16 @@ attribution github => ['https://github.com/iambibhas', 'Bibhas'],
 
 triggers start => 'when is', 'when was', 'what day is', 'what day was';
 
-spice from => '([^/]+)/([^/]+)/([^/]+)';
-spice to => 'http://www.timeanddate.com/scripts/ddg.php?m=whenis&c=$1&q=$2&y=$3&callback={{callback}}';
+spice from => '([^/]+)/([^/]+)/([^/]+)/([^/]+)';
+spice to => 'http://www.timeanddate.com/scripts/ddg.php?m=whenis&c=$2&q=$3&y=$4&callback={{callback}}';
 
 handle query_lc => sub {
     return unless ($_);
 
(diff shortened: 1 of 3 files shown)


daxtheduck:
@marcantonio Hey!
        
Name:
Street Address:
City:
State / Province:
Country:
Phone (Non-US Orders only): 
T-shirt size:
GitHub username:
Link to this Pull Request:

Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

marcantonio:
Also, I noticed this weird visual anomaly:

screen shot 2016-01-30 at 9 28 52 pm

marcantonio:
Another issue:  searching for `when was` and the holiday hasn't happened yet this year returns a future event.

screen shot 2016-01-30 at 9 41 39 pm

moollaza:
> Also, I noticed this weird visual anomaly:

@marcantonio yup that’s a known issue. Its due to the way we're loading & concatenating the CSS. I've filed an issue in the DuckPAN repo: https://github.com/duckduckgo/p5-app-duckpan/issues/310

> searching for when was and the holiday hasn't happened yet this year returns a future event.

hmm we have some logic that searches for the holiday in the current year, and uses that date. The problem is that we don't know if a holiday has occurred until we get the first API response back, and if it has but was in the previous year, we'd need to make a 2nd API call.

I guess ideally the API should give us the previous and next occurrence so we can figure out what to show based on the query.

marcantonio:
> hmm we have some logic that searches for the holiday in the current year, and uses that date. The problem is that we don't know if a holiday has occurred until we get the first API response back, and if it has but was in the previous year, we'd need to make a 2nd API call.

> I guess ideally the API should give us the previous and next occurrence so we can figure out what to show based on the query.

Yeah I saw.  You're right, the best way would be if the API multiple dates or if we could specify multiple years in the query.  In leu of that, do spices ever make more than 1 API call?  We could either make 2 right off the bat (one for the current year and one for last year) or make a second call after the first one returns the wrong date.  Is there precedence for this?

moollaza:
> We could either make 2 right off the bat (one for the current year and one for last year) or make a second call after the first one returns the wrong date. Is there precedence for this?

That's certainly possible. This IA actually already does that when we get no results for the user's country: 

https://github.com/duckduckgo/zeroclickinfo-spice/pull/2455/files#diff-b6a9a492539f7580c055121fa2c66c9dR46

We form a new query and hit the NGINX endpoint again

marcantonio:
Ah, I missed that.  I don't think that code will every actually be hit though, because of L4.  I could be wrong, haven't done any JS in a while.  It seems that L45-49 should be checked in the if block at L4.

marcantonio:
US fallback works now.  Query was missing from the second call too.

marcantonio:
is/was now works for all cases.  Needs more thorough testing.

marcantonio:
@moollaza, I think this is ready for testing.

jagtalon:
Yus! I'm excited for this

PR #2525, opened 2016-02-20
Fixes #2303: Title will be visible without having to click on the poster

### LINKED ISSUE #2303: Anime: Title should be visible without having to click on the poster
@iambibhas I think the title should be visible directly in the instant answer. Sometimes, it's not easy to see on the poster or is in Japanese. Having to click all the posters to find the correct title is annoying.

Maybe display the poster as cards, like many other Instant Answers, with the poster on top of the card and the title underneath, with maybe the year as a subtitle ?

---

IA Page:  http://duck.co/ia/view/anime

lerna:
Fixes #2303: Title will now be visible without having to click on the poster.  Following suggestion from @moollaza, tileBody class is now overwritten to empty string instead of 'isHidden'.  This allows for both title and rating to be visible.  Ellipsis is used to indicate there is overflowed text in longer titles that is then visible upon hover.

IA Page: http://duck.co/ia/view/anime

### DIFF
--- a/lib/DDG/Spice/Xkcd/Display.pm
+++ b/lib/DDG/Spice/Xkcd/Display.pm
@@ -7,11 +7,13 @@ use DDG::Spice;
 triggers startend => "xkcd";
 
 spice to => 'http://xkcd.com/$1/info.0.json';
+spice proxy_cache_valid => '200 1h';
 spice wrap_jsonp_callback => 1;
 
 spice alt_to => {
 	latest => {
-		to => 'http://xkcd.com/info.0.json'
+        	to => 'http://xkcd.com/info.0.json',
+        	proxy_cache_valid => '200 1h'
 	}
 };
 


daxtheduck:
## [Anime](https://duck.co/ia/view/anime)

**Description**: Search for Anime

**Example Query**: [naruto anime](https://beta.duckduckgo.com/?q=naruto anime), [](https://beta.duckduckgo.com/?q=)

**Tab Name**: Anime

**Source**:  

*These are the important fields from the IA page.  Please check these for errors or missing information and update the [IA page](https://duck.co/ia/view/anime)*

---
*This is an automated message which will be updated as changes are made to the [IA page](https://duck.co/ia/view/anime)*

jagtalon:
@lerna cool! is the CSS for `.tile__title--2:hover` still needed? Seems like we can do without it

lerna:
It is there for titles abbreviating to the same string with ellipsis (i noticed some do) for users to hover over and differentiate between those type of titles quickly without having to click on each title.  If that feauture is not needed i can remove it.

PR #2534, opened 2016-02-25
New JustDeleteMe Spice

riqpe:
https://duck.co/ia/view/just_delete_me

screen shot 2016-02-29 at 20 25 15
screen shot 2016-02-29 at 20 23 52
screen shot 2016-02-29 at 20 31 59
screen shot 2016-02-29 at 20 32 48
screen shot 2016-02-29 at 20 24 49
screen shot 2016-02-29 at 20 24 32
screen shot 2016-02-29 at 20 31 36
screen shot 2016-02-29 at 20 31 48


daxtheduck:
@riqpe Hey!
        
Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/spice/just_delete_me/just_delete_me.js
```diff
@@ -0,0 +1,67 @@
...
+                group: 'text',
+                options: {
+                    footer: Spice.just_delete_me.footer,
+                    content: Spice.just_delete_me.content,
+                    moreAt: true,
+                    aux: false
+                },
+                variants: {
+                    tileTitle: '1line-large',
+                    tileFooter: '2line'
```
moollaza: Let's add `tileSnippet: 'small'` to force the tiles to have the same height

share/spice/just_delete_me/just_delete_me.js
```diff
@@ -0,0 +1,67 @@
...
+        var script = $('[src*="/js/spice/just_delete_me/"]')[0],
+            source = $(script).attr("src"),
+            query = source.match(/just_delete_me\/([^\/]+)/)[1],
+            decodedQuery = decodeURIComponent(query).split(" ")[0].toLowerCase();
+
+        if (!decodedQuery || decodedQuery.length < 4) {
+            return Spice.failed('just_delete_me');
+        }
+
+        var exact = api_result.find(function(item) { return item.name.toLowerCase() === decodedQuery}) ||
```
moollaza: It looks like `array.find()` isn't supported by IE, Opera, etc. We need to use methods that work across all browsers.

I would suggest you write this as a loop: `$.foreach(api_result, function(item) {...});`

share/spice/just_delete_me/just_delete_me.js
riqpe: I suggest we go with large, as the empty tiles will look just as empty as small but with large we won't have to tinker with the condensing of texts as it works correctly already without a custom style.
screen shot 2016-02-25 at 20 05 04

share/spice/just_delete_me/just_delete_me.js
riqpe: `array.filter()` is supported and works fine here too.

t/JustDeleteMe.t
```diff
@@ -0,0 +1,45 @@
...
+        '/js/spice/just_delete_me/cloud',
+        call_type => 'include',
+        caller => 'DDG::Spice::JustDeleteMe'
+    ),
+    'cancel cloud account!' => test_spice(
+        '/js/spice/just_delete_me/cloud',
+        call_type => 'include',
+        caller => 'DDG::Spice::JustDeleteMe'
+    )
+);
```
MrChrisW: @riqpe Can you add some query examples that **should not trigger** this IA? e.g `undef`

riqpe: Sure thing. See commit 202e09b.

MrChrisW: :+1:

moollaza:
Deployed to Beta for testing!

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=delete my facebook account)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/just_delete_me_mobile.png)

moollaza:
@riqpe this is looking great. There's some room for improvement but I think this is great for a V1. I'll leave some comments on what I'd like you to do next.

share/spice/just_delete_me/content.handlebars
```diff
@@ -0,0 +1,15 @@
+{{#if notes}}
+  <p>
+    {{notes}}
+  </p>
+{{/if}}
+{{#if delete_url}}
+  <p>
+    <b>URL:</b> <a href='{{delete_url}}'>{{delete_url}}</a>
```
moollaza: Please remove the `<b>` -- we don't typically bold text. If anything we use different colours

share/spice/just_delete_me/content.handlebars
```diff
@@ -0,0 +1,15 @@
+{{#if notes}}
+  <p>
+    {{notes}}
```
moollaza: I think it would be good to have a fallback message that explains you should follow the link.

Alternatively, maybe we shouldn't show a result if it has no instructions?

https://beta.duckduckgo.com/?q=delete+instagram+account&t=h&ia=answer -- it's not entirely clear what's supposed to happen here.

share/spice/just_delete_me/content.handlebars
```diff
@@ -0,0 +1,15 @@
...
+  </p>
+{{/if}}
+{{#if delete_url}}
+  <p>
+    <b>URL:</b> <a href='{{delete_url}}'>{{delete_url}}</a>
+  </p>
+{{/if}}
+{{#if email}}
+  <p>
+    <b>Email:</b> <a href='mailto:{{email}}'>{{email}}</a>
```
moollaza: Remove the `<b>`

Is the user supposed to send an email here requesting their account to be deleted? It's unclear what this email address is for...

share/spice/just_delete_me/footer.handlebars
```diff
@@ -0,0 +1,10 @@
+{{#if delete_url}}
+    <div class="one-line">
+         <a href='{{delete_url}}'>{{delete_url}}</a>
```
moollaza: Let's show descriptive text instead of these URLs.

I think the footer should look like this:

What are your thoughts?

how_to_delete_cloud_account_at_duckduckgo

share/spice/just_delete_me/just_delete_me.js
```diff
@@ -0,0 +1,69 @@
...
+            query = source.match(/just_delete_me\/([^\/]+)/)[1],
+            decodedQuery = decodeURIComponent(query).split(" ")[0].toLowerCase();
+
+        if (!decodedQuery || decodedQuery.length < 4) {
+            return Spice.failed('just_delete_me');
+        }
+
+        var exact = api_result.filter(function(item) { return item.name.toLowerCase() === decodedQuery}),
+            exact = exact.length > 0 ? [exact[0]] : 
+                    api_result.filter(function(item) { return item.domains &&
```
moollaza: I think these filtering functions should be defined above, and referenced here. These are hard to read/understand and I'm not a fan of the double ternary.

moollaza: If you're checking the relevancy of the items we have the `DDG.isRelevant()` and `DDG.stringsRelevant()` functions which are a little easier to use.

I noticed https://beta.duckduckgo.com/?q=delete+icloud+account&t=h&ia=answer gives an incorrect result which needs to be addressed

share/spice/just_delete_me/just_delete_me.js
```diff
@@ -0,0 +1,69 @@
...
+            meta: {
+                sourceName: 'Just Delete Me',
+                sourceUrl: "http://justdelete.me/#" + decodedQuery
+            },
+            data: api_result,
+            normalize: function(item) {
+                return {
+                    delete_url: item.url,
+                    title: item.name,
+                    url: "http://justdelete.me/#" + decodedQuery,
```
moollaza: I would expect the tile to go to the account deletion page (`item.url`), not www.justdelete.me

lib/DDG/Spice/JustDeleteMe.pm
```diff
@@ -0,0 +1,48 @@
...
+        @required = (@required, grep(/^($req[$i]\W*)$/i, @qparts));
+    }
+    #strip required words and return rest as string
+    if (scalar(@required) > 0) {
+        for (my $i = 0; $i<scalar(@req); $i++) {
+            @qparts = grep(!/^($req[$i]\W*)$/i, @qparts);
+        }
+        return join(" ", @qparts);
+    }
+    return "";
```
moollaza: this needs to be `return;` which is the same as `return {}`.

`return ""` is considered a valid return and will issue an API and trigger your Instant Answer

moollaza: My mistake, I realized you have a `return;` above!

lib/DDG/Spice/JustDeleteMe.pm
```diff
@@ -0,0 +1,48 @@
...
+    my $query = validate(@_);
+    if (length($query) < $MIN_QUERY_LENGTH) {
+        return;
+    }
+    else {
+        return $query;
+    }
+};
+
+sub validate {
```
moollaza: I suspect we can probably replace this functionality with one or two regular expressions?

Something like this should cover most of the queries?

```perl
/how (to|can|can (i|one))? (delete|remove|cancel) (his|her|their)? (\w+) accounts? (permanently)?/;
```

Note: I haven't tested this, but I suggest you look at the autocomplete suggestions to see which forms of queries are most popular "delete X account" seems very common :+1:

share/spice/just_delete_me/content.handlebars
riqpe: Alright, will remove.

share/spice/just_delete_me/content.handlebars
riqpe: I was following the convention on the justdeleteme page, which as you say, can be a bit vague at times.

share/spice/just_delete_me/content.handlebars
riqpe: Yes, the email is provided from justdeleteme for requesting account deletion. I could possibly have fallback notes, one for email, other for url, though problem is what to do when there are both.

share/spice/just_delete_me/footer.handlebars
riqpe: I agree on the small tiles it makes sense to just have the descriptive text as you say, though the link might not be directly to the deletion page, as can be seen on IRCCloud on your screenshot.

share/spice/just_delete_me/just_delete_me.js
riqpe: As I mentioned before, the DDG.isRelevant() and DDG.stringsRelevant() does not really return what is expected here, returning too many false positives IMO.

The result for your query on iCloud is correct. There is no entry for iCloud on JustDeleteMe, though there is one for JoliCloud as you noted, which matches. See http://justdelete.me/#icloud

share/spice/just_delete_me/just_delete_me.js
riqpe: I have to disagree here. As this is linked to from the name I would assume to see more details, especially as we will have the deletion URL listed separately, and clearly. Some pages also do not require any extra checks to delete your account, thus just clicking on the link will remove your account, if you are logged in. I think we should keep the item.name linking to more details on JustDeleteMe and keep the deletion URL separated.

lib/DDG/Spice/JustDeleteMe.pm
riqpe: I'll look into that.

PR #2536, opened 2016-02-25
New Public Holidays Spice

sekhavati:
_(Work in progress - see limitations)_

# Overview
This is a new IA that provides the list of public holidays in a calendar year for a country.

# Preview
For now I'm using the `List` template but I'd welcome input from the design team.

preview

# Triggers
The following words are triggers for this IA:
- `public holidays`
- `national holidays`
- `bank holidays`
- `federal holidays`

The trigger words can appear anywhere in the query but are carefully inspected to ensure they are matched in the right context before triggering the IA. Queries are inspected for the following components:

Component| Description| Required | Default | Optional Prefix
--------------|----------------|-------------|----------|---------------------
 `trigger`    | one of the keywords listed above | Yes |  - | -
`country`   |  the country for which holidays to retrieve | No | `$loc` | `in | in the`
`year`       | the year for which the holidays should be retrieved | No | Current Year | `in`

The components can be combined in various ways to form a query that this IA responds to, for example:

`trigger` `country` `year` => `public holidays` `uk` `2016`
`country` `trigger` `year` => `uk` `public holidays` `2016`
`country` `trigger` => `uk` `public holidays`
etc...

The IA will not trigger unless one of the possible variations is matched exactly (start to end of the query).

# Outstanding Issues

1. The UI layout doesn't fit on mobile devices
2. Some public holidays are not listed, for example St.Patricks day in Northern Ireland. We need to request an additional holiday type from the API but first some more investigation is required to ensure this extra set only includes genuine public holidays for _all_ countries

# Future Improvements
These are some ideas for how I might improve the IA in the future:

- Only show holidays that have yet to come by default; clicking `Show More` would display holidays that have passed
- The API response includes a URL to a page containing details about each specific holiday, what's it for, what do people do traditionally etc. It could be useful/interesting to include links to these pages

---

### DIFF
--- a/lib/DDG/Spice/PublicHolidays.pm
+++ b/lib/DDG/Spice/PublicHolidays.pm
@@ -5,7 +5,6 @@ package DDG::Spice::PublicHolidays;
 use strict;
 use DDG::Spice;
 use Locale::Country;
-use Text::Trim;
 
 spice is_cached => 1;
 spice proxy_cache_valid => "200 30d";
--- a/lib/DDG/Spice/PublicHolidays.pm
+++ b/lib/DDG/Spice/PublicHolidays.pm
@@ -11,7 +11,7 @@ spice proxy_cache_valid => "200 30d";
 spice wrap_jsonp_callback => 0;
 
 spice from => "(.+)/(.+)";
-spice to => 'https://api.xmltime.com/holidays?accesskey={{ENV{DDG_SPICE_TIME_AND_DATE_ACCESSKEY}}}&secretkey={{ENV{DDG_SPICE_TIME_AND_DATE_SECRETKEY}}}&version=2&callback={{callback}}&types=federal&country=$1&year=$2';
+spice to => 'https://api.xmltime.com/holidays?accesskey={{ENV{DDG_SPICE_TIME_AND_DATE_ACCESSKEY}}}&secretkey={{ENV{DDG_SPICE_TIME_AND_DATE_SECRETKEY}}}&version=2&callback={{callback}}&types=federal%2Cfederallocal&country=$1&year=$2';
 
 my @triggers  = ('public holidays', 'national holidays', 'bank holidays', 'federal holidays');
 my $triggers  = join('|', @triggers);
--- a/lib/DDG/Spice/PublicHolidays.pm
+++ b/lib/DDG/Spice/PublicHolidays.pm
@@ -1,0 +1,121 @@
+package DDG::Spice::PublicHolidays;
+
+# ABSTRACT: Provides the list of public holidays in a calendar year for a country. 
+
+use strict;
+use DDG::Spice;
+use Locale::Country;
+
+spice is_cached => 1;
+spice proxy_cache_valid => "200 30d";
+spice wrap_jsonp_callback => 0;
+
+spice from => "(.+)/(.+)";
+spice to => 'https://api.xmltime.com/holidays?accesskey={{ENV{DDG_SPICE_TIME_AND_DATE_ACCESSKEY}}}&secretkey={{ENV{DDG_SPICE_TIME_AND_DATE_SECRETKEY}}}&version=2&callback={{callback}}&types=federal&country=$1&year=$2';
+
+my @triggers  = ('public holidays', 'national holidays', 'bank holidays', 'federal holidays');
+my $triggers  = join('|', @triggers);
+my $countries = join('|', share("countries.txt")->slurp(chomp => 1));
+
+triggers any => @triggers;
+    
+handle query_lc => sub {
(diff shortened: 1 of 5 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

lib/DDG/Spice/PublicHolidays.pm
```diff
@@ -0,0 +1,121 @@
...
+    my $year     = qr/[1-9]{1}[0-9]{3}/;
+    
+    # Regexes to ignore optional words that can precede the year or country name
+    my $in       = qr/(?:in )?/;
+    my $inThe    = qr/(?:in the |in )?/;
+    
+    my $chosenYear;
+    my $chosenCountry;
+    
+    # Match: <trigger>
```
moollaza: This is good, but it's a little too verbose and I believe there's a simpler approach.

We already know one of the triggers exists in the query and we don't care to know which it is or where it is. So immediately we can strip it out to simplify our query. e.g. `$query =~ s/$triggers//;`

If there's a year in our query we can easy capture that with a simple regex `\d{4}` and then remove it from our query.

If we assume the queries only match our expected pattern and the query still has a word, we can check if it exists in our country hash. If the query has multiple remaining words we can check if any one of them is a country.

An approach like this should be simpler to implement and easier to maintain or improve

moollaza: The Quandl IA's use a similar approach. Likewise Quixey handles things sorta like this.

sekhavati: @moollaza 

I can have a look at simplifying things in the way you suggested but I'm concious of over-triggering. With the approach you suggested whenever a trigger, country and year appear in the query this IA will fire, regardless of the composition of the query or any extra words which may alter the meaning. My approach thus far was to strictly only respond when I could be certain the query entered was relevant.

There's also some edge cases that need more consideration. Perhaps we could talk more offline?

moollaza: >With the approach you suggested whenever a trigger, country and year appear in the query this IA will fire, regardless of the composition of the query or any extra words which may alter the meaning.

@sekhavati you're correct. If we remove any elements we find from the query and the query still has a length then we could assume it's irrelevant. 

We should at least be able to cut your approach down to fewer compounded regexes by using `|` and `?`:

E.g. `$query =~ /^$holidays $inThe($country)$/` + `/^$holidays $in($year)$/` = `/^$holidays (?:$in($year)|$inThe($country))$/`

> Perhaps we could talk more offline?
 
Not a problem! Send me a message on Slack and we can chat

share/spice/public_holidays/public_holidays.js
```diff
@@ -0,0 +1,54 @@
...
+        
+        DDG.require('moment.js', function() {
+            Spice.add({
+                id: "public_holidays",            
+                name: "Answer",            
+                data: api_result,
+
+                meta: {
+                    sourceName: "timeanddate.com",
+                    sourceUrl: 'http://www.timeanddate.com/holidays/'
```
moollaza: This should link to a page related to the query, i.e. holidays for the searched country (and year if possible)

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=Public Holidays)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/public_holidays_mobile.png)

share/spice/public_holidays/public_holidays.js
sekhavati: @moollaza fixed in e49dbb00ee7f20fd0c318149147598b9181b1027

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=Public Holidays)
                
                [Mobile screenshot](https://ia-screenshots.s3.amazonaws.com/public_holidays_mobile.png)

share/spice/public_holidays/public_holidays.js
```diff
@@ -0,0 +1,64 @@
...
+            });
+            
+            // holiday.urlid is of the form "<country>/<holiday>"
+            var countryUrlId = api_result.holidays[0].urlid.match(/(.+)\/.*/)[1];            
+            
+            Spice.add({
+                id: "public_holidays",            
+                name: "Answer",
+                data: data,
+
```
moollaza: @sekhavati let's add `signal: "high",` -- this will force it to open this IA's tab over others'

moollaza:
@sekhavati Thanks for the excellent work! This LGTM :+1:

PR #2561, opened 2016-03-08
Fix for issue #1529

### LINKED ISSUE #1529: Stocks: Set low signal for dictionary word tickers
For example, if you search for `guid` or `IRC`, you'll get stock quotes--not the obvious things that you would expect that come from other IAs. This is because the Stocks spice triggers if the query is a stock symbol, regardless of whether it's actually a trigger keyword for another spice or other kind of IA.

One way to possibly address this is to consider the list of IA triggers and, if one of them matches, prioritize that IA above the Stocks IA.

------ 

IA Page: https://duck.co/ia/view/stocks

gvanwaelvelde:
When the search query doesn't contain the word 'stock' the signal is set to low. This should prevent this Instant Answer to fire when not searching for stock prices.

Fixes #1529

https://duck.co/ia/view/stocks

### DIFF
--- a/share/spice/stocks/stocks.js
+++ b/share/spice/stocks/stocks.js
@@ -13,7 +13,14 @@
                 id: 'stocks',
                 name: 'Stock',
                 data: api_result,
-                signal: (DDG.get_query().search('stock')) ? 'low' : 'high', 
+                signal: function (){
+                   var exceptions = ['irc', 'guid']
+                      for each (var x in exceptions)
+                        if (!DDG.get_query().search(x)){
+                            return 'low';
+                        }
+                            return 'high';
+                        }, 
                 meta: {
                     sourceName: 'NASDAQ',
                     sourceUrl: url
--- a/share/spice/stocks/stocks.js
+++ b/share/spice/stocks/stocks.js
@@ -13,6 +13,7 @@
                 id: 'stocks',
                 name: 'Stock',
                 data: api_result,
+                signal: (DDG.get_query().search('stock')) ? 'low' : 'high', 
                 meta: {
                     sourceName: 'NASDAQ',
                     sourceUrl: url


daxtheduck:
@gvanwaelvelde Hey!
        
Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
@gvanwaelvelde thanks! Deploying this to beta...

moollaza:
@tommytommytommy it seems like Deep is overriding the signal set in the JS? Do you know if there's any way we can avoid that?

moollaza:
> When the search query doesn't contain the word 'stock' the signal is set to low.

@gvanwaelvelde to clarify, we have a special internal trigger that triggers this Instant Answer when certain organic links appear (e.g. Yahoo Finance) which suggests to us that the query is for a stock ticker. However not every query that has a stock ticker is meant to be a search for that. IRC and GUID are good examples. I think the way we'll want to handle this is by manually grouping a list of those terms and forcing a low signal in only those cases. If I search `aapl` which is the stock symbol for Apple, the Stocks IA is super relevant and deserves a high signal

gvanwaelvelde:
@moollaza I have made some changes. If the search query equals a word that's in a list of exceptions, then the signal gets set to low. 
If this works, we may need to expand the list to contain more exceptions.

moollaza:
> I have made some changes. If the search query equals a word that's in a list of exceptions, then the signal gets set to low. 

@gvanwaelvelde I've spoken with the team about this issue. It turns out they have bigger, better plans for fixing this because we've noticed it triggers on several other "dictionary" words, producing unexpected results, like https://duckduckgo.com/?q=wifi&ia=stock.

Seeing as they'll be addressing this internally, I'm going to close this out.

> If this works, we may need to expand the list to contain more exceptions.

This is also one of the reasons we've decide to tackle this internally -- it turns out there are many dictionary words that are also valid Stock Symbols and it would likely be too many to handle in the JS. As well, if we tackle it internally where we have our special trigger, we can prevent it from ever making the API call, instead of rejecting the data after the fact.

Thanks again for taking the time to help, and sorry we can't use this code!

I had a look and it seems like there is another issue open which can definitely be fixed right now, in the open source code: https://github.com/duckduckgo/zeroclickinfo-goodies/issues/1046

Would you be interested in looking at that?

moollaza:
@gvanwaelvelde I forgot to mention that it turns out changing the signal in the JS doesn't even work right now because the internal code is forcing a high signal, which is another reason why this needs to be fixed internally

PR #2588, opened 2016-03-15
Public Holidays: Add Local Holidays to the list of holiday types retrieved

sekhavati:
This change extends the types of holiday retrieved from the API to include `Local Holidays`. The reason for this change is because it was noted some holidays in the UK weren't showing up, such as St. Patricks Day (Northern Ireland only) and St. Andrews Day (Scotland only) which are [official holidays](https://www.gov.uk/bank-holidays) in those regions. The list of extra holidays this type includes for the UK can be seen [here](http://www.timeanddate.com/holidays/uk/#!hol=2).

**Note**: this IA caches its results for 30 days thus it will need to be flushed when this change goes live to see the difference (if it doesn't do already).

---

### DIFF
--- a/lib/DDG/Spice/PublicHolidays.pm
+++ b/lib/DDG/Spice/PublicHolidays.pm
@@ -11,7 +11,7 @@ spice proxy_cache_valid => "200 30d";
 spice wrap_jsonp_callback => 0;
 
 spice from => "(.+)\/(.+)\/.+";
-spice to => 'https://api.xmltime.com/holidays?accesskey={{ENV{DDG_SPICE_TIME_AND_DATE_ACCESSKEY}}}&secretkey={{ENV{DDG_SPICE_TIME_AND_DATE_SECRETKEY}}}&version=2&callback={{callback}}&types=federal%2Cfederallocal&country=$1&year=$2';
+spice to => 'https://api.xmltime.com/holidays?accesskey={{ENV{DDG_SPICE_TIME_AND_DATE_ACCESSKEY}}}&secretkey={{ENV{DDG_SPICE_TIME_AND_DATE_SECRETKEY}}}&version=2&callback={{callback}}&types=federal%2Cfederallocal%2Clocal&country=$1&year=$2';
 
 my @triggers  = ('public holidays', 'national holidays', 'bank holidays', 'federal holidays');
 my $triggers  = join('|', @triggers);


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
Deployed to Beta for testing!

moollaza:
@sekhavati https://beta.duckduckgo.com/?q=public+holidays+in+uk&ia=answer&iax=1 looks good! Are there any other examples I should check out?

sekhavati:
@moollaza Canada is one that I know of that will display additional holidays as a result of this change. At the weekend if I get time I'll write a little utility script to help identify exactly which countries are affected. In the mean time if your inclined to do so you can use [this link](http://www.timeanddate.com/holidays/uk/#!hol=0) and select the Country and Holiday Types manually, for reference this IA now includes the following types:

- National Holidays (federal)
- Common Local (federallocal)
- Local Holidays (local)

moollaza:
> At the weekend if I get time I'll write a little utility script to help identify exactly which countries are affected. In the mean time if your inclined to do so you can use this link and select the Country and Holiday Types manually, for reference this IA now includes the following types:

Perfect, thanks!

I don't think a script will be necessary, I just wanted some examples to test :+1:

sekhavati:
@moollaza  On reflection I'm a little disconcerted by how many extra holidays this new type pulls in for *some* countries. For the United States there is a significant increase, compare the links below. 

Before: https://duckduckgo.com/?q=public+holidays+us&t=ffsb&ia=answer&iax=1
After: https://beta.duckduckgo.com/?q=public+holidays+us&ia=answer&iax=1

The three holiday types selected are correct for the UK though and is the only combination that retrieves all relevant public holidays for that country.

moollaza:
@sekhavati I have an idea.. we could add a link/button to the subtitle that said "include local events" and either do a requery for "federal and local holidays in UK 2016", or we could add some JS that makes another API call and re-renders the data, or potentially just unhides some hidden cells.

Those are all relatively complex though, so we could also just wait and see if anyone asks for more results and then we can revisit this.

Thoughts?

sekhavati:
@moollaza Apologies for the late reply. We've just had Easter in the UK so I took a few days off to relax.

There are a couple of questions that need answering:

_1. Is this new combination of holiday types the correct set?_
   - It is for the UK but I can't really speak for others
   - If anyone from the US could confirm these extra holidays are indeed correct for those states that would make me more confident we have the right combination (see links earlier)

_2. How do we display this information nicely to end users?_

- I think your suggestion re.design is heading in the direction we may need to go ultimately, especially if it turns out this third holiday type in indeed required
- Similar to what you suggested, I've been pondering on the idea of separating the holidays out into two tables, one which displays holidays applicable to everywhere within the country, and another below it with say a combo-box where a user could select to view state specific holidays
- Perhaps a few mock-ups would help visualise a few ways to approach this? I could do us a few

moollaza:
@sekhavati I've thought about this a bit more. Enabling/disabling the functionality is trivial, so if we need it we can easily add it and similarly we can very easily remove the functionality.

We can also very easily adda check for `local` in the query, or a hash of countries that should show `local` holidays by default (i.e. uk) and only show them according to those rules.

I think the multi-table approach sounds cool but is likely to be more complex than we need or should have. The calendar view the design team had envisioned should alleviate the need for a complex UI.

Seeing as all the data is initially suppressed it's just a matter of how much info is too much.

I think we'll be fine to enable this and if anyone complains we can adjust it so that local is only included for certain queries or countries.

PR #2589, opened 2016-03-15
Public Holidays: Ensure cell data is wrapped on small screens

sekhavati:
Some holidays only apply to certain regions within a country, in these instances the IA displays a third column clarifying where it is observed. On small screens (ie: mobile devices) this third column can appear clipped, I've attached some before/after screenshots below from an iPhone 5. The end result isn't an ideal layout, but is an improvement.

#### Before (Collapsed)
before_collapsed

#### After (Collapsed)
after_collapsed

#### Before (Expanded)
before_expanded

#### After (Expanded)
after_expanded

---

### DIFF
--- a/share/spice/public_holidays/record.handlebars
+++ b/share/spice/public_holidays/record.handlebars
@@ -1,7 +1,7 @@
-<div class="record  {{#if meta.options.keySpacing}}record--keyspacing{{/if}}  {{#if meta.options.rowHighlight}}record--highlight{{/if}}">
+<div class="record">
     <table class="record__body">
         {{#each holidays}}
-            <tr class="record__row  {{#if ../meta.options.rowHighlight}}record__row--highlight{{/if}}">
+            <tr class="record__row">
                 <td class="record__cell  record__cell--key">{{date}}</td>                
                 <td class="record__cell  record__cell--value">{{name}}</td>
                 <td class="record__cell  record__cell--value">{{notes}}</td>
--- a/share/spice/public_holidays/record.handlebars
+++ b/share/spice/public_holidays/record.handlebars
@@ -3,8 +3,8 @@
         {{#each holidays}}
             <tr class="record__row  {{#if ../meta.options.rowHighlight}}record__row--highlight{{/if}}">
                 <td class="record__cell  record__cell--key">{{date}}</td>                
-                <td class="record__cell  record__cell--key">{{name}}</td>                
-                <td class="record__cell  record__cell--key">{{notes}}</td>                
+                <td class="record__cell  record__cell--value">{{name}}</td>
+                <td class="record__cell  record__cell--value">{{notes}}</td>
             </tr>
         {{/each}}
     </table>


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/spice/public_holidays/record.handlebars
```diff
@@ -3,8 +3,8 @@
         {{#each holidays}}
             <tr class="record__row  {{#if ../meta.options.rowHighlight}}record__row--highlight{{/if}}">
                 <td class="record__cell  record__cell--key">{{date}}</td>                
-                <td class="record__cell  record__cell--key">{{name}}</td>                
-                <td class="record__cell  record__cell--key">{{notes}}</td>                
+                <td class="record__cell  record__cell--value">{{name}}</td>
+                <td class="record__cell  record__cell--value">{{notes}}</td>
```
moollaza: @sekhavati to make the most use of space, we should only put in this cell if `notes` is defined. Otherwise we have a blank 3rd column for the majority of results:

```
{{#if notes}}
<td class="record__cell  record__cell--value">{{notes}}</td>
{{/if}}
```

share/spice/public_holidays/record.handlebars
```diff
@@ -3,8 +3,8 @@
         {{#each holidays}}
             <tr class="record__row  {{#if ../meta.options.rowHighlight}}record__row--highlight{{/if}}">
```
moollaza: We can remove the `{{if}}...` here -- that's part of the template and we don't need it here

moollaza: We can also remove the `{{#if}}` above -- they're unnecessary

share/spice/public_holidays/record.handlebars
sekhavati: @moollaza Tried your suggestion but the layout is a little inconsistent afterwards

optional-notes

share/spice/public_holidays/record.handlebars
sekhavati: @moollaza 3740ce041daf4a78696b29ed148a63e176001d8e should address this

share/spice/public_holidays/record.handlebars
moollaza: hmm that's odd -- the table should be filling the width there and letting the 2nd column stretch out.

What browser is that?

sekhavati: @moollaza Firefox on Ubuntu most likely, possibly Windows 7. I can't remember which of my machines I was on at the time.

moollaza:
@sekhavati looks like this has a conflict now -- would you mind updating it?

Id like to test my suggestion to see which browsers/os it breaks on

sekhavati:
@moollaza I've made a commit that should resolve the conflict. I'm not that verse with `git` so you may want to check the change history is as expected; I used the first method documented [here](http://stackoverflow.com/a/28505516).

jdorweiler:
Could this IA use the record template?  It looks like the template this currently uses is just adding an additional row?
https://github.com/duckduckgo/duckduckgo-answerbar-templates/blob/master/shared/record.handlebars

sekhavati:
@jdorweiler It can't use the standard record template because it only has support for two columns of data. This IA extends it for a third column as shown in the screenshots above.

jdorweiler:
@sekhavati would you want to try adding that functionality to the template instead? 

template: https://github.com/duckduckgo/duckduckgo-answerbar-templates/blob/master/shared/record.handlebars
helper that builds the data: https://github.com/duckduckgo/duckduckgo-template-helpers/blob/master/template_helpers.js#L832

sekhavati:
@jdorweiler I'm not sure its fit for purpose, generally speaking. The design/layout is questionable in my opinion but it was the best fit from the templates that existed.

jdorweiler:
@sekhavati sounds good.  Thanks for fixing this :+1:

PR #2597, opened 2016-03-17
Add new Instant Answer for CVE information

MoriTanosuke:
After a nice hint from @moollaza over at https://github.com/duckduckgo/zeroclickinfo-longtail/pull/50 I recreated my new IA to display CVE summarys into a Spice.

This PR creates the new IA (I also created https://duck.co/ia/view/cve_summary) to display CVE information from circl.lu. I also set the source URL to the details page of the searched CVE (http://cve.circl.lu/api/cve/CVE-2016-0815).

The API returns JSON like the following:

````
{
  "Modified": "2016-03-12T16:59:01.523-05:00",
  "Published": "2016-03-12T16:59:00.117-05:00",
  "cvss": null,
  "cwe": "Unknown",
  "id": "CVE-2016-0815",
  "last-modified": "2016-03-12T16:59:01.523-05:00",
  "map_cve_scip": {
    "scipid": "81288",
    "sciplink": "http://www.scip.ch/?vuldb.81288"
  },
  "references": [
    "https://android.googlesource.com/platform%2Fframeworks%2Fav/+/5403587a74aee2fb57076528c3927851531c8afb",
    "http://source.android.com/security/bulletin/2016-03-01.html"
  ],
  "summary": "The MPEG4Source::fragmentedRead function in MPEG4Extractor.cpp in libstagefright in mediaserver in Android 4.x before 4.4.4, 5.x before 5.1.1 LMY49H, and 6.x before 2016-03-01 allows remote attackers to execute arbitrary code or cause a denial of service (memory corruption) via a crafted media file, aka internal bug 26365349.",
  "vulnerable_configuration": [],
  "vulnerable_configuration_cpe_2_2": []
}
````

The IA then displays the *id* and the *summary*, the source URL is set to the page http://cve.circl.lu/cve/CVE-2016-0815 where the user can read more about the CVE and also finds links to the references.

If there is anything I need to fix or clean up on this PR, please give me a heads up and maybe a link to the appropriate documentation.

---
IA page: https://duck.co/ia/view/cve_summary

### CODE REVIEW

**Comment by daxtheduck** [comment]
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

File: lib/DDG/Spice/Cve_summary.pm [hunk 1]
```diff
@@ -0,0 +1,42 @@
+package DDG::Spice::Cve_summary;
```
moollaza: @MoriTanosuke Let's use a more canonical package name. I would suggest `CveSummary` because Perl uses CamelCase

lib/DDG/Spice/Cve_summary.pm
```diff
@@ -0,0 +1,42 @@
...
+# to instant answer development
+
+use DDG::Spice;
+
+spice is_cached => 1;
+# API url
+spice to => 'http://cve.circl.lu/api/cve/$1';
+spice wrap_jsonp_callback => 1;
+
+# Metadata.  See https://duck.co/duckduckhack/metadata for help in filling out this section.
```
moollaza: You can remove all this metadata -- this info now lives on the IA Page :+1:

lib/DDG/Spice/Cve_summary.pm
```diff
@@ -0,0 +1,42 @@
...
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#category
+category "reference";
+# Uncomment and complete: https://duck.co/duckduckhack/metadata#topics
+# topics "";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Cve_summary.pm";
+attribution github => ["GitHubAccount", "MoriTanosuke"],
+            twitter => "carsten_r";
+
+# Triggers
+triggers any => "cve";
```
moollaza: Initially I thought we'd want to use a regex trigger here, but seeing as we actually break query strings into "words" on non-alphanumeric characters, a query of "CVE-2016-0815" would actually trigger with `triggers start => "cve"`

I don't think we want to use `triggers any` though.

lib/DDG/Spice/Cve_summary.pm
```diff
@@ -0,0 +1,42 @@
...
+
+# Handle statement
+handle remainder => sub {
```
moollaza: let's `handle query_lc` and then make sure that the query is of the form `CVE-\d{4}-\d+`. You'll probably want to also strip out a leading `cve` if it exists.

Judging by the autocomplete the majority of searches are just `cve-####-####`

MoriTanosuke:
@moollaza I tried to make the changes you mentioned, but unfortunately I now don't get my data object into the JS file. I guess I created a name mismatch by renaming the perl module, but I'm too tired now to find the exact place. The trigger and the regex check seem to work though.

Will continue on this tomorrow evening. Thanks for the quick responses. :smile:

MoriTanosuke:
Ok, I found the issue. The API only returns a result if for upper case CVE like https://cve.circl.lu/api/cve/CVE-2016-3191. https://cve.circl.lu/api/cve/cve-2016-3191 will return an empty result.

I fixed this in my latest commit. Hope the code is ok now.

share/spice/cve_summary/cve_summary.js
```diff
@@ -0,0 +1,41 @@
...
+            templates: {
+                group: 'text',
+                options: {
+                    content: Spice.cve_summary.content,
+                    moreAt: true
+                }
+            },
+            normalize: function(item) {
+                return {
+                    title: api_result.id,
```
moollaza: @MoriTanosuke it looks like we have other useful info like the `name` of the issue etc, I think we should set our `title` to the CVE Name? and set the subtitle as the CVE code?

MoriTanosuke: Will take a look what kind of information is available for most of the CVE and check the documentation for appropriate properties to use. Thanks for your help.

moollaza:
@MoriTanosuke https://beta.duckduckgo.com/?q=cve-2009-355&ia=answer

It looks like this one doesn't have a summary!

In that case, we should instead say "No summary provided"

share/spice/cve_summary/cve_summary.js
```diff
@@ -0,0 +1,41 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_cve_summary = function(api_result){
+
+        // fail the spice if the API does not return a result
+        if(api_result == undefined) {
```
moollaza: @MoriTanosuke it looks like they actually return an empty object, not an undefined object.

You'll probably want to use http://api.jquery.com/jQuery.isEmptyObject/ to verify the object isn't empty!

MoriTanosuke:
I'll pick this up again next week. I'm moving this week, so not much time to hack on DDG...

MoriTanosuke:
@moollaza I set the *Published* date as subtitle. Other than that, I don't think that there is any useable information on all the CVEs.

I also changed the regex to only match on *CVE-XXXX-YYYY* with 4 digits in group *X* and *Y*. Your test with *cve-2009-355* did not have a result, but *cve-2009-0355* has: https://cve.circl.lu/api/cve/CVE-2009-0355

Please, review again. :smile:

share/spice/cve_summary/cve_summary.js
```diff
@@ -0,0 +1,42 @@
...
+                group: 'text',
+                options: {
+                    content: Spice.cve_summary.content,
+                    moreAt: true
+                }
+            },
+            normalize: function(item) {
+                return {
+                    title: api_result.id,
+                    subtitle: "Published " + api_result.Published,
```
moollaza: We should use moment.js to format this for legibility: e.g. `Tues, Jan 16, 2016`

moollaza:
```js
DDG.require('moment', function() {
    Spice.add({ ... });
});
```

moollaza:
@MoriTanosuke Thanks! This LGTM :+1:

daxtheduck:
@MoriTanosuke this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

PR #2601, opened 2016-03-19
New VRCTeams Spice

mattr555:
This IA fetches statistics for a given VEX Robotics Competition team. It displays them in the infobox along with information about their division, location, etc.

The data source, VexDB.io, is regarded in the community as a really good data source, as the canonical data is split up into a lot of csv files rather than a nice, consumable JSON api.

Unfortunately, the data is split into multiple endpoints, so I used the `spice alt_to` attribute exhibited in [xkcd](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Xkcd/Display.pm#L12-L16) to add these in. I used jQuery's `$.when().then()` to fire the needed requests off simultaneously rather than sequentially. I'm unsure about its compatibility with IE, but I haven't found anything saying that it's incompatible (I'm assuming they're using a polyfill).

https://duck.co/ia/view/vrc_teams

### DIFF
--- a/lib/DDG/Spice/VRC/Teams.pm
+++ b/lib/DDG/Spice/VRC/Teams.pm
@@ -1,0 +1,32 @@
+package DDG::Spice::VRC::Teams;
+
+# ABSTRACT: Get information for registered teams of the current season of the VEX Robotics Competition
+
+use DDG::Spice;
+
+spice is_cached => 1;
+spice proxy_cache_valid => "200 1d";
+spice to => 'http://api.vexdb.io/v1/get_teams?team=$1';
+spice wrap_jsonp_callback => 1;
+
+spice alt_to => {
+    skills => {
+        is_cached => 1,
+        proxy_cache_valid => "200 6h",
+        to => 'http://api.vexdb.io/v1/get_skills?team=$1&season=current'
+    },
+    rankings => {
+        is_cached => 1,
+        proxy_cache_valid => "200 1h",
+        to => 'http://api.vexdb.io/v1/get_rankings?team=$1&season=current'
+    }
+};
+
+triggers any => 'vex team', 'vex', 'vrc team', 'vrc';
+
+handle remainder => sub {
+    return unless $_ =~ /^(\d{1,5}[A-Z]?|[A-Z]{1,5}\d?)$/i;
+    return $1;
+};
+
+1;
--- a/share/spice/vrc/teams/vrc_teams.js
+++ b/share/spice/vrc/teams/vrc_teams.js
@@ -53,7 +53,6 @@
                 l.push({label: 'Programming', value: best_programming});
             }
 
-            console.log(l);
             return l.length ? l : undefined;
         }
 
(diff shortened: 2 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
> I used jQuery's $.when().then() to fire the needed requests off simultaneously rather than sequentially.

@mattr555 nicely done -- does this info need to be present before rendering though, or could we make those calls as the tiles render?

If so we should be able to put the calls into and `onShow()` so that the IA doesn't need to wait for all the calls to return to start rendering

moollaza:
These are all IAs that load and re-render content after display: https://github.com/duckduckgo/zeroclickinfo-spice/search?utf8=%E2%9C%93&q=rerender

moollaza:
@mattr555 this is now on Beta :+1:

mattr555:
@moollaza I tried using `onItemShown` and `meta.rerender` but it doesn't seem to be working for the infoboxes. According to the dev console, the requests themselves fire and the infobox list is set up right, but setting `item.infoboxData` doesn't take effect.

cc @bsstoner

moollaza:
@mattr555 thanks for explaining the issue. @bsstoner do you have any ideas about why it's not working?

I'm not sure if the infoBox is meant to be supported by rerender?

bsstoner:
Sorry for the delayed response, I had to go back to the code to see how re-rendering works. Unfortunately it looks like we don't have support for re-rendering with the InfoBox.

mattr555:
@bsstoner Thanks for the update.

The IA is really slow as it is, so slow that it doesn't render until too late sometimes. Can the infobox be dynamically updated or appended to some other way?

bsstoner:
Unfortunately, I don't think we really have a good way to do that right now.

mattr555:
Ok. I've reverted back to the original for now. 

@moollaza Could we fast-track the review for a v1 release? I'd love to have this out for the World Competition on 4/16? Thanks! :grin:

mattr555:
It looks like there have been changes to the data upstream as the new season kicks off next week: all of the teams have been marked unregistered, so the IA will not trigger. It's strange, because there's still the championship to play! I can [hardcode the name of the season](http://vexdb.io/the_data/rankings) but that's not super maintainable. I could also add a message that tells the user that the team isn't registered for this season, but this means we can't really get the team's stats.

moollaza:
> It's strange, because there's still the championship to play!

@mattr555 is this a bug then on their side? Are they aware they've done this?

Can we show them what you were trying to build to see if they can make changes so things will work better for the IA?

Hardcoding the season could work until it's over and then we can find a better solution.

mattr555:
@moollaza I've looked into it more and it seems like hardcoding the season name actually won't be necessary: this year's season is still considered "current," it's just that the `is_registered` flags have been set to false. I think this is the case because teams can register for next year's season while attending the championship this year. Should be triggering now :+1:

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=vex team 765A)
                
                [Mobile screenshot](https://images.duckduckgo.com/iu/?u=https://ia-screenshots.s3.amazonaws.com/vrc_teams_mobile.png)

moollaza:
@mattr555 This LGTM :+1:

Thanks again for taking the time to contribute! We really appreciate it :smiley:

daxtheduck:
@mattr555 this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

PR #2610, opened 2016-03-21
ExpandURL: Update to use new API

### LINKED ISSUE #2605: ExpandURL: API Gone offline - Update to new API: ExpandURL.com
It looks like http://longurl.org is offline -- For now I've turned off this IA. We need to determine if the API is offline permanently and if so, find an alternative or remove the Instant Answer until we have an API.   

------
IA Page:  http://duck.co/ia/view/expand_url 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @adman

MrChrisW:
- Use new API untiny.me

  - Update tests to include undef triggers

  - Fix test file formatting

  - Use api_result not api_response

  - Include `untiny` trigger

Fixes #2605
______________________________

IA Page: http://duck.co/ia/view/expand_url

### DIFF
--- a/lib/DDG/Spice/ExpandURL.pm
+++ b/lib/DDG/Spice/ExpandURL.pm
@@ -4,9 +4,21 @@ package DDG::Spice::ExpandURL;
 use strict;
 use DDG::Spice;
 
-triggers start => "http", "https", "expand", "expandurl", "unshorten";
+primary_example_queries "http://bit.ly/a", "untiny bit.ly/lolcatz";
+description "Expand shortened URLs using Untiny.me";
+name "ExpandURL";
+icon_url "/i/untiny.me.ico";
+source "Untiny";
+code_url "https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/ExpandURL.pm";
+topics "social";
+category "computing_tools";
+attribution github => ['https://github.com/mrchrisw','Chris'],
+            github => ['https://github.com/adman','Adman'];
 
-spice to => 'http://api.longurl.org/v2/expand?url=$1&format=json&callback={{callback}}';
+triggers start => "http", "https", "expand", "expandurl", "unshorten", "untiny";
+
+spice to => 'http://untiny.me/api/1.0/extract?url=$1&format=json';
+spice wrap_jsonp_callback => 1;
 
 handle query => sub {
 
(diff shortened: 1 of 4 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

MrChrisW:
@moollaza Any reason you suggested updating the existing IA over the creation of a new one? *Keeping commit history in tact?* Just interested.. :smile:

moollaza:
> Any reason you suggested updating the existing IA over the creation of a new one? Keeping commit history intact? Just interested..

Yes, keeping the commits and triggering history. Also, because it means we have to deprecate 1 IA Page, and setup a new one to go live. This way once the code is merged and the repo released the IA will be back up and running.

lib/DDG/Spice/ExpandURL.pm
```diff
@@ -4,9 +4,21 @@ package DDG::Spice::ExpandURL;
 use strict;
 use DDG::Spice;
 
-triggers start => "http", "https", "expand", "expandurl", "unshorten";
-
-spice to => 'http://api.longurl.org/v2/expand?url=$1&format=json&callback={{callback}}';
+primary_example_queries "http://bit.ly/a", "untiny bit.ly/lolcatz";
```
moollaza: @MrChrisW we can remove all this metadata -- all this lives in the IA Page now. We'll be deprecating the Perl metadata entirely soon at which point these will no longer be recognized.

MrChrisW: :+1:

moollaza:
@MrChrisW I've deployed this to beta, but for some reason the API response isn't getting through our NGINX. Seeing as it's not an HTTPS API it may be on their end -- can you try switching to api.longurl.org to see if it works better?

MrChrisW:
@moollaza You mean http://expandurl.com/? I tested that API locally and had issues with the API returning `406 not Acceptable`. Since switching to the untiny.me API I've had no issues locally, the resource is loaded correctly. - See http://untiny.me/api/1.0/extract?url=bit.ly%2Fa&format=json

moollaza:
> You mean http://expandurl.com/?

Whoops, yes!

> I tested that API locally and had issues with the API returning 406 not Acceptable

Confirmed. I think that API isn't happy with our request headers and is also sending the wrong content-type back for the "JSON" results :frowning: 

> See http://untiny.me/api/1.0/extract?url=bit.ly%2Fa&format=json

I just tested again via curl and see now that when we send an empty accept header (we do this to coax  APIs into _**not**_ Gzipping the response, they're actually sending us XML back.

We should be able to use the DDG XML endpoint to handle this, but @MrChrisW before we go down that road, can you try getting Untiny.me to handle our call correctly?

`curl -H 'Accept-Encoding:' -L http://untiny.me/api/1.0/extract?url=bit.ly%2Fa&format=json` if you try that you should get XML back instead of JSON which is totally a bug on their end

MrChrisW:
@moollaza Definitely an API issue, strange they return XML even with the `format` param! I'll contact them to try and get this resolved.

On a side note, I'm interested in why I was unable to replicate the issues experienced in the production environment in development (DuckPAN) both seem to set `Accept-Headers` to empty. It would be handy if both environment worked the same (in terms of accept-headers) so we can catch these things before they hit BETA.

https://github.com/duckduckgo/duckduckgo/blob/master/lib/DDG/Rewrite.pm#L142

https://github.com/duckduckgo/p5-app-duckpan/blob/master/lib/App/DuckPAN/Web.pm#L192

moollaza:
> Actually on testing it looks like you've missed out quotes surrounding the URL in the curl request and this is causing the format param to be dropped (due to the & symbol) - See below!

Ah! Good catch. I can confirm now that their API is returning JSON as expected. This leaves me clueless as to why this isn't working on Beta.

I'll have to get the ops team to investigate...

> On a side note, I'm interested in why I was unable to replicate the issues experienced in the production environment in development (DuckPAN) both seem to set Accept-Headers to empty.

I know how you feel. Replicating our NGINX in Curl feels impossible sometimes. We still have some APIs that refuse to work which even our ops team hasn't figured out yet...

jdorweiler:
@moollaza any update?  I installed on beta and it still seem to be a problem.  Didn't we have a similar problem with [arXiv](https://duck.co/ia/view/arxiv)? 
https://github.com/duckduckgo/zeroclickinfo-spice/pull/2431/files#diff-d2855c788f783249d81b57f6d1666ba2R12

moollaza:
@MrChrisW we found the problem! The api is redirecting `/extract?` to `/extract/?` So silly...

moollaza:
Closing and continuing in #2816

PR #2614, opened 2016-03-22
Refine couprex trigger words

b2ddg:
Refining the triggering conditions for couprex due to interference from unrelated queries.

/cc @AdamSC1-ddg 

---
IA Page: https://duck.co/ia/view/couprex

### DIFF
--- a/lib/DDG/Spice/Couprex.pm
+++ b/lib/DDG/Spice/Couprex.pm
@@ -6,7 +6,7 @@ use DDG::Spice;
 
 spice is_cached => 1;
 
-my @triggers = ("coupon", "coupon code", "promo code", "discount coupon", "discount code", "promotional code", "deal", "discount");
+my @triggers = ("coupon", "coupon code", "promo code", "discount coupon", "discount code", "promotional code");
 
 push (@triggers, map {$_."s"} @triggers);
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

AdamSC1-ddg:
Hey @b2ddg - good catches on these, @moollaza had been talking about how to refine the trigger space. Since most of the listings on the site right now are discounts and deals (only about 30-40% are coupons). 

Instead of removing the words is there common trends we could de-limit with? Such as "how to" "with" etc so that we would trigger on "shoe deals" but not "how to deal with people" etc?

b2ddg:
Hey @AdamSC1-ddg! 
It's not totally clear how to separate the query space here. For the words 'deal' and 'discount', we can probably trigger on those if they're at the beginning or end of the query only, but that doesn't seem appropriate for the other words, like coupon which should clearly trigger this IA when it's present anywhere in the query.

b2ddg:
@moollaza Let's move forward with this and handle some of overtriggering internally.

moollaza:
@b2ddg Thanks! This LGTM :+1:

Issue #2609, opened 2016-03-21
Dictionary: "define fungi" shows two play buttons


jagtalon:
And they're in the subtitle--it should be right beside the title.

https://duckduckgo.com/?q=define+fungi&ia=definition
![screen shot 2016-03-21 at 9 58 58 am](https://cloud.githubusercontent.com/assets/81969/13920893/9bd1080e-ef4b-11e5-8b73-34fb90a685aa.png)

---
https://duck.co/ia/view/dictionary_definition
Maintainer: @jagtalon

moollaza:
Hey @duckduckgo/duckduckhack-active-contributors! Anyone interested in fixing this issue?

It'll require a bit of JS knowledge. We had made some changes the the DDG Playbutton for the [Forvo Spice](https://duck.co/ia/view/forvo) and I think this IA will need some adjustments as a result.

A good place to start looking is here: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/dictionary/definition/dictionary_definition.js#L187

marianosimone:
@moollaza: I'd love to take a look. However, it seems like in my dev environment only 1 play button is shown (working from codio, with latest duckpan installed)

moollaza:
@marianosimone it looks like we're making a secondary API call in production for `fungus`. We should probably add a check that if a player exists on the page, we shouldn't add another?

tagawa:
Fixed internally.

PR #2657, opened 2016-04-07
Show only one player for audio in dictionary definition

marianosimone:
If a player for the same audio file is already present, do not show a new one

This was discussed with @moollaza in #2609, but I feel this is a hacky fix more than an actual one. Shouldn't we find out why the audio is being tried to be added twice instead?

IA Page: https://duck.co/ia/view/dictionary_definition

### DIFF
--- a/share/spice/dictionary/definition/dictionary_definition.js
+++ b/share/spice/dictionary/definition/dictionary_definition.js
@@ -184,10 +184,14 @@ var ddg_spice_dictionary = {
         // Audio URL should go to DDG's proxy server for privacy reasons.
         url = '/audio/?u=' + url;
 
-        this.playBtn = new DDG.Views.PlayButton({
-            url: url,
-            after: this.$el.find('.zci__def__pronunciation')
-        });
+        var pronunciationEl = this.$el.find('.zci__def__pronunciation');
+
+        if (pronunciationEl.siblings('[data-url="' + url + '"]').length == 0) {
+            this.playBtn = new DDG.Views.PlayButton({
+                url: url,
+                after: pronunciationEl
+            });
+        }
     }
 }
 


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
> Shouldn't we find out why the audio is being tried to be added twice instead?

@marianosimone sorry I thought I had explained. This IA checks if we're dealign with a singular or plural word. In the case of a plural, it gets the information for the singular and calls `this.render()` which grabs the audio for the current word.

https://github.com/marianosimone/zeroclickinfo-spice/blob/fix/2609_double_audio_in_dictionary/share/spice/dictionary/definition/dictionary_definition.js#L124

I believe this bug has always existed, but the player implementation only allowed 1 instance to exist so the 2nd player was likely being suppressed silently. We recently improved the player, allow multiple buttons on the page for the Forvo Spice and now this issue has cropped up.

In the case of plurals we should not get the audio for the singular term when we make the 2nd set of API calls.

moollaza:
I would argue we should abandon the current implementation using `bind` and re-write the code so it's more linear and uses `getJSON` with callback functions. That approach would make this code easier to follow and understand because as it's written it's much more difficult to follow along and understand the code path to the audio player being displayed.

marianosimone:
@moollaza yeah, I see how the singular/plural detection works and how we choose to either show the current definition or go for the singular one.

What I don't follow is how that causes `/audio` to be called twice (and why that only happens in prod)

moollaza:
> What I don't follow is how that causes /audio to be called twice (and why that only happens in prod)

@marianosimone it looks like it's a result of our organic link ("deep") trigger for this Instant Answer. It sees the Merriam Webster result and creates an API call for the singular word in those cases. Meanwhile because of the query, the Spice is triggered the normal way which also fires off a request for the audio and so we add another player to the page.

We should add some logic to prevent another player being added (which I think you've done already).

@tommytommytommy Is there a way we can prevent the deep trigger from duplicating efforts for queries that already trigger this IA? I.e. the ones that contain "definition", "define", etc

tommytommytommy:
Based on the keyword triggers for this IA, we should be able to block the duplicate internal triggers. I'll follow up internally with @jdorweiler on that component of this PR.

jdorweiler:
Assigning to @tommytommytommy since he's working on an internal fix for this.

jdorweiler:
We fixed this internally to block the second trigger to the spice.  Thanks @marianosimone

marianosimone:
@jdorweiler awesome! I'd rather have a proper solution than this hack :+1:

Issue #2637, opened 2016-03-31
Beta Instant Answers (Internal): Modified time incorrect


GuiltyDolphin:
I consistently get incorrect times for the 'Last Updated....' field, it was 5 hours behind during EST and switched to 4 hours behind after the switch to EDT, my TZ is currently BST (though was GMT until a few days ago).

![screenshot from 2016-03-31 11 48 49](https://cloud.githubusercontent.com/assets/8598426/14173659/34d300fa-f737-11e5-98f7-7940b1b619ff.png)

------
IA Page:  http://duck.co/ia/view/experimentalia
Maintainer: @vikashvverma

vikashvverma:
@GuiltyDolphin @moollaza I'm not able to run the trigger for this IA locally, Can you help me out?

jdorweiler:
@vikashvverma with duckpan?  Try this `SHOW_IA_TESTING=1 duckpan server`.  It triggers on all queries so I have it disabled unless you set an environment variable.
https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Experimentalia.pm#L15

vikashvverma:
@jdorweiler I did that still it says, `Sorry, no hit for your instant answer`

jdorweiler:
@vikashvverma can you post your output from running that?

vikashvverma:
@jdorweiler I'm trying to access it at this url http://localhost:3000/?q=testing&t=h&ia=iatesting, duckpan server is running  on port 3000

jdorweiler:
@vikashvverma is your repo up to date?  Try setting the environment variable with  `export SHOW_IA_TESTING=1` and then run duckpan.

jdorweiler:
Or if that fails you could just manually set this value to `1`.
https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Experimentalia.pm#L15

vikashvverma:
@jdorweiler It works now, one of the dependency was not installed. Thanks

vikashvverma:
@GuiltyDolphin the problem is because `moment.js` is giving relative time based on timezone,  I will work on this

PR #2660, opened 2016-04-07
Fix IA Date

vikashvverma:
Fixes https://github.com/duckduckgo/zeroclickinfo-spice/issues/2637

IA Page: http://duck.co/ia/view/experimentalia

### DIFF
--- a/share/spice/experimentalia/experimentalia.js
+++ b/share/spice/experimentalia/experimentalia.js
@@ -52,8 +52,7 @@
                     }
                 },
                 normalize: function(item) {
-                    var new_date = new Date(item.date+" EDT"),
-                        m_date = moment(new_date),
+                    var m_date = moment(item.utc_date),
                         rel_date = m_date.fromNow(),
                         creation_date = m_date.format("MMM D, YYYY");
                     
--- a/share/spice/experimentalia/experimentalia.js
+++ b/share/spice/experimentalia/experimentalia.js
@@ -52,7 +52,7 @@
                     }
                 },
                 normalize: function(item) {
-                    var new_date = new Date(item.date),
+                    var new_date = new Date(item.date+" EDT"),
                         m_date = moment(new_date),
                         rel_date = m_date.fromNow(),
                         creation_date = m_date.format("MMM D, YYYY");


vikashvverma:
@GuiltyDolphin This is just a hack but I believe we should update https://beta.duckduckgo.com/installed.json to include time zone as well.

GuiltyDolphin:
@vikashvverma This doesn't really fix the issue for two reasons:

1. This wouldn't cope with switches from EDT to EST.
2. It doesn't tell me *when* it was deployed! Which was the whole point of the issue!

You should be able to use `$loc` to determine what offset you should be using to display the modified time - heck, you might even be able to change the TZ with moment.js and have it display the correct time.

vikashvverma:
@GuiltyDolphin `TZ` requires `moment.tz.js` which I believe is not an available dependency. Also `installed.json` only contains date and doesn't mention anything else which I believe is date when the cheat sheet is last modified(not deployed).

Also updating `installed.json` to include the timezone as well might be easier way to fix this

GuiltyDolphin:
@vikashvverma You can get the time zone directly from the Perl - via `$loc` (there might be a JS equivalent?) - so you can just ping that in the request to the JS and use that time zone. 

Do you mean you can't set the time zone in `moment.js`? It seems to be able to take *some* forms of TZs....

---

Anyway, what do you think of the second suggestion? To display the absolute time in which the IA was deployed (in standard DDG+time format `%d %b %Y %T %Z` (maybe without time zone if we do the correction)).

vikashvverma:
@GuiltyDolphin How would we know the absolute time in which it was deployed because it comes from `installed.json` which does not contain any time zone information?

GuiltyDolphin:
@vikashvverma Doesn't it? Do a quick search for `date` in the file:

screenshot from 2016-04-08 11 13 51

@jdorweiler Is that the deploy date?

vikashvverma:
@GuiltyDolphin That's what I've done in `.js` file. The date seems to be in `EDT` TZ

GuiltyDolphin:
@vikashvverma It is at the moment - but it was EST a while back, I believe it is following the server time (I may be wrong).

vikashvverma:
@GuiltyDolphin While creating `installed.json` if we add TZ as well then there won't be any problem

GuiltyDolphin:
@vikashvverma I agree - I think it should be part of the date though, so `%a %b %d %T %Z %Y` rather than `%a %b %d %T %Y`.

vikashvverma:
@GuiltyDolphin yeah

GuiltyDolphin:
@vikashvverma Is this something you are able to do or does it need to come from the platform side?

vikashvverma:
I think it is coming from platform side : [https://beta.duckduckgo.com/installed.json](https://beta.duckduckgo.com/installed.json)

GuiltyDolphin:
@vikashvverma Okay - @jdorweiler is this something we can sort out?

jdorweiler:
@vikashvverma @GuiltyDolphin that date is the local time in EDT.  I think the date is used in a few spots by the IA pages so we would need to find and test all those cases.  Could you just assume the date is EDT in the spice?

vikashvverma:
@jdorweiler @GuiltyDolphin that's what I have done in this PR. Is this a good idea? `moment.js` resolves the issue if we append TZ `EDT` at the end

GuiltyDolphin:
@jdorweiler My main worry is the non-static nature of the TZ - when parts of America switched from EST to EDT the offset changed as well, so having `EDT` on the end doesn't solve the issue for all times of the year.

jdorweiler:
@GuiltyDolphin right, thanks.  Let me summarize then to make sure I'm on the same page:
- We can't use moment.tz.js since it's not an installed dependency
- Appending the time zone won't work because of EDT vs EST

Two potential ways to fix this on the backend would be to either switch to UTC time, in which case the frontend will again have to figure out EST vs EDT, or I can add the time zone to the date string.  Sounds like adding the timezone to the date string is the best solution.  @vikashvverma @GuiltyDolphin what do you think?

jdorweiler:
Also the IA page code that uses the date string from beta is here:
https://github.com/duckduckgo/community-platform/blob/master/src/js/ia/Helpers.js#L28

We would have to update and test that as well.

GuiltyDolphin:
@jdorweiler Yeah, adding TZ information would probably work. Though it would make more sense to normalize everything to UTC. For the TZ-friendly format we should probably use `%Y-%m-%dT%T%Z`.

jdorweiler:
Sounds good, I'll add a second date field in the data that's UTC `%Y-%m-%dT%T%Z` that we can use for the spice.  That way we can fix this and don't need to worry about breaking anything on the IA pages.

jdorweiler:
@vikashvverma @GuiltyDolphin I added a `utc_date` to installed.json
selection_320

vikashvverma:
@jdorweiler It seems the given `utc_date` can not be parsed to date object

vikashvverma:
@jdorweiler The `utc_date` string gives error while creating Date object out of it

GuiltyDolphin:
@jdorweiler It's not showing for me - is it on the live version already?

jdorweiler:
@vikashvverma what format do you need?  I can change it to whatever format is easiest to work with. 
@GuiltyDolphin it's probably cached by your browser. Try `ctrl+shift+r` to force it to reload.

GuiltyDolphin:
@jdorweiler That did the trick!

@vikashvverma Try `moment("2010-01-01T05:06:07", moment.ISO_8601);` (with appropriate date). If that doesn't work, it might be that we need to provide the offset rather than `EDT`; but it's worth a try.

vikashvverma:
@jdorweiler @GuiltyDolphin The date format should follow this http://momentjs.com/docs/#/parsing/string/

@GuiltyDolphin I tried doing that.  It creates `m_date` but if you do `m_date.isValid()` the it returns false because the date string given is not accepted by moment.js

GuiltyDolphin:
@vikashvverma It looks like it should work if we include the TZ as an offset rather than short-name, @jdorweiler can we try that?

jdorweiler:
@vikashvverma @GuiltyDolphin I updated that field.  Seems okay now.
selection_324

vikashvverma:
@jdorweiler Is this an issue?

screen shot 2016-04-12 at 7 47 30 pm

jdorweiler:
@vikashvverma sorry, the date string is still messed up.  That's a utc time and the local offset.  I'll update it have both the localtime and local timezone offset.

jdorweiler:
How does it look now?
selection_325

edgesince84:
@vikashvverma,  thanks again for fixing the IA date times! You’ve been doing a really good job here and we'd like you to help make DuckDuckGo even better for yourself and others interested in creating Instant Answers .

When you're ready to move on, please: 

- Create a new Instant Answer for searches you do related to Daylight Savings Time 

OR

- Test live time Instant Answers (or those in development) for bugs and suggest improvements

Detailed instructions explaining how you can help are available at the top of this [page](https://github.com/duckduckgo/duckduckgo/wiki/Creating-or-Improving-Instant-Answers).

Thanks again!

jdorweiler:
@vikashvverma @GuiltyDolphin I installed it on beta.  Looks good to me (it always was) but can you confirm that it's showing the right time for you?

GuiltyDolphin:
@jdorweiler :+1:

screenshot from 2016-04-13 15 09 01

vikashvverma:
@jdorweiler It seems fine.

See the snaps for comparison

on beta:

screen shot 2016-04-13 at 7 38 12 pm

on Github:

screen shot 2016-04-13 at 7 38 26 pm

jdorweiler:
Great thanks for all the help @vikashvverma and @GuiltyDolphin!

jdorweiler:
@GuiltyDolphin actually you're assigned so I'll let you click the green button :)

GuiltyDolphin:
@jdorweiler Hehe, click click.

@vikashvverma Thanks! :dolphin:

PR #2671, opened 2016-04-09
New Spice: Cricket Scores & Schedule

vikashvverma:
This is an attempt to build an IA for live cricket scores & schedule.

Needs a lot of improvement

Suggestions & feedbacks are welcome.

Screenshots:

screen shot 2016-04-10 at 1 03 33 am

cricket

screen shot 2016-04-10 at 10 15 24 pm

IA Page: https://duck.co/ia/view/cricket_scores_schedule


daxtheduck:
@vikashvverma Hey!
        
Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

vikashvverma:
@moollaza Can you review this

vikashvverma:
@moollaza @GuiltyDolphin Can you give your feedback on this spice

moollaza:
@vikashvverma this should emulate the design of our existing sports IAs (tiles expand horizontally, not vertically). It looks like you've copied Googles' design?

It looks like the expansion doesn't really add much value beside showing bigger logos and each teams' record?

We should be able to easily fit all this info into a single, fixed tile. For example the small MLB tiles show everything aside from the record:

blue_jays_at_duckduckgo

You could probably mimic this and put the record to the right of the team names. Our design team is going to need to draft something if we need to go the fully custom route, but I'd like to see if we can achieve something that will work with standard text templates or as little customization as possible.

@xaviervalarino @Jedidiah @altern8tif -- I know you all have design experience :) -- What are your thoughts here?

moollaza:
@vikashvverma also we're going to need 1x and 2x versions of the logos and they should all be the same size, or ideally SVG logos so we can support high-res displays

lib/DDG/Spice/Cricket.pm
```diff
@@ -0,0 +1,27 @@
...
+spice is_cached => 1;
+spice proxy_cache_valid => "200 1d"; # defaults to this automatically
+
+spice wrap_jsonp_callback => 0; # only enable for non-JSONP APIs (i.e. no &callback= parameter)
+
+# API endpoint - http://docs.duckduckhack.com/walkthroughs/forum-lookup.html#api-endpoint
+spice to => 'https://query.yahooapis.com/v1/public/yql?q=select%20*%20from%20cricket.series.ongoing&format=json&diagnostics=true&env=store%3A%2F%2F0TxIGQMQbObzvU4Apia0V0&callback={{callback}}';
+
+# Triggers - http://docs.duckduckhack.com/walkthroughs/forum-lookup.html#triggers
+triggers query_lc => qr/cricket|ipl scores{0,1}|schedules{0,1}/;
```
moollaza: This should be a word trigger. Use a regex in the `handle` if necessary

share/spice/cricket/content.handlebars
```diff
@@ -0,0 +1,84 @@
+<div class="card" id="{{matchid}}">
+    <div class="zci-cricket-score-head">
+        <span class="zci-cricket-score-head-date" style="float: left;">{{date}}</span>
+        {{#if stage}}
+            <span style="float: right; text-transform: capitalize;">{{matchNo}}</span>
```
moollaza: We shouldn't use inline styles, let's use CSS classes 👍

share/spice/cricket/content.handlebars
```diff
@@ -0,0 +1,84 @@
...
+        {{else}}
+            <span style="float: right;">{{startTime}}</span>
+        {{/if}}
+    </div>
+    <div class="zci-cricket-content {{matchid}}" onclick="show({{matchid}});">
+        <span>{{mtype}}</span>
+        <span class="zci-cricket-team">
+                <span>
+                    {{#if this.teams.0.logo}}
```
moollaza: You shouldn't need to reference `this` -- `{{#if teams.0.logo}}` should work

share/spice/cricket/cricket.css
```diff
@@ -0,0 +1,136 @@
+.zci--cricket .card {
```
moollaza: The tile design should stay as we have it -- no box shadows etc. That's all Google design, not DDG 

https://duckduckgo.com/?q=blue+jays&iai=8764&ia=mlb

share/spice/cricket/cricket.css
```diff
@@ -0,0 +1,136 @@
...
+
+.zci-cricket-content {
+    color: #888;
+    padding: 20px;
+    border-top: 2px solid #F7F7F7;
+    border-bottom: 2px solid #F7F7F7;
+}
+
+.zci-cricket-content > span:first-child {
+    float: left;
```
moollaza: we have a `pull--left` class, likewise `pull--right` that apply left/right floats

share/spice/cricket/cricket.css
```diff
@@ -0,0 +1,136 @@
...
+}
+
+.zci-cricket-team {
+    display: inline-block;
+    padding-left: 20px;
+    padding-right: 20px;
+    min-width: 200px;
+}
+
+.zci-cricket-team > span {
```
moollaza: just use a div ;)

share/spice/cricket/cricket.css
```diff
@@ -0,0 +1,136 @@
...
+    display: block;
+}
+
+.zci-cricket-team > span > span > span {
```
moollaza: use a class -- this is inefficient

share/spice/cricket/cricket.css
```diff
@@ -0,0 +1,136 @@
...
+}
+
+.zci-cricket-team-expanded-logo {
+    vertical-align: middle;
+    max-width: 100px;
+    max-height: 50px;
+}
+
+.zci-cricket-inactive {
+    display: none;
```
moollaza: I believe we have a `is-hidden` class that you can use instead

share/spice/cricket/cricket.css
```diff
@@ -0,0 +1,136 @@
...
+    border-width: 0.25em 0.25em 0 0;
+    content: '';
+    display: inline-block;
+    height: 0.45em;
+    left: 0.15em;
+    position: relative;
+    top: 0.15em;
+    vertical-align: top;
+    width: 0.50em;
+    color: #D18B49;
```
moollaza: All text colours should be applied using our classes, e.g. `text-primary` etc -- this ensure the colors are correct when the user switches themes

Checkout https://duckduckgo.com/styleguide

share/spice/cricket/cricket.css
```diff
@@ -0,0 +1,136 @@
...
+    /*float: left;*/
+}
+
+.zci-cricket-match {
+    color: #888;
+    padding: 20px;
+    border-bottom: 2px solid #F7F7F7;
+}
+
+.chevron::before {
```
moollaza: this needs to be pre-fixed with `.zci--cricket`

share/spice/cricket/cricket.js
```diff
@@ -0,0 +1,159 @@
+(function (env) {
+    "use strict";
+    var SCORECARD_ENDPOINT = 'https://query.yahooapis.com/v1/public/yql?q=select%20*%20from%20cricket.scorecard%20where%20match_id%3D{$1}&format=json&env=store%3A%2F%2F0TxIGQMQbObzvU4Apia0V0&callback=',
```
moollaza: This should be defined as `alt_to` endpoints in the Perl -- otherwise these are direct client request which expose the user's IP

share/spice/cricket/cricket.js
```diff
@@ -0,0 +1,159 @@
...
+                    };
+                },
+                templates: {
+                    item: "base_expanding_item",
+                    elClass: {
+                        tileExpand: "zci-cricket-footer"
+                    },
+                    options: {
+                        content: Spice.cricket.content,
+                        moreAt: false
```
moollaza: Is there nowhere we can send them for more details?

share/spice/cricket/cricket.js
```diff
@@ -0,0 +1,159 @@
...
+                    }
+                },
+                onItemShown: function (item) {
+                    item.teams.map(function (team, i) {
+                        $.ajax({
+                            url: DDG.get_asset_path('cricket', "assets/" + team.teamid + ".png"),
```
moollaza: Why isn't this just part of the Handlebars template?

share/spice/cricket/cricket.js
```diff
@@ -0,0 +1,159 @@
...
+                        });
+                        return team;
+                    });
+                    fetchScore(item.matchid)
+                }
+
+            });
+        });
+        //TODO: return latest match id
+        function getSelected(matches) {
```
moollaza: Why not define this within the `require()` above and use Moment?

You can also wrap this in another `require()` if you really want

moollaza:
@vikashvverma looking at your Handlebars -- I wanted to point out that you don't need to put all the content into a single file. Handlebars supports sub-templates and you can easily load them with our ``{{include ...}}` helper.

I'm thinking if you break out the portions that need to be modified you can use the Handlebars JS API to simply re-render those templates, with updated data to get your HTML and push that into the DOM. That way you don't need to deal with any HTML building in the JS. Can you please investigate if that approach will work?

All your templates are pre-rendered and made global before the IA runs so you can access all of them

share/spice/cricket/cricket.js
vikashvverma: @moollaza The problem is Yahoo rate limit the API calls and if we have too much requests then we might not be able to show live score.

We can consider other APIs in case we want to show live score.

vikashvverma:
@moollaza Thanks for reviewing. I am working on making the necessary changes.

Also it seems [](https://duckduckgo.com/?q=Giants+game) is not open sourced. So I don't have access to exact template being used so that I can use existing template or inherit the template to use with this IA

lib/DDG/Spice/Cricket.pm
vikashvverma: I will update this

share/spice/cricket/cricket.css
vikashvverma: Thanks for this link

share/spice/cricket/cricket.js
vikashvverma: I was just wondering what to display there

share/spice/cricket/cricket.js
moollaza: > The problem is Yahoo rate limit the API calls and if we have too much requests then we might not be able to show live score.

I don't understand -- what does rate-limiting have to do with making the API call from the client? We can't allow an client -> 3rd party calls because that's a privacy breach. Also, it means that everyone who loads the page would make a direct request.

Proxying the API call through DDG is necessary and let's us cache the response which reduces that # of calls. If the IA ends up being popular enough to hit rate limits DDG can speak to Yahoo (they're a partner) or make the SportsRadar API a higher priority.

All the more reason to use the backend Perl approach -- DDG can cache the response for as long as needed and then we'll reduce the number of

share/spice/cricket/cricket.js
moollaza: A link to Yahoo's Cricket score page. We should be sending them to the source for more info and in this case Yahoo has a place where they're using this data. https://cricket.yahoo.com/

moollaza:
@vikashvverma all the Sports IA's are open source: https://github.com/duckduckgo/zeroclickinfo-spice/tree/master/share/spice/sports/mlb/games

share/spice/cricket/cricket.js
vikashvverma: @moollaza Yahoo rate limits the YQL APIs per IP per day: http://stackoverflow.com/questions/9346582/what-is-the-query-limit-on-yahoos-finance-api

Anyway I will make it internal API call and later decide on upgrading as it scales.

Also we can't cache Live Score API call

share/spice/cricket/cricket.js
vikashvverma: @moollaza You mean logo?

share/spice/cricket/content.handlebars
```diff
@@ -0,0 +1,84 @@
+<div class="card" id="{{matchid}}">
+    <div class="zci-cricket-score-head">
+        <span class="zci-cricket-score-head-date" style="float: left;">{{date}}</span>
+        {{#if stage}}
+            <span style="float: right; text-transform: capitalize;">{{matchNo}}</span>
+        {{else}}
+            <span style="float: right;">{{startTime}}</span>
+        {{/if}}
+    </div>
+    <div class="zci-cricket-content {{matchid}}" onclick="show({{matchid}});">
```
GuiltyDolphin: We could do  this in the JS with jQuery for handlers.

share/spice/cricket/content.handlebars
```diff
@@ -0,0 +1,84 @@
...
+                    {{this.teams.1.Team}}
+                </span>
+            </div>
+        </div>
+        <div class="zci-cricket-team-expanded">
+            <div>
+                <span class="score-{{this.teams.0.teamid}}">{{this.teams.0.score}}</span>
+            </div>
+            <div>
+                <span class="score-{{this.teams.1.teamid}}">{{this.teams.1.score}}</span>
```
GuiltyDolphin: We could use an `{{#each` here.

vikashvverma:
@moollaza I can't find svg logo for all teams but some of the teams have svg logo which I will update

vikashvverma:
@moollaza Can we get latest commit to be deployed on beta?

vikashvverma:
Updated IA screenshot:

screen shot 2016-06-19 at 1 38 45 pm

@moollaza Checking if latest commit can get deployed on beta

zekiel:
@vikashvverma I was playing around with this on Beta to see if it's still active. Seems like the images are off -- is that the SVG problem you described? Also, are you still working on this or are you concentrating somewhere else? 
screen shot 2016-07-25 at 9 26 02 am

vikashvverma:
@zekiel It works fine locally. I think it's almost done. Any feedback?

vikashvverma:
Images are loading fine locally:

screen shot 2016-07-25 at 7 10 57 pm

vikashvverma:
@zekiel @moollaza It seems to be fine now on beta: 

screen shot 2016-07-30 at 9 55 58 pm

pjhampton:
Hey @vikashvverma 

[DuckDuckHack is now in Maintenance Mode](https://duckduckhack.com/) and from now on, we are only accepting issues and PRs for essential bugs and bug fixes.

Unfortunately, this is non-essential and will be closed.

We appreciate you taking the time to contribute and apologize for not being able to include your work.

Issue #2651, opened 2016-04-05
Holiday: False trigger


tagawa:
This IA shouldn't be triggered on the following query:
[when is california primary 2016](https://duckduckgo.com/?q=when+is+california+primary+2016&ia=answer)

As [reported on Twitter](https://twitter.com/xarph/status/717064457227112448).

------
IA Page:  http://duck.co/ia/view/holiday 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @sekhavati

PR #2678, opened 2016-04-12
Holiday: Reimplement relevancy checking

sekhavati:
#2651 highlighted that our relevancy checking with this IA is somewhat deficient. This is only the tip of the iceberg though, for example any query of the form `when is california <random word> 2016` triggers the same issue, so does `when is california 2016`, and there are probably many more examples. As a result the way this IA performs its relevancy checking is going to be re-written to use the following format:

`[<prefix>] <holiday> [<country>] [<year>]`

Where:
`<prefix>` = when is, what day is ...
`<holiday>` = the name of a known holiday
`<country>` = the country the holiday being searched for applies to
`<year>` = the year in which the holiday occurs

If year and/or country are not supplied they default to the current year and users location respectively.

---
IA Page: http://duck.co/ia/view/holiday


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

sekhavati:
For the record this PR fixes #2722 and #2723

sekhavati:
@moollaza I've made some good progress with this today. The server/client side code is practically done so thats most of the heavy lifting out of the way. The majority of the remaining work involves scrapping and sanitizing the holiday data for each country into its own file which I've made a start on, see `gb.txt`. The tests will also need updating and it looks like theres a conflict I'll need to resolve.

alohaas:
@sekhavati awesome progress here! Keep up the excellent work 💯

moollaza:
@sekhavati wow this is way more than I was expecting! 👍 💯 😮 

I think just having a master list of all holidays names will be enough to significantly reign in the triggering of this Instant Answer.

If we have a master list, we could then map the holiday names to common aliases/spellings, and countries if needed.

That can be an improvement to what you have here. How much is left? Are the biggest countries covered?

sekhavati:
@moollaza There's about 12 countries left to go and a bit more work coding wise.

moollaza:
> There's about 12 countries left to go and a bit more work coding wise.

Excellent! Please let me know when you're ready for feedback!

iambibhas:
@sekhavati good job here. You're doing what we thought wont be necessary back then. When I made this we were relying on the timeanddate api to return relevant result for the given keyword and year. We even [discussed making list of holidays, but decided not to at that point](https://github.com/duckduckgo/zeroclickinfo-spice/pull/1701#issuecomment-89427683). It was eventually going to grow out of that and it did.

Good to see you taking control of it. :+1: Let me know if I can help somehow.

moollaza:
@iambibhas good to see you again! @sekhavati this is looking awesome -- how much is left? Can I move this along -- I don't think we have to wait for entire coverage of the world unless there are any major countries left?

sekhavati:
@moollaza Almost done, I just need to fix the tests which I'll try find time to look at tomorrow. There are a few rough edges but I'll leave those for now as I think its better that we get something live even if its not perfect.

sekhavati:
@moollaza This is ready for feedback now. As I mentioned before there are some shortcomings I haven't yet addressed but this should do for now.

lib/DDG/Spice/Holiday.pm
```diff
@@ -2,68 +2,87 @@ package DDG::Spice::Holiday;
...
-spice to => 'http://www.timeanddate.com/scripts/ddg.php?m=whenis&c=$2&q=$3&y=$4&callback={{callback}}';
+spice to => 'http://www.timeanddate.com/scripts/ddg.php?m=whenis&c=$1&q=$2&y=$3&callback={{callback}}';
 
-handle query_lc => sub {
-    return unless ($_);
+my @triggers  = ('when is', 'what day is');
+my $countries = join('|', share("countries.txt")->slurp(chomp => 1));
 
-    my ($t, $q, $c, $y);
+triggers start => @triggers;
```
moollaza: @sekhavati now that we have a list of all supported holidays, the triggers should be the unique set of holiday names. That way "father's day" or "father's day 2017" will trigger this and show the correct information.

Based on the user's country we can check if the named holiday exists for their country and if not, fallback to the US for the API call.

Does this make sense?

Can you create a "holidays.txt" file with the unique names of all holidays across all countries? We should probably build it with a script that uses the existing `.hols` files. Also, it may be best to rename all those to a standard file extension? I think we can even put them in their own subdirectory

e.g. `/share/spice/holiday/countries/al.txt`

share/spice/holiday/ae.hols
```diff
@@ -0,0 +1,25 @@
+al hijra
+arafat
+december solstice
+eid al adha
+eid al adha holiday 1
```
moollaza: Looks like we can remove the "holiday #"  entries?

share/spice/holiday/br.hols
```diff
@@ -0,0 +1,35 @@
...
+labor day
+march equinox
+may day
+mothers day
+new years day
+new years eve
+our lady aparecida
+public service holiday
+republic proclamation day
+rio 2016 summer olympics end
```
moollaza: this is cool!

share/spice/holiday/by.hols
```diff
@@ -0,0 +1,27 @@
+catholic christmas day
+catholic easter sunday
+constitution day
+day 1
```
moollaza: seen a few instances of this "day 1", "day 2", etc -- we should remove

moollaza:
@sekhavati this one is having some trouble deploying. Can you please `git pull upstream master` or rebase it? There aren't any conflicts but it's a little dated which I think is causing problems

moollaza:
@sekhavati this has been deployed on Moollaza and is working as expected:

https://moollaza.duckduckgo.com/?q=when+is+easter+sunday+2017+in+usa&t=hb&ia=answer

/cc @jdorweiler

moollaza:
@sekhavati Thanks! This LGTM :+1:

lib/DDG/Spice/Holiday.pm
```diff
@@ -13,29 +13,23 @@ spice wrap_jsonp_callback => 0;
...
-    #   "father's day" -> "fathers day"
-    $query =~ s/[^\w\s\-]//;
-    $query =~ s/-/ /;
-
     # Current year and users location are the defaults unless otherwise specified by the query
     my $defaultYear = (localtime(time))[5] + 1900;
     my $defaultCountry = $loc->country_name;
 
     # Regexes to match components of queries relevant to this IA
     my $country  = qr/$countries/;
```
moollaza: Please move these regex out of `handle` as they're static

lib/DDG/Spice/Holiday.pm
```diff
@@ -13,29 +13,23 @@ spice wrap_jsonp_callback => 0;
...
     my $year     = qr/[1-9]{1}[0-9]{3}/;
+    my $prefix   = qr/^(when is |what day is )/;
     
     # Regexes to ignore optional words that can precede the year or country name
     my $in       = qr/(?:in )?/;
```
moollaza: same here 👍

share/spice/holiday/holidays.txt
```diff
@@ -0,0 +1,1367 @@
...
+american eagle day
+american independence day
+american indian heritage day
+amtseinführung museveni
+amun jadid
+ancestors’ day
+annexation of guanacaste
+anniversary of khomeinis death
+anniversary of president laurent kabila’s assassination
+anniversary of prime minister patrice emery lumumba’s assassination
```
moollaza: let's remove any instances of `'s` and normalize the query accordingly

share/spice/holiday/holidays.txt
```diff
@@ -0,0 +1,1367 @@
...
+anniversary of prime minister patrice emery lumumba’s assassination observed
+anniversary of the arengo
+anniversary of the decease of m r ŝtefánika
```
moollaza: if possible, we should also try to normalize UTF8 chars.

e.g. this would be `stefanika` -- I suspect a Perl module exists to do this. If not it's not very important

share/spice/holiday/holidays.txt
moollaza: or are these required to lookup the holiday on timeanddate?

moollaza:
@sekhavati @jdorweiler I just realized I made a mistake and deployed the old Holiday code so it's not actually working on Moollaza.

However, I believe I just found out the reason why it isn't working: 

https://github.com/sekhavati/zeroclickinfo-spice/blob/19b1faf7d70990bd82cd0502736258eee1f6dba4/lib/DDG/Spice/Holiday.pm#L71

That line is always going to return `false` in production. You can't access the share dir in that manner in productions @sekhavati 

We'll need to try loading the `share` file and if unsuccessful, `return`

PR #2706, opened 2016-04-23
Use new GBFS feed for CitiBike NYC

### LINKED ISSUE #2156: CitiBike: Need to normalize dates to standard format and UTC Timezone

### LINKED ISSUE #2745: Bike Sharing New York City (Citi Bike): IA Disabled, API URL Changed?
This IA has been taken offline because it started failing.

It looks like the URL for the API has changed and is now redirecting to a new subdomain. The `spice to` needs to be updated accordingly to get this IA online again

------
IA Page:  http://duck.co/ia/view/citi_bike_nyc 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @marianosimone

marianosimone:
The new GBFS feeds will be the only maintained from now on
More info: https://github.com/NABSA/gbfs/
As a side effect, #2156 is fixed, as they now use POSIX timestamps

Fixes #2745, Fixes #2156 

---
IA Page: https://duck.co/ia/view/citi_bike_nyc

### DIFF
--- a/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
+++ b/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
@@ -16,80 +16,101 @@
         }
     };
 
+    function get_query() {
+        var script = $('[src*="/js/spice/bike_sharing/citi_bike_nyc/"]')[0],
+            source = $(script).attr("src");
+        return source.match(/citi_bike_nyc\/([^\/]+)/)[1];
+    }
+
     env.ddg_spice_bike_sharing_citi_bike_nyc = function(api_result) {
-        if (!api_result || !api_result.stationBeanList) {
+        if (!api_result || !api_result.data || !api_result.data.stations) {
             return Spice.failed('citi_bike_nyc');
         }
 
-        var script = $('[src*="/js/spice/bike_sharing/citi_bike_nyc/"]')[0],
-            source = $(script).attr("src"),
-            query = source.match(/citi_bike_nyc\/([^\/]+)/)[1];
+        $.getJSON("https://gbfs.citibikenyc.com/gbfs/en/station_status.json")
+            .done(function(status_api_result) {
+                if (!status_api_result || !status_api_result.data || !status_api_result.data.stations) {
+                    return Spice.failed('citi_bike_nyc');
+                }
 
-        DDG.require(['moment.js', 'maps'], function() {
-            moment.locale('en', {
-                relativeTime : {
-                    past: "%s ago",
-                    s:    "1 sec",
-                    ss:   "%d sec",
-                    m:    "1 min",
-                    mm:   "%d min",
-                    h:    "1 hour",
-                    hh:   "%d hours",
-                    d:    "1 day",
-                    dd:   "%d days",
-                    M:    "1 month",
-                    MM:   "%d months",
-                    y:    "1 year",
-                    yy:   "%d years"
-                }
-            });
-            Spice.add({
-                id: 'citi_bike_nyc',
-                name: 'Bike Sharing',
-                meta: {
-                    sourceName: 'Citi Bike NYC',
-                    sourceUrl: 'https://www.citibikenyc.com/stations',
-                    primaryText: 'Showing ' + api_result.stationBeanList.length + ' Stations',
-                    pinIcon: 'ddgsi-circle',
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
@marianosimone apologies for the delay, deploying to Beta

To confirm, this IA now requires 2 calls per station?

Can the status calls be made on the frontend? After initial display?

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -16,80 +16,104 @@
...
-                        var pointOfReference = L.latLng(references[query]);
+
+                var stations_information = api_result.data.stations.reduce(function(stations, station) {
+                    stations[station.station_id] = station;
+                    return stations;
+                }, {});
+
+                var full_data = status_api_result.data.stations.map(function(station_status) {
+                    var station_information = stations_information[station_status.station_id];
+                    if (station_information) {
```
moollaza: `$.extend()` might be a little simpler to user here?

marianosimone: :+1: good catch

marianosimone:
@moollaza this IA now requires two calls per query (not station): one gets the stations themselves (name, location), and a second the current status (number of available bikes, docks, etc.)

If I understand how this works, the second call IS being made in the frontend (line 30). Am I missing something?

moollaza:
> @moollaza this IA now requires two calls per query (not station): one gets the stations themselves (name, location), and a second the current status (number of available bikes, docks, etc.)

Ah ok, thanks for clarifying. For some reason I thought something was looping in there, should have looked closer.

> If I understand how this works, the second call IS being made in the frontend (line 30). Am I missing something?

@marianosimone you're correct that it's making the request from the frontend, but what I meant was, can we make the second API call _after_ the `Spice.add()` call, using the data from the first call to build most of the Instant Answer display.

That way we only need to wait for 1 API call to start showing something to the user, rather than waiting for 2 round trips because calling `Spice.add()`

For example, the Amazon Instant Answer builds all the tiles and uses `onItemShown()` to make an additional call for each tile to request it's rating/review data.

If that's not possible though it's OK, XKCD works by making two calls to show the IA.

marianosimone:
I see what you mean, but in this particular case I don't think it makes too much sense: Until the second call is made, there's no info on availability, which is what's more prominent on the card (we *could* show the card with just the name and something denoting that more data is being loaded... I'll let you guys decide if you think it's a good idea)

Also, it seems like handlers like `onItemShown()` don't update the DOM, so we would need some way to force the rendered template

moollaza:
> Until the second call is made, there's no info on availability, which is what's more prominent on the card

Understood. This will be fine then 👍 

> Also, it seems like handlers like onItemShown() don't update the DOM, so we would need some way to force the rendered template

Actually they do! We pass the `item` through to the callback you define. It can call `item.set({key1: val, key2: val}});` which will update the underlying model for the item. You also need to set `meta: { rerender: ['key1', 'key2'] }` and then the DOM will automatically update when those values are changed in the model 👍 

Here's some examples: https://github.com/duckduckgo/zeroclickinfo-spice/search?utf8=%E2%9C%93&q=rerender

moollaza:
@marianosimone Thanks! This LGTM :+1:

marianosimone:
🎉  it got merged!!!

Just a quick note about `onItemShow`, @moollaza : I got confused by the last part of the documentation (http://docs.duckduckhack.com/frontend-reference/display-reference.html - "The item Argument" section):

> Events relevant to specific items pass an item to their handler functions. This is a reference to the item's data object, rendered to the template. Modifying the properties will not update the DOM.

PR #2725, opened 2016-05-03
New Syntax Spice

anthonyn60:

- [x] New Instant Answer: Syntax 

Lets you look up syntax for your programming language

###### Description of changes

- Adds syntax lookup to DuckDuckGo IA

###### People to notify
@moollaza 

---

Instant Answer Page: https://duck.co/ia/view/syntax

[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @anthonyn60


daxtheduck:
@anthonyn60 Hey!
        
Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

anthonyn60:
Also, can the tab name be changed to programming? It's not quite a cheat sheet (classified it wrong at first I guess)

anthonyn60:
Is there any update my pull request?

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,30 @@
+package DDG::Spice::Syntax;
+
+# ABSTRACT: Look up syntax for a programming language
+
+# Start at http://docs.duckduckhack.com/walkthroughs/forum-lookup.html if you are new
```
mintsoft: Can you bomb out these comments? They're only really for instruction (with the exclusion of the # ABSTRACT)

share/spice/syntax/syntax.css
```diff
@@ -0,0 +1,4 @@
+.zci--syntax {
```
mintsoft: If there's no styling, let's delete this file :)

share/spice/syntax/syntax.handlebars
```diff
@@ -0,0 +1,3 @@
+
+{{description}}
```
mintsoft: If you're using the default template then we should be able to get away without a `.handlebars` file

anthonyn60: I need the handlebar to put a <pre> tag on the code section. Also for whatever reason, if I don't include the description handlebar it doesn't show.

anthonyn60:
Changes made and tested.

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,27 @@
...
+spice is_cached => 1;
+spice proxy_cache_valid => "200 1d"; # defaults to this automatically
+
+spice wrap_jsonp_callback => 1; # only enable for non-JSONP APIs (i.e. no &callback= parameter)
+
+# API endpoint - http://docs.duckduckhack.com/walkthroughs/forum-lookup.html#api-endpoint
+spice to => 'https://syntaxdb.com/api/v1/concepts/search?q=$1';
+
+# Triggers - https://duck.co/duckduckhack/spice_triggers
+triggers any => 'syntax', 'syntaxdb', 'syntaxcenter';
```
moollaza: this should be `startend` -- `any` is too broad, especially given the `remainder` below.

For example if the query is just "syntax" this will trigger and attempt to show a result

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,27 @@
...
+
+# Handle statement
+handle remainder => sub {
+
+    # Query is in $_ or @_, depending on the handle you chose...if you
+    # need to do something with it before returning
+    return $_;
```
moollaza: We need to do more query checking in here -- either with a regex to ensure the query follows an expected format, or to ensure that it contains terms like "for loop", "while loop", etc. and the name of a programming language.

As it stands, any query that contains the above trigger words, **any**where in the query is going to trigger this an issue an API call which we don't want.

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,47 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_syntax = function(api_result) {
+        
+        function getParameterByName(name) {
```
moollaza: Let's move this function definition outside of the callback

moollaza: Actually, we can remove this funciton altogether, we have a `DDG.get_query()` function you should be using

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,47 @@
...
+        var script = $('[src*="/js/spice/syntax/"]')[0],
+            source = $(script).attr("src"),
+            query = source.match(/syntax\/([^\/]+)/)[1],
+            decodedQuery = decodeURIComponent(query);
+
+        // Validate the response (customize for your Spice)
+        if (!api_result || api_result.error || api_result.length == 0) {
+            return Spice.failed('syntax');
+        }
+        var query = getParameterByName('q');
```
moollaza: Please use `DDG.get_query()`

lib/DDG/Spice/Syntax.pm
moollaza: For example all these would currently trigger this:

syntax_for_dummies_at_duckduckgo

syntax_for_dummies_at_duckduckgo

anthonyn60: Definitely didn't think of non-programming related syntax queries, I'll work on making a list of keywords and languages and writing a regex.

For `any`, the reason I did that was because if someone searched something like `Java syntax for loop` then I'd want the IA to trigger. I originally did have it as `startend`.

anthonyn60: Latest commit addresses this issue.

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,31 @@
...
+spice to => 'https://syntaxdb.com/api/v1/concepts/search?q=$1';
+
+# Triggers - https://duck.co/duckduckhack/spice_triggers
+triggers any => 'syntax', 'syntaxdb', 'syntaxcenter';
+my $languages = join "|", share('languages.txt')->slurp;
+my $concepts = join "|", share('concept.txt')->slurp;
+
+# Handle statement
+handle remainder => sub {
+    if ($_ =~ m/\b($languages)\b/x) {
```
GuiltyDolphin: Why the `x`? That's for ignoring white space *in* the regex so you can document it.

Also, we can use `&&` here rather than two `if` statements; I'd recommend `return unless $_ =~ /.../ && $_ =~ /.../`

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,31 @@
...
+# Caching - http://docs.duckduckhack.com/backend-reference/api-reference.html#caching
+spice is_cached => 1;
+spice proxy_cache_valid => "200 1d"; # defaults to this automatically
+
+spice wrap_jsonp_callback => 1; # only enable for non-JSONP APIs (i.e. no &callback= parameter)
+
+# API endpoint - http://docs.duckduckhack.com/walkthroughs/forum-lookup.html#api-endpoint
+spice to => 'https://syntaxdb.com/api/v1/concepts/search?q=$1';
+
+# Triggers - https://duck.co/duckduckhack/spice_triggers
```
GuiltyDolphin: Lets get rid of the auto-generated comments.

t/Syntax.t
```diff
@@ -0,0 +1,36 @@
...
+        call_type => 'include',
+        caller => 'DDG::Spice::Syntax'
+    ),
+    # Concept without language
+    'syntax for loop' => undef,
+    # Language without concept should trigger nothing
+    'syntax java' => undef,
+    # Syntax by itself should trigger nothing
+    'syntax' => undef,
+    # Syntax with garbage should trigger nothing
```
GuiltyDolphin: I think the `should trigger nothing` is implicit here.

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,36 @@
...
+        // Validate the response (customize for your Spice)
+        if (!api_result || api_result.error || api_result.length == 0) {
+            return Spice.failed('syntax');
+        }
+        var query = DDG.get_query().replace("syntax", '');
+        // Render the response
+        Spice.add({
+            id: "syntax",
+            // Customize these properties
+            name: "Programming",
```
GuiltyDolphin: We should probably use `Answer` here.

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,36 @@
...
+            data: api_result[0],
+            meta: {
+                sourceName: "SyntaxDB.com",
```
GuiltyDolphin: Just use `SyntaxDB`

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,36 @@
...
+                sourceUrl: 'https://syntaxdb.com/reference/search?search=' +  query,
```
GuiltyDolphin: This should just be `https://syntaxdb.com`.

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,31 @@
...
+triggers any => 'syntax', 'syntaxdb', 'syntaxcenter';
+my $languages = join "|", share('languages.txt')->slurp;
```
GuiltyDolphin: You'll want to do `join '|', map { quotemeta $_ } ...` so we don't mess up the regexes.

share/spice/syntax/syntax.js
anthonyn60: The reason I use the dynamically generated `sourceUrl` is because I want them to get the respective results page for their query when they click `More at SyntaxDB`.

GuiltyDolphin: @anthonyn60 Fair dos.

GuiltyDolphin:
@moollaza I think this is ready for Beta.

GuiltyDolphin:
@anthonyn60 Looks like it's over triggering: https://beta.duckduckgo.com/?q=syntax+java&ia=cheatsheet

zekiel:
cool stuff @anthonyn60 ! This is back on beta: https://beta.duckduckgo.com/?q=syntax+for+each+in+java&ia=cheatsheet 

@moollaza  is out until Thursday or so but I don't see much out place beyond the tab name being "Cheat Sheet" vs. "Answer" Shouldn't prevent a merge, though, and we can make an issue for it afterward. Someone will get back to this shortly! In the meantime, care to update the [programming coverage table](https://github.com/duckduckgo/duckduckgo/wiki/Programming-IA-Coverage) for this and/or take on something else?

anthonyn60:
Appreciate it @zekiel! I'm not entirely sure why the tab is like that, I've changed it everywhere. Also, at the time of initial deployment there was an issue with over-triggering, which since has been fixed (a redeploy should solve it!).

zekiel:
Could be that ID is set to `syntax` so I've updated your IA page to something more unique: "programming_syntax"  Can you update ID in spice.add to see if that fixes? If not, we'll take a look before merge. 
Really excited to see how this gets better with use from the community!

anthonyn60:
@zekiel Where exactly in the code would this be?

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,32 @@
+(function (env) {
+    "use strict";
+    env.ddg_spice_syntax = function(api_result) {
+        if (!api_result || api_result.error || api_result.length == 0) {
+            return Spice.failed('syntax');
+        }
+        var query = DDG.get_query().replace("syntax", '');
+        Spice.add({
+            id: "syntax",
```
GuiltyDolphin: @anthonyn60 Heere!

@zekiel Might we not also have to update the share directory?

zekiel: Yes, it will need to be updated in several places where it's referenced. The module name in particular. For fear of breaking the process, I'm going to wait for @moollaza  to help with this.

moollaza:
@anthonyn60 I've changed the ID of the IA to `syntax` which matches the code. I think I changed it to `programming_syntax` before leaving because that is a little more description, but that doesn't match the code which complicates things.

I think ideally this would be called `syntax_db` with the Perl package being `DDG::Spice::SyntaxDB`

For now this is fine and works. The tab name is now correct on Beta

daxtheduck:
Automated screenshot from [beta.duckduckgo.com](https://beta.duckduckgo.com/?q=syntax for loop in java)
                
                [Mobile screenshot](https://images.duckduckgo.com/iu/?u=https://ia-screenshots.s3.amazonaws.com/syntax_mobile.png?JXbYfBN9h5)

moollaza:
@anthonyn60 this is looking good, but I think the API matching needs to be stricter, or we need to use relevancy checking on the client side.

https://beta.duckduckgo.com/?q=java+syntax+for+string&ia=programming

Take a look at the autocomplete queries for `java syntax for` and you'll see many potential queries that will trigger this and shouldn't return a result

moollaza:
Actually it looks like there are only a few that have problems the relevancy is pretty good, I'll see if it breaks in any other places

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,32 @@
...
+            data: api_result[0],
+            meta: {
+                sourceName: "SyntaxDB",
+                sourceUrl: 'https://syntaxdb.com/reference/search?search=' +  query,
+                sourceIconUrl: 'https://syntaxdb.com/images/favicon.ico'
+            },
+            normalize: function(item) {
+                return {
+                    title: item.concept_search,
+                    subtitle: "Syntax",
```
moollaza: This should be more specific. I would prefer "Java Syntax", "Python Syntax", etc to re-affirm that I'm seeing the correct language

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,25 @@
...
+
+spice wrap_jsonp_callback => 1;
+spice to => 'https://syntaxdb.com/api/v1/concepts/search?q=$1';
+
+triggers any => 'syntax', 'syntaxdb', 'syntaxcenter';
+
+my $languages = join "|", map { quotemeta $_ } split "\n" , share('languages.txt')->slurp ;
+my $concepts = join "|", map { quotemeta $_ } split "\n", share('concept.txt')->slurp;
+
+handle remainder => sub {
```
moollaza: should be `remainder_lc` to ensure case-insensitive matching

moollaza: e.g. https://beta.duckduckgo.com/?q=Python+syntax+for+if+else&ia=iatesting

lib/DDG/Spice/Syntax.pm
```diff
@@ -0,0 +1,25 @@
...
+    if ($_ =~ /\b($languages)\b/ && $_ =~ /\b($concepts)\b/) {
```
moollaza: This feels a little too broad, but seems to work. We may have to reign this in later

share/spice/syntax/syntax.js
```diff
@@ -0,0 +1,32 @@
...
+                };
+            },
+            templates: {
+                group: 'text',
```
moollaza: We may need to use the `info` template, with `meta: { chompContent: true }` to ensure the text snippet isn't too long. The snippets seems to be short enough, but right now this could potentially fill the page if the code snippet is big enough

moollaza:
A couple more examples:

- https://beta.duckduckgo.com/?q=javascript+syntax+for+not+equal&ia=programming
- https://beta.duckduckgo.com/?q=javascript+syntax+for+or&ia=programming

Overall, @anthonyn60 this LGTM :+1: 

Thanks for the awesome work :)

daxtheduck:
@anthonyn60 this is now merged! It should be live on DuckDuckGo within one week (or during the next release) and we'll be sure to notify you when it happens.

anthonyn60:
@moollaza I actually had this thought when I was developing a bot for SyntaxDB, so I addressed this with NLP through LUIS.ai (although that's client side). Definitely an interesting problem and I'll have a look into server side solutions soon enough.

PR #2785, opened 2016-06-03
proxy citibike nyc status calls

### LINKED ISSUE #2784: CitiBike: Need to proxy .getJSON requests
Running `$.getJSON("https://gbfs.citibikenyc.com/gbfs/en/station_status.json")` directly in the Spice is not allowed since it's a breach of privacy (it accesses a resource outside of DuckDuckGo). We need to proxy this.

How do we proxy?

See example proxy endpoints in the Dictionary Instant Answer here https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/lib/DDG/Spice/Dictionary/Definition.pm#L10..L25 and how it's used in the JS https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/dictionary/definition/dictionary_definition.js#L98

Don't be afraid to ask for help! 

------
IA Page:  https://duck.co/ia/view/citi_bike_nyc
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @marianosimone

marianosimone:
fixes #2784

https://duck.co/ia/view/citi_bike_nyc

### DIFF
--- a/lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm
+++ b/lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm
@@ -10,6 +10,12 @@ spice wrap_jsonp_callback => 1;
 spice is_cached => 0;
 spice proxy_cache_valid => '200 304 15m';
 
+spice alt_to => {
+	citibike_nyc_status => {
+		to => 'https://gbfs.citibikenyc.com/gbfs/en/station_status.json'
+	}
+};
+
 my @places = ("new york city", "new york", "nyc", "ny", "brooklyn", "manhattan", "queens");
 my $place_re = join "|", @places;
 
--- a/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
+++ b/share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
@@ -27,7 +27,7 @@
             return Spice.failed('citi_bike_nyc');
         }
 
-        $.getJSON("https://gbfs.citibikenyc.com/gbfs/en/station_status.json")
+        $.getJSON('/js/spice/bike_sharing/citibike_nyc_status/')
             .done(function(status_api_result) {
                 if (!status_api_result || !status_api_result.data || !status_api_result.data.stations) {
                     return Spice.failed('citi_bike_nyc');


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

marianosimone:
@jagtalon I think I'm headed in the right direction, but this still doesn't work.

For some reason, the regular handler gets called time and time again when I trigger the IA (so `handle_stations_data` gets called infinitely, then `/stations_status` gets called, returns the expected data, but the `handle_stations_status` handler is not invoked... but `handle_stations_data` is, with the corresponding API result, not the statuses)

share/spice/bike_sharing/citi_bike_nyc/bike_sharing_citi_bike_nyc.js
```diff
@@ -22,95 +22,100 @@
...
+                            lastCommunication: moment(new Date(parseInt(station.last_reported)*1000)).fromNow()
+                        };
+                    }
                 });
-        });
-    }
+            });
+        }
+    };
+    env.ddg_spice_bike_sharing_citi_bike_nyc = bike_sharing_citi_bike.handle_stations_data.bind(bike_sharing_citi_bike);
```
moollaza: Personally, I hate this `bind` approach we've used elsewhere. IMO it makes the code very hard to follow linearly.

I blame @jagtalon ;)

I would do it this way:

```
env.ddg_spice_bike_sharing_citi_bike_nyc = function(api_result) {
  // validate initial API results

  // make 2nd api call
  $.getJSON( '/js/spice/bike_sharing/nyc_status', function( data ){
    // response validation

    // cleanup data

    DDG.require(['moment.js', 'maps'], function() {
        ...
        Spice.add({ ... });
    });
}
```

Does that make sense?

moollaza: This way you don't need to any `bind` and there's no callback context switching. The execution follows as the code is written.

marianosimone: Yeah... that is the way I built it initially, but I was following the example and applied this change as well when using the proxy. Will roll back ;)

lib/DDG/Spice/BikeSharing/CitiBikeNYC.pm
```diff
@@ -10,6 +10,12 @@ spice wrap_jsonp_callback => 1;
 spice is_cached => 0;
 spice proxy_cache_valid => '200 304 15m';
 
+spice alt_to => {
+	status => {
```
marianosimone: mhhh... that doesn't work. Remember that this is initially "namespaced" by the directory structure, so the right URL should be `/js/spice/bike_sharing/citi_bike_nyc/status`.

Removing `citi_bike_nyc` obviously doesn't work, because we are one dir below that. But the inner URL rewriting system doesn't seem to understand nested files, returning the same for both `/js/spice/bike_sharing/citi_bike_nyc/status` and (the non-existent) `/js/spice/bike_sharing/citi_bike_nyc/fuuuuu`; in both cases, it returns the first level reponse

moollaza: @marianosimone the `alt_to` path prepends the name of the subdirectory, if one exists:

`/js/spice/subdir_name/alt_to/`

In this case, it's inside `BikeSharing` so the path is `/js/spice/bike_sharing/status`.

If there is no subdirectory, then the path is just `/js/spice/alt_to` -- so in those cases the `alt_to` name should be named to indicate what IA it belongs to.

The KXCD IA is a good example to follow. We have `lib/DDG/Spice/XKCD/Display.pm` which defines an `alt_to`, called "status".

Thus the endpoints are `/js/spice/xkcd/display` and `/js/spice/xkcd/latest` and the JS calls the latter: https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/share/spice/xkcd/display/xkcd_display.js#L44

----

> Removing citi_bike_nyc obviously doesn't work, because we are one dir below that. But the inner URL rewriting system doesn't seem to understand nested files, returning the same for both /js/spice/bike_sharing/citi_bike_nyc/status and (the non-existent) /js/spice/bike_sharing/citi_bike_nyc/fuuuuu; in both cases, it returns the first level reponse

What's happening here is that you're calling the endpoint for the CitiBikeNyc.pm `spice to` and sending it the param "status". However, your `spice to` doesn't use `$1` anywhere, so the `status` word is ignored. 

AFAIK, we can't nest endpoints within each other because of how the NGINX rewrites work, meaning any extension of `/js/spice/bike_sharing/citi_bike_nyc/` is going to be re-written.

If we try to have them nested, we'd run into issues with reserved keywords:

i.e. `/js/spice/sub_dir/endpoint_a/keyword1/keyword2`, `/js/spice/sub_dir/endpoint_a/endpoint_a_b/keyword1/keyword2`

How do you differentiate the 2nd url from `/js/spice/sub_dir/endpoint_a/keyword1/keyword2/keyword3`?

NGINX just applied a regular expression to the string `i.e. `keyword1/keyword2/keyword3`, attempts to capture part of it, and proxy the request to the `spice to` URL, replacing `$1, $2, $3` if they're defined and captured.

Does that make more sense?

marianosimone: Done... I was actually understanding how things worked.. but was being p0wned by the fact that you need to add a trailing `/` to the URL for the NGINX to apply the rewrite rule :(

moollaza:
Deploying to Beta...

moollaza:
@marianosimone this is now working on beta!

Excellent work 👍 💯

PR #2520, opened 2016-02-18
Github: Enhance github search results (Closes #2518)

swapagarwal:
Show repository language in search results.
Closes #2518 
IA Page: https://duck.co/ia/view/github

PR #2855, opened 2016-06-27
Update GitHub Spice

jdorweiler:

Choose the most relevant items and use the following title template to name
your Pull Request.

- [ ] New Instant Answer: **`New {IA Name} Instant Answer`**
- [x] Improvement
    - [ ] Bug fix: **`{IA Name}: Fix {Issue number or one-line description}`**
    - [x] Enhancement: **`{IA Name}: {Description of Improvements}`**
- [ ] Non–Instant Answer
    - [ ] Other (Role, Template, Test, Documentation, etc.): **`{SpiceRole/Templates/Tests/Docs}: {Short Description}`**

###### Description of changes

1. remove triggering from the `Github.pm`. We're going to do triggering internally. 
2. only show single results. 
3. added infobox
4. added request for wikipedia data for popular repos

---

Instant Answer Page: https://duck.co/ia/view/github


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

jdorweiler:
Ready for some feedback @moollaza @duckduckgo/community-leaders 
Big changes here are:
1. Show a single result
2. Make a separate request to our api for repos that we know have wiki articles. Use that for the abstract text (fatheads.json)
3. Added an info box

note: you may need to `ctrl+r` on beta to load the new JS. It should look something like this.

selection_392

Snapy:
Here is some feedback.
1. Why did you completely remove the possibility to display more results ? Sometimes it's useful (e.g. for projects where forks are better or still maintained) so if you could consider reimplementing this feature it would be nice or give an option to show forks or similar projects.
3. This is really great and all the info is useful.
I don't know if it's possible but if the same can be done for organizations it would be a good addition.

share/spice/github/fatheads.json
```diff
@@ -0,0 +1,15 @@
+{
+    "jquery":       { "url": "1/jquery&ia=about" },
+    "react":        { "url": "?q=react+javascript+library&ia=about" },
+    "opencv":       { "url": "1/opencv&ia=about" },
+    "bootstrap":    { "url": "?q=bootstrap+front-end+framework&ia=about" },
+    "angular.js":   { "url": "1/angular_js&ia=about" },
+    "angular":      { "url": "1/angular_js&ia=about" },
+    "Font-Awesome":  { "url": "1/font_awesome&ia=about" },
```
moollaza: should this be lowercased?
do we need an alias for `fontawesome`?
YAML may be better if we're going to have lots of these and aliases?

share/spice/github/github.js
```diff
@@ -6,73 +6,96 @@
...
+                                        description : fatheads[key].abstract, 
+                                        image: fatheads[key].image
+                                    });
+                                }
+                            });
+                    }
+                }
             });
+            
+            function get_days(date) {
```
moollaza: Moment can do this for you 😃  `moment(date).fromNow()`

http://momentjs.com/docs/#/displaying/fromnow/

share/spice/github/github.js
```diff
@@ -6,73 +6,96 @@
...
+                var last_update = new Date(date),
+                now = new Date(),
+                elapsed_days = days(now - last_update);
+                
+                if (elapsed_days <= 1){
```
moollaza: Can you moment here as well: http://momentjs.com/docs/#/displaying/difference/

`if ( moment().diff(date, 'days') == 1 ) { ... }`

share/spice/github/github.js
```diff
@@ -6,73 +6,96 @@
...
+            results = results[0];
 
-        DDG.require("moment.js", function() {
+        var key = results.full_name.replace(/^.*\//, '');
+        results.title = key;
+        results.url = results.html_url;
+        results.imageTile = true;
+        
+        var fatheads;
+        $.getJSON(DDG.get_asset_path('github', 'fatheads.json'), function (data) { fatheads = data; });
```
moollaza: The callback function here needs to encompass all of the code below. `fatheads` will only be defined when that file is successfully loaded and on a slow connection for example, that could take 1+ seconds, but the code will keep executing and we'd have problems:

```js
$.getJSON(DDG.get_asset_path('github', 'fatheads.json'), function (data) { 
    var fatheads = data;
    DDG.require("moment.js", function() { ...
    });
});
```

jdorweiler:
@Snapy 
>1. Why did you completely remove the possibility to display more results ? Sometimes it's useful (e.g. for projects where forks are better or still maintained) so if you could consider reimplementing this feature it would be nice or give an option to show forks or similar projects.

Do you have an example? The current multiple tile results don't seem very useful to me. I generally just skip the IA and click through the organics below. I could see there being cases where multiple results are better though. 

>3. This is really great and all the info is useful.
I don't know if it's possible but if the same can be done for organizations it would be a good addition.

That's a good idea. I was thinking of making issues to expand the IA further to do organizations, users, and maybe even code search.

Snapy:
Such an example I can think of is uBlock : the one from gorhill (5670 :star:) is the maintained and legitimate version whereas the one from chrisaljoudi (7763 :star:) is not maintained anymore but is displayed by default (maybe because more people starred the project).
Also still with uBlock I've remarked that it triggers both the about IA ([link](https://beta.duckduckgo.com/?q=github+ublock&ia=about)) which is the legitimate repository and the software IA ([link](https://beta.duckduckgo.com/?q=github+ublock&ia=software)) which is about the unmaintained chrisaljoudi repository.
Other examples are privacy badger (with Chrome and Firefox repositories but defaults to Firefox) or HTTPS Everywhere which also triggers the about IA with just the repo description and the software IA (by default) which has now more information.
Also there is PR #2520 that should add repo dev language when/if it is implemented.
So if you can reconsider parts of this PR to take into account everything it would make DDG an even greater resource for devs. Thanks !

jdorweiler:
@Snapy Thanks for the feedback!
I added multiple results as long as the number of stars is within 70% of the top result. 70% was arbitrarily chosen just to work for the ublock case. I still need to work on the tile view. 
selection_393


The privacy badger case is harder to handle. The api request is specifically looking for the firefox version since this is the top scoring organic. We'll have to leave that as an issue for now.

share/spice/github/fatheads.json
jdorweiler: Uppercase actually. It's the exact repo name. 
YAML sounds good.

jdorweiler:
Ready to go again.
- Showing multiple results 
https://beta.duckduckgo.com/?q=ublock+repo&ia=software

- Single result
https://beta.duckduckgo.com/?q=font+awesome+github&ia=software

Snapy:
Great !
Works also nicely when searching for privacy badger. I don't see what changed with your latest commit ( 8d203ce).

share/spice/github/fatheads.json
jdorweiler: Do you know an easy way to read YAML in JS? If not lets just leave it as JSON for now to keep the PR moving.

jdorweiler:
@moollaza ready for review again.

share/spice/github/fatheads.json
moollaza: Whoops, I was thinking this was being used by Perl. JSON is good 👍

moollaza:
@jdorweiler Thanks! This LGTM :+1:

I think we could move some of the important info to a subtitle, but this is good for now.

E.g.

> **Skeleton**
> Updated 8 days ago  |  12138 Stars  |  2159 Forks
>
> Skeleton: A Dead Simple, Responsive Boilerplate for Mobile-Friendly Development

PR #2923, opened 2016-07-26
MetaCPAN: Migrate API v0 to v1

### LINKED ISSUE #898: MetaCPAN - Switch to new API, show more data
Currently we are using an older API which only surfaces information for a given module. The API we're using is also case sensitive,  so "app::duckpan" will not show any results for "App::DuckPAN": https://duckduckgo.com/?q=cpanm+App%3A%3ADuckPAN, https://duckduckgo.com/?q=cpanm+app%3A%3Aduckpan.

It seems that the API has since been updated allowing for a general search endpoint:
http://api.metacpan.org/v0/release/_search?q=cpanminus

I suggest we switch to the new API to allow for true "search" of MetaCPAN Packages. In the event that the search does not exactly match a module name we could return a list of results and in the event that a module name is correctly indicated we should only show that result.

The API currently provides tons of info, most of which we probably don't need. You can choose which elements to return by indicating, for example `&fields=name,id` in the URL. `&fields=version,name,author,date,distribution` seems like a good start :)

E.g. http://api.metacpan.org/v0/release/_search?q=string%20trim&fields=name,abstract,version,metadata,author,download_url,date,distribution

This doc provides some insight into all the available fields:https://github.com/CPAN-API/cpan-api/blob/master/lib/MetaCPAN/Document/File.pm

https://duck.co/ia/view/meta_cpan

gabriellhrn:
###### Upgrade MetaCPAN API from v0 to v1

- [x] Improvement
    - [x] Enhancement: **`MetaCPAN: Migrate API v0 to v1`**

###### Which issues does this fix?

Fixes #898 - Switches to new API

###### People to notify

@jbarrett 

###### Work in progress

Looks like we need to sort the results somehow. The most relevant results are not displayed first.

---

Instant Answer Page: https://duck.co/ia/view/meta_cpan


daxtheduck:
@gabriellhrn Hey!
        
Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

jbarrett:
Thanks @gabriellhrn! Hmmm, that is not what I expected...

cpan_json

Looks like we'll need to a little more work on the search clauses... For v0 I basically [pulled them from metacpan-web](https://gist.github.com/jbarrett/b9acf1a415bfe846c03c69b7f08eead3).

Here's the [newer script to generate the v1 search](https://gist.github.com/jbarrett/ebf3da8065215049f1eb737098179f38) which attempts the same approach, which I guess doesn't work :smile: 

Should probably port the HTTP Request parts over to the second script so we can get a feedback loop going.

The aggregate and sort parts would need to be done in js, but I'm hoping we don't have to go there since CPAN search would hopefully become a secondary function of this spice, with showing docs being the primary.

gabriellhrn:
@jbarrett's changes fixed how the results are displayed. Thanks :smile: 

captura de tela de 2016-07-27 21-43-56

jbarrett:
Thanks @gabriellhrn! I'm happy enough to merge this as-is, we can move on to thinking about what to do for exact matches and clicks.

Thoughts, @sahildua2305?

MariagraziaAlastra:
Looks good, thanks @gabriellhrn and @jbarrett! :+1:

PR #2949, opened 2016-08-08
New Skyscanner Spice

samheyman:

First Skyscanner spice that will show users the best deals that Skyscanner has from their country to other countries.

###### People to notify (@mention interested parties)
@moollaza 

---

Instant Answer Page:https://duck.co/ia/view/skyscanner_flights
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @samheyman


daxtheduck:
@samheyman Hey!
        
Thanks!

daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
```diff
@@ -0,0 +1,154 @@
...
+                flight_quote: quote_ref,
+                flight_datetime: quote_date_time,
+                flight_outbound_date: outbound_date,
+                flight_return_date: return_date
+            };
+        }
+        var settings = {
+            "async": true,
+            "crossDomain": true,
+            "url": "https://gateway.skyscanner.net/travel-api/v1/entities?external_ids=" + listOfCountryIds + "&enhancers=images&apikey=09fd8de5844d4b1d982a320ad5dee5b8",
```
moollaza: @samheyman we can't make this API call directly to SkyScanner from the Client as we consider that a Privacy breach (their IP would be exposed)

Fortunately there's an easy fix! You can create an additional endpoint in the Perl, which let's you define another API endpoint to be proxied.

```perl
spice alt_to => {
    sky_scanner_entities => {
        to => 'https://gateway.skyscanner.net/travel-api/v1/entities?external_ids=$1&enhancers=images&apikey={{ENV{DDG_SPICE_SKYSCANNER_APIKEY}}}'
        proxy_cache_valid => '418 1d' #disable api response caching, if needed
    }
}
```

Then from the JS you can do

```js
$.getJSON('/js/spice/sky_scanner_entities/' + listOfCountryIds, function(data) {
    // do work
    DDG.require('moment.js', function() {
        Spice.add({ ... });
    });
});
```

samheyman:
@moollaza I've moved the api call to the perl script. Thanks for your help!

moollaza:
@samheyman this is now live on Beta! https://beta.duckduckgo.com/?q=Skyscanner&ia=flights

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,39 @@
...
+use DDG::Spice;
+use strict;
+use warnings;
+
+# Caching - http://docs.duckduckhack.com/backend-reference/api-reference.html#caching
+spice is_cached => 1;
+spice proxy_cache_valid => '200 1d'; # defaults to this automatically
+
+spice wrap_jsonp_callback => 1; # only enable for non-JSONP APIs (i.e. no &callback= parameter)
+spice to => 'http://partners.api.skyscanner.net/apiservices/browseroutes/v1.0/GB/GBP/en-GB/$1/anywhere/anytime/anytime?apikey={{ENV{DDG_SPICE_SKYSCANNER_APIKEY}}}';
```
moollaza: Seeing as we can get the user's locale/country via the `$loc` I think it would be good to specify the locale here to get localized currency?

I'm seeing prices in Pounds which doesn't mean too much to me in Canada...

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,39 @@
...
+spice alt_to => {
+    skyscanner_images => {
+        to => 'https://gateway.skyscanner.net/travel-api/v1/entities?external_ids=$1&enhancers=images&apikey={{ENV{DDG_SPICE_SKYSCANNER_IMAGES_APIKEY}}}'
+        # proxy_cache_valid => '418 1d' #disable api response caching, if needed
+    }
+};
+
+# Triggers - https://duck.co/duckduckhack/spice_triggers
+#triggers any => 'skyscanner flights to', 'skycanner show me flights to ', 'skyscanner inspire me';
+triggers any => 'skyscanner', 'skycanner flights', 'skyscanner inspire me', 'skyscanner deals';
```
moollaza: Let's use `triggers startend` -- `any` is too broad

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,39 @@
...
+
+# Handle statement
+handle remainder => sub {
+    # get user's location for default origin
```
moollaza: Let's ensure there is no remainder, meaning the query is identical the trigger phrase.

```perl
handle remainder => sub {
    return if $_;
    ...
}
```

share/spice/skyscanner_flight_search/footer.handlebars
```diff
@@ -0,0 +1,6 @@
+<div class="pull-left" style="font-size:1.2em">
```
moollaza: Let's use one of our `tx--##` font-size classes from https://ddg.gg/styleguide

e.g. `tx--21`

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
```diff
@@ -0,0 +1,155 @@
...
+                flight_destination_airport: destination_airport,
+                flight_destination_code: destination_code,
+                flight_origin_code: origin_code,
+                flight_quote: quote_ref,
+                flight_datetime: quote_date_time,
+                flight_outbound_date: outbound_date,
+                flight_return_date: return_date
+            };
+        }
+        var settings = {
```
moollaza: Looks like we can remove this?

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
```diff
@@ -0,0 +1,155 @@
...
+                    }
+                },
+
+                normalize: function(item) {
+                    return {
+                        disclaimer: "found " + moment(item.flight_price_age).fromNow() + " ago",
+                        title: item.flight_destination_city,
+                        image: item.flight_destination_city_image,
+                        url: "http://partners.api.skyscanner.net/apiservices/referral/v1.0/GB/GBP/en-GB/" + item.flight_origin_code + "/" + item.flight_destination_code + "/" 
+                            + moment(item.flight_outbound_date).format("YYYY-MM-DD") + "/" + moment(item.flight_return_date).format("YYYY-MM-DD") + "?apiKey=te1561648834359",
```
moollaza: Can we remove the API key here? I'd prefer to keep our API keys hidden

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
```diff
@@ -0,0 +1,155 @@
...
+        };
+
+        //$.ajax(settings).success(function (response) {
+          
+        $.getJSON('/js/spice/skyscanner_images/' + listOfCountryIds, function(response) {
+            var flights = listOfRoutes.map(function(route,index) { return build_flight_route(route, index, response); });
+            
+            // Fail if no flights to show
+            if (flights.length === 0) {
+                return Spice.failed('No flights found');
```
moollaza: The input here needs to be the Spice's `id`, not a message :)

e.g. `return Spice.failed('skyscanner_flight_search")`

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
samheyman: That's only a short version of the api key (16 chars instead of the full 32) to avoid exposing the full api key in the redirect url. I can hide this short api key in the same way as the full api key but felt it was a bit of an overkill as it cannot be used for anything else. Would you prefer me to add it as {{ENV{DDG_SPICE_SKYSCANNER_SHORT_APIKEY}}}?

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
samheyman: yep, indeed settings are no longer needed.

share/spice/skyscanner_flight_search/footer.handlebars
samheyman: yes, good idea.

lib/DDG/Spice/SkyscannerFlightSearch.pm
samheyman: Yes, that's a very good point. I'll have a chat with my colleagues how best to do it.

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
samheyman: ah, yes I see.

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
moollaza: > That's only a short version of the api key (16 chars instead of the full 32) to avoid exposing the full api key in the redirect url.

Oh, okay I hadn't noticed that. I'm more comfortable knowing it's not the full API key 👍 

> Would you prefer me to add it as {{ENV{DDG_SPICE_SKYSCANNER_SHORT_APIKEY}}}?

Unfortunately that approach won't work on the frontend, we don't have any way to hide API keys, but so far we haven't needed to.

@samheyman does this URL require our API key? It seems to work without that portion of the URL? Is this for SkyScanner to validate the queries are coming from a valid source?

lib/DDG/Spice/SkyscannerFlightSearch.pm
moollaza: To clarify, what I meant, you can `return` multiple values in the `handle` and for example if the locale is the 2nd value, you can replace `en-GB` with `$2`. You can pass along as many values as needed.

When you're passing 2+ values though, you need to define a `spice from` regex to capture them.

For two values, `spice from => "/([^\/]+)/([^\/]+)"` should work, for three, `spice from => "/([^\/]+)/([^\/]+)/([^\/]+)`, etc.

Should be lots of examples in the repo 👍

share/spice/skyscanner_flight_search/skyscanner_flight_search.js
samheyman: Yes, it's to allow us to keep track of all our incoming traffic.

lib/DDG/Spice/SkyscannerFlightSearch.pm
samheyman: Yep, I was looking into that for the next step with origin, destination and dates.
The issue here is that your service returns a slight different format of locale as far as I can see (en_GB) whereas our input is en-GB. I need to insure there is a 1-1 match for all locales (i.e. that we support all your possible values) and then replace the _ with -.
For the currencies I need to find a way to map the locale to the currency. We have an API endpoint that returns all the currencies that we support but I can't remember how we select currencies based on the locale. I'll ask my colleagues working on the APIs tomorrow morning. 

And, yes, lots of good examples in the repo - and I'm slowly getting my head around perl :-)

moollaza: @samheyman thanks for the explanation!

I would suggest we create a mapping of locales -> currencies with a YAML/JSON file that we store in the /share/ directory. Then it can be loaded in the Perl to determine which currency we should use. I suppose we can fallback to USD as it's the most popular currency?

You may be able to find/use another Perl module to map locales in our format to another

samheyman:
@moollaza let me know when you have deployed and I'll test. Not sure why the currency and destination lookups are failing.
`http://partners.api.skyscanner.net/apiservices/referral/v1.0/GB/USD/en-GB/UK/undefined/2016-11-14/2016-11-25?apiKey=te1561648834359`

moollaza:
@samheyman deployed. Everything looks to be working correctly 👍

samheyman:
Great thanks Zaahir. I've realised it's because I only had 
"UK": "GBP" 
in the market to currency mapping file.
I've added 
"GB": "GBP" 
and now I see UK prices in £.
I've committed the changes so that should be us done. For now :-)

moollaza:
@samheyman changes should be live. Can you please confirm?

samheyman:
Yes, that seemed to have fixed the currency issue. However there seems to be an issue with loading the cities on beta.
Locally if I search "skyscanner flights from canada to paris" I get prices from Canada to Paris
On beta it returns prices to different countries (using "anywhere"). Was the citySkyscannerCodes.json file also pushed?

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
...
+# Load the list of cities and their Skyscanner code (using Skyscanner code as IATA code is ambiguous in certain cases such as Berlin where the city 
+# and one of the airports have the same code)
+my $cities_raw = share('citySkyscannerCodes.json')->slurp;
+my $cities = decode_json($cities_raw);
+
+# Handle statement
+handle remainder => sub {
+    my $origin = "";
+    my $destination = "";
+
```
moollaza: We need to verify the `$loc`, `$loc->country_code`, `$lang`, `$lang->locale` and are defined before using them. They're not guarenteed to be defined:

`return unless $loc && $lang && ..`

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
...
+        
+    # get language locale (replace DDG's '_' with '-' for Skyscanner compatibility), if none default to 'en-US'
+    my $locale = $lang->locale;
+    $locale =~ tr/_/-/;
+    if ($locale eq "") {
+        $locale = "en-US";
+    }   
+    
+    # get currency from the json file using the market, if none default to USD
+    my $currency = $currencies->{$market};
```
moollaza: `my $currency = $currencies->{$market} // ""`

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
...
+my $cities_raw = share('citySkyscannerCodes.json')->slurp;
+my $cities = decode_json($cities_raw);
+
+# Handle statement
+handle remainder => sub {
+    my $origin = "";
+    my $destination = "";
+    
+    # get user's location for the market if available (airline and travel agent prices depend on the market), if none default to 'US'
+    my $market = $loc->country_code;
```
moollaza: `my $market = $loc->country_code // "US"`

the `//` gracefully handles the `undef` case

lib/DDG/Spice/SkyscannerFlightSearch.pm
samheyman: The return unless breaks the flow of my program. Can I do it like this?
` my $market = "";
    
    # get user's location for the market if available (airline and travel agent prices depend on the market), if none default to 'US' 
    if ($loc->country_code) {
       $market = $loc->country_code;
    } else {}
        $market = "US";
    }`

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
...
+    print "\n\nMarket: " . $market;
+    print "\nCurrency: " . $currency . "\n\n";
+
+    # query must be in the form 
+    # [from][origin][to][destination]
+    # or [origin][to][destination]
+    # or [from][origin]
+    # or [destination]
+    # 
+    my @query = split(/\s+to\s+/, $_);
```
moollaza: we're assuming the query contains "to" here -- I think we need to handle the cases where it doesn't? 

`return unless $query =~ m/\sto\s/;`

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
...
+    #print "\nOrigin: " . $query[0];
+    #print "\nDestination: " . $query[1];
+    #print "\n*********************\n\n";
+
+    # determine origin country or city (use lowercase), if no match use user's country
+    $origin = lc($query[0]);
+    if ($origin eq "") {
+        $origin = $market;
+    } else {
+        if (exists $countries->{$origin}) {
```
moollaza: `$origin = $countries->{$origin} // $market;`

moollaza: I believe you can do `$origin = $countries->{$origin} // $cities->{$origin} // $market`

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
...
+            $origin = $market;
+        }
+    } 
+    
+    # determine destination country or city (use lowercase), if no match use 'anywhere'
+    $destination = lc($query[1]);
+    if ($destination eq "") {
+        $destination = "anywhere";
+    } else {
+        if (exists $countries->{$destination}) {
```
moollaza: Same here `$a // $b // $default;`

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
...
+            $destination = $countries->{$destination};
+        } elsif (exists $cities->{$destination}) {
+            $destination = $cities->{$destination};
+        } else {
+            $destination = "anywhere";
+        }
+    } 
+    
+    return $market, $locale, $currency, $origin, $destination;
+    return;
```
moollaza: this line will never be reached because of the `return` above

lib/DDG/Spice/SkyscannerFlightSearch.pm
```diff
@@ -0,0 +1,113 @@
+package DDG::Spice::SkyscannerFlightSearch;
+
+# ABSTRACT: This IA will call Skyscanner's API to retrieve the cheapest flight prices to 
+# different countries from the user's country
+
+use DDG::Spice;
+use strict;
+use warnings;
+use JSON;
```
moollaza: `use JSON::MaybeXS`

lib/DDG/Spice/SkyscannerFlightSearch.pm
samheyman: but does that not mean that we just exit and not run the call if there is no "to"?
the tests run so I assumed it was ok:
```

# if origin specified then query should use it
    'skyscanner flights from United Kingdom' => expected_output('US/en-US/USD/gb/anywhere'),
     
    # if destination specified then query should use it
    'skyscanner flights to United States' => expected_output('US/en-US/USD/US/us'),
```

lib/DDG/Spice/SkyscannerFlightSearch.pm
moollaza: Hmm I'm not sure we should default to US if we don't know the user's location. I think I might consider that an irrelevant result?

Regardless, we ned to check `$loc` exists before checking `$loc->country_code`

samheyman: is this ok for checking that $loc exists?
`my $market = $loc->country_code // "US";`

samheyman: Agree that defaulting to US is not ideal but I can't think of another solution other than using the IP (if you guys support that). Do you have any logs to see how often the $loc->country is not available? If it's infrequent then it's not too much of an issue.

moollaza: @samheyman I don't know how often the $loc is not available. I should be there for the majority of places.

In cases where we don't have it, we should just ignore the query though. All major cities and countries are supported as far as I know

PR #3046, opened 2016-11-15
Use translated token for the name of the maps IA in Spice.add()

bsstoner:
The token is for "Maps", not "maps", and it unfortunately has a context of "feedback form". This will force it to use the existing translation for the Maps IA link.

https://duck.co/translate/token/1346

I wish we could just remove the context on the existing token, but this works. @andrey-p would removing the "feedback form" context break the feedback form? Does all the text there have that context?

/cc @bbraithwaite

### DIFF
--- a/share/spice/maps/maps/maps_maps.js
+++ b/share/spice/maps/maps/maps_maps.js
@@ -15,7 +15,7 @@ DDG.require('maps',function(){
         Spice.add({
             data: response,
             id: "maps_maps",
-            name: lp("feedback form", "Maps",
+            name: lp("feedback form", "Maps"),
             model: "Place"
         });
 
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

andrey-p:
@bsstoner Yeahhh, as far as I know removing the context will basically make it a new token.

How about we do the following:

- merge in this change so the Maps token isn't untranslated anymore
- create a separate token without the context, so that a translation can be added
- at some point in the future, swap them over

andrey-p:
This isn't working for me BTW, tried in Finnish which is translated:

screenshot from 2016-11-16 12-40-05

Isn't that `name` param passed to the `lp` function already? Wouldn't it get double wrapped, and does that work?

bsstoner:
@andrey-p Is "Maps" even in the feedback form? I don't see it. Can I just remove the context now?

screen shot 2016-11-16 at 8 11 52 am

bsstoner:
Closing, this isn't the right way to solve it regardless. The name is getting pulled from meta data.

PR #3097, opened 2017-01-17
Flights: Trim query before splitting on \s to ensure we have an airline in the query

moollaza:
This handles a bug where a list of all airport codes is returned when one is not specified in the query.

The bug occurs because we split the query on `\s`, but we weren't **_trimming_** the query after replacing words in it. Thus, the check for `scalar(@query) > 2` in `identifycodes` passes, but we don't actually have two words, we have an array: `[ '', 'word2' ]`.

## People to notify
@jdorweiler 

---

Instant Answer Page: https://duck.co/ia/view/route

### DIFF
--- a/lib/DDG/Spice/Flights/Route.pm
+++ b/lib/DDG/Spice/Flights/Route.pm
@@ -3,6 +3,7 @@ package DDG::Spice::Flights::Route;
 
 use strict;
 use DDG::Spice;
+use Text::Trim;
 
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

jdorweiler:
:+1:

PR #3122, opened 2017-01-27
IsItUp: Add support for 'is it down' forms

rasikapohankar:
## Description of new Instant Answer, or changes

Adds support for triggers such as:
1. `isitdown`, 'is it down'
2. `isitup`, `is it up`
3. `is X up right now`
4. `status of X`

## Related Issues and Discussions

Fixes #2837 (finish #2955)

## People to notify

@GuiltyDolphin

---

Instant Answer Page: https://duck.co/ia/view/is_it_up

PR #3135, opened 2017-02-08
IsItUp: Return if query does not match regex

### LINKED ISSUE #3136: IsItUp: False positive triggering
It shouldn't trigger here:
https://duckduckgo.com/?q=t-online.de

screen shot 2017-02-11 at 8 06 02 am

------
IA Page:  http://duck.co/ia/view/is_it_up 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @rasikapohankar

rasikapohankar:

Tiny change, return if query does not match the regex

Please refer [comment](https://github.com/duckduckgo/zeroclickinfo-spice/pull/3122#discussion_r99644571)
Fix #3136 

## People to notify

@moollaza @GuiltyDolphin 

---

Instant Answer Page: https://duck.co/ia/view/is_it_up

### DIFF
--- a/lib/DDG/Spice/IsItUp.pm
+++ b/lib/DDG/Spice/IsItUp.pm
@@ -28,7 +28,7 @@ handle query_lc => sub {
 
     my $query = $_;
 
-    $query =~ $query_regex;
+    return unless $query =~ $query_regex;
 
     $root_url = $2;
 
--- a/lib/DDG/Spice/IsItUp.pm
+++ b/lib/DDG/Spice/IsItUp.pm
@@ -15,7 +15,7 @@ spice to => 'https://isitup.org/$1.json?callback={{callback}}';
 
 spice proxy_cache_valid => "418 1d";
 
-my $query_start_regex = qr/(?:is|isitup|isitdown|is it up|is it down|status of|)/;
+my $query_start_regex = qr/(?:is\s|isitup|isitdown|is it up|is it down|status of|)/;
 my $query_end_regex   = qr/(?:up|down|working|online|status|up right now|)/;
 my $url_regex         = qr/(?:https?:\/\/)?([\p{Alnum}\-]+(?:\.[\p{Alnum}\-]+)*?)(?:(\.\pL{2,})|)/;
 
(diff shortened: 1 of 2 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

rasikapohankar:
@moollaza @GuiltyDolphin I added [this](https://github.com/duckduckgo/zeroclickinfo-spice/pull/3135/commits/d5f2a2304160da21306b0cd4415a79e9d8e4f50a) for #3136. Can you please review?

It is triggering on any website name containing the trigger words. My understanding was that it triggers only when the trigger is a word by itself, so something like `t-online.de` containing the trigger word `online`, will not trigger even if the regex is allowing it. Did I miss something? :confused:

rasikapohankar:
@moollaza Fixing #3136 here, can you please review?

moollaza:
> Did I miss something?

No, this is technically the expected behaviour, but it has unexpected results. Internally we break the query up into "words" by separating it not only on spaces, but also "-" and "."

So the triggering system sees this as `t online de`. However the `remainder` in this case is going to be empty because removing the "trigger" means removing "t-online.de" from the query.

This is documented in https://github.com/duckduckgo/duckduckgo/issues/122

It seems like one other issue with this IA is that the `start`, `end` and `query` regex all end with `|)` which I think makes it match on an empty string...

Let's update the tests to confirm this works as expected 👍

rasikapohankar:
@moollaza Added some tests for the changes

rasikapohankar:
Thanks, I did not know about duckduckgo/duckduckgo#122

@moollaza I added `| )` at the end so that it could match on queries which can have words at the beginning or end or both, like, `is X up` and `isitup X`. Its needed because the IA originally checks for presence of a word at the start [here](https://github.com/rasikapohankar/zeroclickinfo-spice/blob/b269e727fed77cf2af2a183c7301d41592cac580/lib/DDG/Spice/IsItUp.pm#L52). Do you have any suggestions for doing it in a different way?

Also added back the `\s` after `is` in the regex, which was matching `isitdown` and making the root url `itdown`

moollaza:
Deploying to Beta 👍

bsstoner:
This should be in production now 🚀

PR #3143, opened 2017-02-14
JavaScript: Switch NPM to new API and tile view

### LINKED ISSUE #3098: JavaScript: Improving the NPM Package Search
It seems that this is a popular IA, but is failing to trigger on a lot of queries in the search space. It currently represents [1.81% of the search space](https://forum.duckduckhack.com/t/javascript-search-overview/94/14). We can improve this. I think this is because of:

  1) Triggers
  2) Spaces in package names
  3) People using DDG as if it were npm's onsite search engine.

The final point is important here. It seems there aren't a lot of queries looking for something specific, but rather a suggestion. ie. `json pretty print npm` or `npm json web token`. If we update the API endpoint to the below link, this should solve this problem and the specifics will still show up.

http://npmsearch.com/query?q=json+web+tokens&fields=name,description,repository,homepage,author

Documentation: https://github.com/nodesource/npmsearch

This should make it work just like the [CPAN spice](https://duckduckgo.com/?q=metacpan+WWW%3A%3ADuckDuckGo&ia=software&iai=9). 

One thing I have noticed about this API though, is it doesn't direct a link back to npm, rather both *homepage* and *repository* return the github source which is no good. Appending the package name to the URL `https://www.npmjs.com/package/` will redirect that though. 

------
IA Page:  https://duck.co/ia/view/npm
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @remixz

adampolar:
Changes from a single detail response to a search style response

Fixes #3098 
Note that this leaves the README.md with an image of the previous incarnation of the npm spice

## People to notify

@pjhampton

---

Instant Answer Page: https://duck.co/ia/view/npm

### DIFF
--- a/lib/DDG/Spice/Npm.pm
+++ b/lib/DDG/Spice/Npm.pm
@@ -1,13 +1,17 @@
 package DDG::Spice::Npm;
-# ABSTRACT: Returns package information from npm package manager's registry.
+# ABSTRACT: Search npm for package information
 
 use strict;
 use DDG::Spice;
+use warnings;
 
-triggers startend => 'npm', 'nodejs';
+spice is_cached => 1;
+spice proxy_cache_valid => '200 1d';
+
+triggers startend => 'npm', 'nodejs', 'node package';
 triggers start => 'npm install';
 
-spice to => 'http://registry.npmjs.org/$1/latest';
+spice to => 'http://npmsearch.com/query?q=$1&fields=name,description,repository,homepage,author,rating,version';
 spice wrap_jsonp_callback => 1;
 
 handle remainder_lc => sub {
--- a/lib/DDG/Spice/Npm.pm
+++ b/lib/DDG/Spice/Npm.pm
@@ -8,7 +8,7 @@ use warnings;
 spice is_cached => 1;
 spice proxy_cache_valid => '200 1d';
 
-triggers startend => 'npm', 'nodejs', 'node package';
+triggers startend => 'npm', 'nodejs', 'node js', 'node package';
 triggers start => 'npm install';
 
 spice to => 'https://api.npms.io/v2/search?q=$1';
--- a/lib/DDG/Spice/Npm.pm
+++ b/lib/DDG/Spice/Npm.pm
@@ -11,7 +11,7 @@ spice proxy_cache_valid => '200 1d';
 triggers startend => 'npm', 'nodejs', 'node package';
 triggers start => 'npm install';
 
-spice to => 'http://npmsearch.com/query?q=$1&fields=name,description,repository,homepage,author,rating,version';
+spice to => 'https://api.npms.io/v2/search?q=$1';
 spice wrap_jsonp_callback => 1;
 
 handle remainder_lc => sub {
--- a/share/spice/npm/npm.js
+++ b/share/spice/npm/npm.js
@@ -23,18 +23,20 @@
             data: api_result.results,
             meta: {
                 sourceName: "npmjs",
-                itemType: (api_result.results.length === 1) ? 'npm package' : 'npm packages',
+                itemType: (api_result.total === 1) ? 'npm package' : 'npm packages',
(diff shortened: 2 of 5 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

adampolar:
Hi, sorry for the late change. I realised that the author is not always present in the package object and publisher looks like it is, so I decided to use that for the footer instead (this also matches the prominent display of the publisher on npmjs.com).

Another issue I wanted to ask about is the query stripping area of the code. I stole that from metacpan spice and slapped it in my own (I couldn't find a common area to use in this framework) but am a bit unsure that it's the right thing to do. In particular as it's digging out user input could it potentially be used for XSS attacks? Please could someone offer guidance on this?

moollaza:
> I realised that the author is not always present in the package object and publisher looks like it is, so I decided to use that for the footer instead (this also matches the prominent display of the publisher on npmjs.com).

@adampolar That sounds fine 👍 

> but am a bit unsure that it's the right thing to do

Are you referring to lines 9-11?  If so, that's perfectly fine 👍

adampolar:
> Are you referring to lines 9-11? If so, that's perfectly fine :+1:

@moollaza Yep that's it. Great thanks for the assurance.

lib/DDG/Spice/Npm.pm
```diff
@@ -1,13 +1,17 @@
...
 
 use strict;
 use DDG::Spice;
+use warnings;
 
-triggers startend => 'npm', 'nodejs';
+spice is_cached => 1;
+spice proxy_cache_valid => '200 1d';
+
+triggers startend => 'npm', 'nodejs', 'node package';
```
moollaza: @adampolar please add `node js` -- I see that appearing a lot in the autocomplete suggestions!

share/spice/npm/npm.js
```diff
@@ -2,96 +2,55 @@
...
+                
                 return {
-                    title: item.name + " " + item.version,
-                    subtitle: item.description,
-                    infoboxData: boxData,
+                    title: item.package.name,
+                    subtitle: "version: " + item.package.version,
+                    description: item.package.description,
+                    url: item.package.links.npm,
+                    rating: item.score.final.toFixed(2),
```
moollaza: @adampolar it seems like we should be sorting results in order of decreasing rating?

I'm wondering if this is actually something we need to show to the user though because its an npms.io concept and it may not be clear what "score" means or where it's coming from.

For example, https://npms.io/search?q=wikipedia seems to have a 0.06 score before a 0.66 -- that seems like a sorting error we can handle on our end?

lib/DDG/Spice/Npm.pm
adampolar: @moollaza Will do!

share/spice/npm/npm.js
adampolar: @moollaza 

> it seems like we should be sorting results in order of decreasing rating?

I'm not sure, I don't think it is a "search relevancy score", the score is actually about how well maintained etc. a package is.  it might throw up some weird ones, for example, mocha would then have enzyme as its first result. (see https://npms.io/search?q=mocha). The search relevancy score is provided by the searchScore value.

> I'm wondering if this is actually something we need to show to the user though because its an npms.io concept and it may not be clear what "score" means or where it's coming from.

I agree with this. Do you think the date parameter might be more relevant here?

moollaza:
Deploying changes...

moollaza:
@adampolar Thanks! This LGTM :+1:

Excellent work 💯

adampolar:
Thanks @moollaza ! Do I need to update the page at  https://duck.co/ia/view/npm?

moollaza:
> Thanks @moollaza ! Do I need to update the page at https://duck.co/ia/view/npm?

Nope!

PR #3211, opened 2017-03-27
Package Tracking: Add Fedex, UPS, USPS Regex Triggers

moollaza:
The existing Fedex, UPS, and USPS Goodies constantly return false-positive results because we rely on our regex to determine if a tracking code belong to one courier or another.

This PR migrates those triggers to the Package Spice so we can trigger on queries that only contain alphanumeric strings that match Fedex, USPS, and UPS tracking number patterns. The API will determine if the number is valid and which carrier it actually belongs to and provide us with the current package status.

This is way more reliable and will prevent us from sending the user to the wrong courier website.

This will also enable us to deprecate those 3 Goodies.

Partially resolve https://github.com/duckduckgo/zeroclickinfo-goodies/issues/3973

## People to notify
@Owlree @pjhampton 

---

Instant Answer Page: https://duck.co/ia/view/package_tracking

### DIFF
--- a/t/PackageTracking.t
+++ b/t/PackageTracking.t
@@ -87,14 +87,16 @@ ddg_spice_test(
     70000000000000000001 => undef,
     'what is 70000000000000000000' => undef,
 
+    # Too short
+    'usps 12345' => undef,
 
     # TODO: Display input form prompting for tracking number?
     'fedex package tracking' => undef,
+    'fedex package tracker' => undef,
     'package tracking' => undef,
     'package tracking online' => undef,
 
     'fedex' => undef,
-    'fedex website' => undef,
     'fedex website' => undef
 );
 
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

t/PackageTracking.t
```diff
@@ -9,31 +9,84 @@ spice is_cached => 1;
...
+    ),
+    'usps 37346365253153' => test_spice(
+        '/js/spice/package_tracking/37346365253153',
+        call_type => 'include',
+        caller => 'DDG::Spice::PackageTracking'
+    ),
+
+    # Bad Queries
+    70000000000000000001 => undef,
+    'what is 70000000000000000000' => undef,
```
pjhampton: I think we need to be a bit more aggressive here with the bad queries. neither of these have trigger words. Can we add a couple of more with a couple of the trigger words, such as `usps 373463`?

pjhampton:
Apart from that nit, this is good to ship

PR #3297, opened 2017-05-22
Maps: add zip code skip words, allow low relevance results through when DDG.isRelevant is true.

dharb:
We needed to add a couple skip words to the array to improve results when users are searching by zip or postal code. I think the best way to explain is by example:

Searching '19087 zip code' gives a high level map signal. However, the top result from mapbox has place_name: "19087, Pennsylvania, United States" and relevance: 0.49, which is below our filtering threshold. This is where the DDG.isRelevant method comes into play: it compares the query (minus a selection of skip words) to the place name of the top result for similarity. By adding these skip words to that array, DDG.isRelevant will return true, which overrides the relevance # we get from mapbox.

We then need to make sure that the top result isn't filtered out due to its relevance being < 0.6, but that disambiguation items with low relevance are properly filtered out.

https://app.asana.com/0/0/334903345804983

## People to notify
@moollaza 

---

Instant Answer Page: https://duck.co/ia/view/maps_maps

### DIFF
--- a/share/spice/maps/maps/maps_maps.js
+++ b/share/spice/maps/maps/maps_maps.js
@@ -1,6 +1,7 @@
 DDG.require('maps',function(){
     ddg_spice_maps_maps = function(response) {
-        var skipArray = [ "directions", "map", "maps", "st", "street", "ave", "avenue", "dr", "drive", "pl", "place", "apt", "suite", "latitude", "longitude" ];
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

moollaza:
This is live on Beta: https://beta.duckduckgo.com/?q=19087+zip+code&t=hd&ia=web

PR #3329, opened 2017-06-13
Currrency: Swapped API versions

pjhampton:
# PLEASE DON'T MERGE ME - WIP 🎸 

This swaps out the old XE API with the new one. **The end user shouldn't notice a difference**. 

## People to notify

@moollaza 

---

Instant Answer Page: https://duck.co/ia/view/currency

### DIFF
--- a/share/spice/currency/currency.js
+++ b/share/spice/currency/currency.js
@@ -382,17 +382,11 @@
                     $more_at_link_charts = $currency.find("#zci__more-at--info");
 
                     // apends all the currency names to the selects
-                    for( var i = 0 ; i < currencies.length ; i ++ ) {
-                        $selects.append(
-                            "<option value=" 
-                            + currencies[i].symbol 
-                            + ">"
-                            + currencies[i].name 
-                            + "</strong> (" 
-                            + currencies[i].symbol 
-                            + ")" 
-                            + "</option>"
-                        );
+                    for ( var i = 0 ; i < currencies.length ; i ++ ) {
+                        $('<option>', {
+                            value: currencies[i].symbol,
+                            text: currencies[i].name + " (" + currencies[i].symbol + ")"
+                        }).appendTo($selects);
                     }
 
                     // sets the default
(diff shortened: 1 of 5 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/spice/currency/currency.js
```diff
@@ -286,14 +278,17 @@
...
             }
 
-            var endpoint = "/js/spice/currency/1/" + from + "/" + to;
-            $.get(endpoint, function(payload) {
-
-                // jsonp is returned from the API so we have to alter the contents
+            $.ajax({
+                url: "/js/spice/currency/1/" + from + "/" + to,
+                beforeSend: function( xhr ) {
+                    xhr.overrideMimeType( "text/plain; charset=x-user-defined" );
```
moollaza: What is this for?

share/spice/currency/currency.js
```diff
@@ -286,14 +278,17 @@
...
+                },
+                error: function(request, status, error) {
+                    Spice.failed('currency');
+                }
+            }).done(function(payload) {
                 var response = JSON.parse(payload.trim().replace(/^[^\(]*\(/, '').replace(/\);$/, ''));
```
moollaza: Perhaps we should just make an `alt_to` in the Perl so we don't need to parse the response via regex -- that seems like a much safer solution.

share/spice/currency/currency.js
moollaza: What's this for?

pjhampton: Both your comments are related. The api is returning a jsonp response which for one reason or another is throwing an error each time I run `$.ajax`, `$.getJSON` so I'm returning it as text and parsing into valid json. In the above line I'm overriding the MIME type of the response to plain text.

moollaza: Ah, I see. I think avoiding these work around with a new endpoint is a simpler solution. We can make alt to called `currency_json` and disable the jsonp wrapping.

Then you can `$.getJSON('/js/spice/currency_json/' ... `

moollaza:
Closing as it appears this won't be moving forward :(

pjhampton:
No problemo @moollaza 👌  onwards and upwards :)

Issue #3101, opened 2017-01-20
Libraries.io: Comprehensive Package Manager Search 


pjhampton:
## Description

There is still a lot of package managers that we don't any have any Spice / Fathead for. This includes:

  - Go Search
  - WordPress Plugins Search
  - Sublime Plugins Search
  - and so on...

Using this API should [hopefully] improve coverage of a few of the languages and be fairly easy to maintain in the long run.

## People to notify
   - @moollaza 
   - @pjhampton

## Get Started
- [x] 1) Claim this issue by commenting below
- [x] 2) Review our [Contributing Guide](https://github.com/duckduckgo/zeroclickinfo-spice/blob/master/CONTRIBUTING.md)
- [x] 3) [Set up your development environment](https://docs.duckduckhack.com/welcome/setup-dev-environment.html), and fork this repository
- [x] 4) Create a Pull Request

## Resources
- Join [DuckDuckHack Slack](https://quackslack.herokuapp.com/) to ask questions
- Join the [DuckDuckHack Forum](https://forum.duckduckhack.com/) to discuss project planning and Instant Answer metrics
- Read the [DuckDuckHack Documentation](https://docs.duckduckhack.com/) for technical help

------
IA Page:  http://duck.co/ia/view/libraries 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @pjhampton

karuppiah7890:
@pjhampton Yeah, I actually tried a bit of code. And then got stuck in some places and then stopped after that. And I don't think I can continue that work immediately.

karuppiah7890:
@pjhampton Give me a day. And I will update you more on this and see what I did. It's been two months and I don't remember a thing now.

pjhampton:
Hey @karuppiah7890, 

How are you getting on? 😄

karuppiah7890:
Ah, I was caught up. Yet to try renaming all files and trying. I planned to use the IA name of your IA page just for checking how the Duckpan server works and if my code runs. But I didn't do it.

By the way, my code is entirely different. It simply searches for libraries in all languages given a "description of the functionality", hoping libraries.io searches even description attribute in it's search API - most probably it does.

But your IA page mentions that it's for searching libraries from different languages, using their names. And looks like we gotta write code from scratch. Or does it search based on description too ? And show many results

pjhampton:
Hey @karuppiah7890 

[...]

PR #3330, opened 2017-06-14
Libraries IA: Rewrite stale PR

pjhampton:

To start with, this IA will introduce **<strike>18</strike> 16 new package managers** 🤡  🎉  I forked the stale PR from a community member and completely rewrote it to include single view, familiar favicons, infobox data, and fuzzy search.

-  alcatraz (Swift)
-  atom,
-  bower
-  cran (R)
-  Pub (Dart)
- dub (D-lang)
- elm
- emacs
- go
- <strike>groovy</strike> (No longer on libraries.io it seems)
- <strike>haxe</strike> (Already exists)
- inqlude (C++/Qt)
- julia
- meteor (node)
- nimble (nim)
- nuget (c#)
- shards (crystal)
- wordpress

## Images

**Tile View**
screen shot 2017-06-14 at 10 12 14 pm

**Julia detail view**
screen shot 2017-06-14 at 9 13 08 pm

**Elm detail view**
screen shot 2017-06-14 at 9 08 13 pm

**Go Lang Package view**
screen shot 2017-06-14 at 9 07 39 pm

Fixes: https://github.com/duckduckgo/zeroclickinfo-spice/issues/3101

## People to notify

@moollaza 
/ cc. @bsstoner

---

Instant Answer Page: https://duck.co/ia/view/libraries


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

pjhampton:
I had to drop support for Dart, Groovy and Haxe as the api was returning 500 for everything 😖

moollaza:
> I had to drop support for Dart, Groovy and Haxe as the api was returning 500 for everything 😖

That's unfortunate. I'm assuming it was working before? @andrew, would you happen to know if there were any recent issues with the Libraries.io API?

pjhampton:
Yeah, it was working fine last week. Not sure what happened 👍

andrew:
I'm traveling at the moment, please do open issues on the libraries API bugs

andrew:
Looks like we had one of the web servers fall over whilst I was travelling but it didn't get removed from the load balancer so that could have explained the 500 errors you were seeing,

pjhampton:
Ahh cool. Thanks for response @andrew

lib/DDG/Spice/Libraries.pm
```diff
@@ -0,0 +1,50 @@
...
+    chomp($line);
+    my @aliases = split(/,/, $line);
+    push(@trigger_aliases, @aliases);
+    $alias_hash{$aliases[0]} = \@aliases;
+}
+
+triggers start => @trigger_aliases;
+
+spice from => '(.+)/(.+)';
+spice to => 'https://libraries.io/api/search?platforms=$1&q=$2&api_key=34b827d44ed29b80497b1634c2e7e115';
```
moollaza: Let's replace this with an ENV Var. I'll put a key into the production machines 👍

moollaza: `DDG_SPICE_LIBRARIESIO_APIKEY`

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,207 @@
...
+        
+        var text_template =  {
+            group: 'text',
+            options: {
+                content: Spice.libraries.content,
+                moreAt: true
+            }
+        }
+
+        var template, meta, signal;
```
moollaza: I would initialize template and signal to `tile_template` and `low`. Then you only need to set it for special cases.

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,207 @@
...
+        var pkgname = api_result[0].name;
+
+
+        meta = {
+            search_term: search_items,
+            total: api_result.length,   
+        }
+
+        // determine which template to show and high/low signal
+        if(api_result.length <= 1) {
```
moollaza: Shouldn't this be `length === 1`? If length is 0, it would have returned above

moollaza: I would write this as `if ( total === 1 ) {`

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,207 @@
...
+            template = text_template;
+            signal = "high";
+        } else if(pkgname.toLowerCase() == query) {
```
moollaza: I think would should combine these `else-ifs`? This block looks a little redundant

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,207 @@
...
+            templates: template,
+            sort_fields: {
+                rank: function(a, b) {
+                    return a.rank > b.rank ? -1 : 1;
+                }
+            },
+            sort_default: 'rank',
+            normalize: function(item) {
+
+                console.log(item.name);
```
moollaza: whoops, should be removed

share/spice/libraries/libraries.js
moollaza:
```js
if(pkgname.toLowerCase() === query
    || ( ... )
    || ( ... )) {
              template = "text_template";
              signal = "high";
}

if (total > 1) {
 api_result = api_result.shift();
}
```

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,207 @@
...
+
+        // determine which template to show and high/low signal
+        if(api_result.length <= 1) {
+            template = text_template;
+            signal = "high";
+        } else if(pkgname.toLowerCase() == query) {
+            template = text_template;
+            api_result = api_result[0];
+            signal = "high";
+        } else if(api_result[0].platform === "Elm" && pkgname.split("/")[1] == query) {
```
moollaza: nit: mixed use of `===` and `==` -- `===` should be fine for both checks 👍

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,207 @@
...
+        } else {
+            template = tile_template;
+            signal = "low";
+            meta.sourceName = 'libraries.io';
+            meta.sourceUrl = 'https://libraries.io/api/search?q=' + search_items;
+        }
+
+        Spice.add({
+            id: 'libraries',
+            name: 'Libraries',
```
moollaza: I think `name` should be `Software` like we do elsewhere.

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,207 @@
...
+                    url = item.homepage;
+                } else if(item.package_manager_url != null && item.package_manager_url != "") {
+                    url = item.package_manager_url;
+                } else if(item.repository_url != null && item.homepage != item.repository_url) {
+                    url = item.repository_url;
+                } else {
+                    url = 'https://libraries.io/api/search?q=' + search_items;
+                }
+
+                if (item.name && item.description) {
```
moollaza: Should we bail earlier if these are not true?

i.e. at the top `if (!item.name && item.description) { return }`

Alternatively, you could set a relevancy block to require these to be defined. You'd need to set the relevancy block depending on whether we're doing a list or detail view

share/spice/libraries/libraries.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    // sets up the install script for text view
+    // eg. elm-package install elm-lang/html
+    function getInstallScript(pkg_manager, pkg_name, pkg_version) {
+        var index = pkg_manager.toLowerCase();
+        var text = install_scripts[index] || false;
+
+        if(index === "nuget" || index === "pub") {
+            text = text.replace('{{package}}', pkg_name);
```
moollaza: nit: you can chain the replaces here: `text = text.replace(...).replace(...)`

pjhampton:
I've updated this to prevent the overtriggering.

moollaza:
@pjhampton Thanks! This LGTM :+1:

moollaza:
@pjhampton Thanks! This LGTM :+1:

PR #3369, opened 2017-07-19
Expand package tracking fallback

bsstoner:
- Expands use of the fallback in cases where the upstream times out or errors.
- Lowers the signal when the fallback is added so that we don't show it unless we absolutely have to (e.g. the Answer bar already opened and left space for it).

/cc @pjhampton @moollaza


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

PR #3431, opened 2017-09-28
call duckbar.failed() for empty maps responses

bsstoner:
## People to notify

---

Instant Answer Page: https://duck.co/ia/view/{ID}


bsstoner:
👍

PR #3465, opened 2017-11-30
Amazon: Implement UI Tests

moollaza:
Implements new UI Experiment by using a new custom Spice template that allows for a Brand link.

---
Instant Answer Page: https://duck.co/ia/view/products

### DIFF
--- a/share/spice/amazon/amazon.js
+++ b/share/spice/amazon/amazon.js
@@ -17,7 +17,7 @@
 
         if (api_result.results.length > 1) {
             templates = {
-                item: 'products_item',
+                item: Spice.amazon.branded_products_item,
                 options: {
                     buy: 'products_amazon_buy',
                     badge: 'products_amazon_badge',
--- a/share/spice/amazon/amazon.js
+++ b/share/spice/amazon/amazon.js
@@ -16,8 +16,9 @@
         };
 
         if (api_result.results.length > 1) {
+            var isBrandExp = DDG.page.ads.adxExperiment === 'prod_brand_v1';
             templates = {
-                item: Spice.amazon.branded_products_item,
+                item: isBrandExp ? Spice.amazon.branded_products_item : 'products_item',
                 options: {
                     buy: 'products_amazon_buy',
                     badge: 'products_amazon_badge',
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/spice/amazon/branded_products_item.handlebars
```diff
@@ -0,0 +1,27 @@
+<div class="tile  tile--pr tile--c has-detail  tile--{{parentId}} opt--t-xxs" data-link="{{url}}">
```
moollaza: For the purpose of the experiment, I've simply copied this from the open-source template to allow for easily adding the new Brand link when needed.

If the test is a success, the Products template should be updated to allow for the Brand link

moollaza:
@bbraithwaite I've added style for the dark theme 👍 

juicer at duckduckgo 2017-12-06 20-54-12

share/spice/amazon/amazon.js
```diff
@@ -16,8 +16,9 @@
         };
 
         if (api_result.results.length > 1) {
+            var isBrandExp = DDG.page.ads.adxExperiment === 'prod_brand_v1';
```
bbraithwaite: We can use env here over DDG.

Also I'm pretty sure that env.page.ads will never cause a null reference error here, but it's worth double, double checking!

share/spice/amazon/branded_products_item.handlebars
bbraithwaite: 👍 Nice approach, much cleaner than adding IF then THIS etc to the other template.

share/spice/amazon/amazon.js
moollaza: > We can use env here over DDG.

@bbraithwaite I don't think so? AFAIK `env === window` here. So it'd need to be `env.DDG.ads`

>but it's worth double, double checking!

I'll add an extra check 👍

bbraithwaite: @moollaza ah you're right re: `DDG` vs `env`, I'm so used to the internal code creating the IIFE with (DDG). 👍

bbraithwaite:
👍

PR #3469, opened 2017-12-12
Amazon: Shop Now button experiment

moollaza:
More updates to Amazon for UI experiments

---

Instant Answer Page: https://duck.co/ia/view/products

### DIFF
--- a/share/spice/amazon/button_products_item.handlebars
+++ b/share/spice/amazon/button_products_item.handlebars
@@ -1,0 +1,31 @@
+<div class="tile  tile--pr tile--c has-detail  tile--{{parentId}} opt--t-xxs" data-link="{{url}}">
+    <div class="tile__media  tile__media--pr">
+        <img src="" data-src="{{{imageProxy img}}}" alt="{{title}}" class="tile__media__img  js-lazyload" />
+        {{#and showBadge meta.options.badge}}{{{include meta.options.badge}}}{{/and}}
+    </div>
+    <div class="tile__body  tile__body--pr">
+        {{#if url}}
+            {{{formatTitle heading el="h6" className="tile__title" classNameSec="tile__title--pr" href="url" optSub=true ellipsis=100}}}
+        {{else}}
+            {{{formatTitle heading el="h6" className="tile__title" classNameSec="tile__title--pr" optSub=true ellipsis=100}}}
+        {{/if}}
+    <div class="tile__tx  tile--pr__sub  one-line">
+        {{#and meta.options.price price}}
+            <span class="tile--pr__price  price">{{price}}</span>
+        {{/and}}
+        {{#and meta.options.price meta.options.brand price brand}}
+            <span class="sep  tile__sep"></span>
+        {{/and}}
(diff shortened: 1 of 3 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/spice/amazon/button_products_item.handlebars
```diff
@@ -0,0 +1,31 @@
...
+        {{/if}}
+    <div class="tile__tx  tile--pr__sub  one-line">
+        {{#and meta.options.price price}}
+            <span class="tile--pr__price  price">{{price}}</span>
+        {{/and}}
+        {{#and meta.options.price meta.options.brand price brand}}
+            <span class="sep  tile__sep"></span>
+        {{/and}}
+        {{#and meta.options.brand brand}}
+            <span class="tile--pr__brand">by {{brand}}</span>
```
moollaza: I've replaced the `DDG.Text.MADE_BY_BRAND` here because Spice templates aren't able to use `DDG.Text`.

I think this is OK because the `by` doesn't seem to be translated? The test also only applies to US users, so should be mostly English.

share/spice/amazon/button_products_item.handlebars
```diff
@@ -0,0 +1,31 @@
...
+        {{/and}}
+    </div>
+    {{#if meta.options.rating}}
+    <div class="tile__tx  tile__rating  one-line">
+        {{{starsAndReviews rating reviewCount url_review true}}}
+    </div>
+    {{/if}}
+    </div>
+    <button class="tile__button btn">Shop Now</button>
```
bbraithwaite: I guess the same point applies to the hard-coded English here, too? If this test is locked down to en_US then I guess it's ok?

bbraithwaite:
Code-wise this LGTM!

One thing, is it me or is the button text not quite middle-aligned?

btn-align

tommytommytommy:
The English hard coded changes are okay to me, but how do we internationalize later if we decide to globally roll this out?

moollaza:
> how do we internationalize later if we decide to globally roll this out?

@tommytommytommy If we decide to roll out globally, I'll have to create a token for the 'Shop Now' text, and use `DDG.Text` as we do for they `By`, then it will be translated like all other UI elements. Of course, this requires porting the button to the internal template.

share/spice/amazon/button_products_item.handlebars
moollaza: It looks like `By` is intended to be translated? But when I tried a few languages I wasn't able to see it change...

moollaza:
@bbraithwaite what browser/screen-size?

I'm seeing it centered on Chrome, FF, and Safari. The padding on both sides of the text is set to be the same and the `.btn` class applies `text-align: center`

bbraithwaite:
@moollaza I should have been clearer, I meant the vertical alignment.

moollaza:
@bbraithwaite I've updated the line-height which has adjusted the text to be centered now 👍

PR #3495, opened 2018-03-02
use jquery ajax function over get

bbraithwaite:
**Internal Review Only**

This replaces the $.get function with $.ajax so that we can get the data type to text.

### DIFF
--- a/share/spice/currency/currency.js
+++ b/share/spice/currency/currency.js
@@ -287,13 +287,15 @@
             }
 
             var endpoint = "/js/spice/currency/1/" + from + "/" + to;
-            $.get(endpoint, function(payload) {
-
-                // jsonp is returned from the API so we have to alter the contents
-                var response = JSON.parse(payload.trim().replace(/^[^\(]*\(/, '').replace(/\);$/, ''));
-                response = response.conversion;
-                Converter.resetConverter(response);
-
+            $.ajax({ 
+                url: endpoint,
+                dataType: 'text',
+                success: function(payload) {
+                    // jsonp is returned from the API so we have to alter the contents
+                    var response = JSON.parse(payload.trim().replace(/^[^\(]*\(/, '').replace(/\);$/, ''));
+                    response = response.conversion;
+                    Converter.resetConverter(response);
+                }
             });
         },
 
--- a/share/spice/currency/currency.js
+++ b/share/spice/currency/currency.js
@@ -289,9 +289,8 @@
             var endpoint = "/js/spice/currency/1/" + from + "/" + to;
             $.ajax({ 
                 url: endpoint,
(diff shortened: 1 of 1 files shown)


daxtheduck:
Metadata placeholder post
            
---
            
**How to find your IA page**
- New Instant Answer
              
- Existing Instant Answer

share/spice/currency/currency.js
```diff
@@ -287,13 +287,15 @@
...
-            $.get(endpoint, function(payload) {
-
-                // jsonp is returned from the API so we have to alter the contents
-                var response = JSON.parse(payload.trim().replace(/^[^\(]*\(/, '').replace(/\);$/, ''));
-                response = response.conversion;
-                Converter.resetConverter(response);
-
+            $.ajax({ 
+                url: endpoint,
+                dataType: 'text',
```
nilnilnil: Isn't this supposed to be `script`?

share/spice/currency/currency.js
```diff
@@ -287,13 +287,15 @@
...
+                success: function(payload) {
+                    // jsonp is returned from the API so we have to alter the contents
```
nilnilnil: JavaScript is returned.

share/spice/currency/currency.js
```diff
@@ -287,13 +287,15 @@
...
+                    var response = JSON.parse(payload.trim().replace(/^[^\(]*\(/, '').replace(/\);$/, ''));
```
nilnilnil: Should we maybe make a way to avoid this in `DDG::Rewrite` -- namely a way to return only JSON? I can see ignoring that though. In any case, it seems like we'd need to update some of the above.

share/spice/currency/currency.js
bbraithwaite: Technically, yes but that's the problem I'm trying to patch here.

jQuery would execute the script as is:

`$.ajax({ url: 'https://sakurity.com/jqueryxss', dataType: 'script' });`

Digging a little deeper into this, as is in production this makes a needless call to the main `ddg_spice_currency` each time you use the dropdown to change the currency once the IA is loaded. We could just make this JSON.

share/spice/currency/currency.js
bbraithwaite: Will remove -- I just copied the code as it was.

share/spice/currency/currency.js
bbraithwaite: That would be the most appropriate solution, that would mean `dataType: 'text'` could then be `dataType: 'json'` and this replace logic wouldn't be needed. Not sure how to go about makign the Perl changes, though...

jagtalon:
Nice looks good to me :)

PR #3506, opened 2018-04-06
Add affiliate label for amazon

bsstoner:
@tommytommytommy

### DIFF
--- a/share/spice/amazon/amazon.js
+++ b/share/spice/amazon/amazon.js
@@ -40,7 +40,17 @@
                 hideAttribution: true,
                 iconOnlyMobile: true
             };
-        }
+
+        } else if (DDG.page.ads.adxExperiment === 'prod_affiliate_v2') {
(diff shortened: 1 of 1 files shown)


tommytommytommy:
@bsstoner I think that I might have missed a dependency step in deploying this; I'm running into two problems (I pulled the branch, changed into the directory, called `cache-assets.pl` and then `fe && grunt`:
1) For desktop, the affiliate text does show up, but the top left seems to be missing the word "Results" in "Results for [query]"?
2) In mobile, I'm not actually seeing changes yet. I still get the old "Amazon" text in the upper left and no "affiliate" text.

screen shot 2018-04-09 at 11 14 13 am
screen shot 2018-04-09 at 11 18 08 am

bsstoner:
Made one more change here, just removing the translation calls since the text is experimental and only for en users, no point in translating it yet.

PR #3534, opened 2018-08-23
Remove chain detection from Spice

aaronharsh:
@moollaza  @tommytommytommy  @duendex 

This PR removes chain detection from Spice.  An internal PR adds the same chain detection back in elsewhere. Discussions on implementation and testing are in the internal PR.  These two PRs should be deployed at the same time.


aaronharsh:
@moollaza  -- are you the right reviewer for this PR?

PR #3550, opened 2018-10-06
Holiday: Normalize holiday data and query to handle 'day'

### LINKED ISSUE #3448: Holiday: "When is thanksgiving" should trigger -- normalize data, or modify query to match
The query "when is thanksgiving" doesn't trigger this IA because the list of Holidays for "Canada" and "USA" (canada.txt, us.txt) have the name of the holiday as "thanksgiving day", so a check for "thanksgiving" does not match.

We should either remove "day" from all holidays, and strip it from the query when present to allow it to match, or when the query doesn't contain day and no matches are found, we should append it to the given name to try and match.

I think its better to normalize the data and query, rather than modify the query and search for a match, but the latter is probably a quicker solution. 

There's lots more that can be done to improve this Instant Answer if someone is willing. Given the popularity of this query space, I think it's certainly worth the effort.

------
IA Page:  http://duck.co/ia/view/holiday 
[Maintainer](http://docs.duckduckhack.com/maintaining/guidelines.html): @sekhavati

machineloop:
The query "when is thanksgiving" doesn't trigger this IA because the list of Holidays for "Canada" and "USA" (canada.txt, us.txt) have the name of the holiday as "thanksgiving day", so a check for "thanksgiving" does not match.

This PR removes "day" from all holidays, and strips it from the query when present to allow it to match the normalized data.

Fixes #3448 

## People to notify
CC: @sekhavati @moollaza 

---

IA Page: http://duck.co/ia/view/holiday
Maintainer: @sekhavati

### DIFF
--- a/lib/DDG/Spice/Holiday.pm
+++ b/lib/DDG/Spice/Holiday.pm
@@ -68,6 +68,10 @@ handle query_lc => sub {
     $query =~ s/[^\w\s\-]//;
     $query =~ s/-/ /;
 
+    # Sanitize the query by replacing "day", eg:
+    #   "st patricks day" -> "st patricks"
+    $query =~ s/[[:space:]]day$//;
+
     # Extract holiday from query
     $query =~ s/(?<holiday>$triggers_re)//;
     $chosen_holiday = $+{holiday};
--- a/share/spice/holiday/sanitize-holidays.sh
+++ b/share/spice/holiday/sanitize-holidays.sh
@@ -30,6 +30,9 @@ for country in `find ./countries -iname '*.txt' | xargs`; do
     sed -i 's/^[[:space:]]*//' $country   
     sed -i 's/[[:space:]]*$//' $country
 
+    # remove trailing suffix of "day", eg: "thanksgiving day"
+    sed -i 's/[[:space:]]day$//' $country
+
     # sort and remove duplicate entries
     sort -u -o $country $country
 
--- a/share/spice/holiday/countries/ad.txt
+++ b/share/spice/holiday/countries/ad.txt
@@ -1,12 +1,12 @@
-all saints day
-ascension day
+all saints
+ascension
 assumption of mary
 bank holiday
-boxing day
+boxing
 carnival
-christmas day
+christmas
 christmas eve
-constitution day
+constitution
 daylight saving time ends
 daylight saving time starts
 day of st charles borromeu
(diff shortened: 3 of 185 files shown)


daxtheduck:
## [Holiday](https://duck.co/ia/view/holiday)

**Description**: Gives the next occurrence of the specified holiday. If a year is specified, it will give the event date for the given year.

**Example Query**: [when is halloween](https://beta.duckduckgo.com/?q=when%20is%20halloween), [when is christmas day in the uk](https://beta.duckduckgo.com/?q=when%20is%20christmas%20day%20in%20the%20uk), [when is memorial day 2018](https://beta.duckduckgo.com/?q=when%20is%20memorial%20day%202018)

**Tab Name**: Answer

**Source**: timeanddate.com

*These are the important fields from the IA page.  Please check these for errors or missing information and update the [IA page](https://duck.co/ia/view/holiday)*

---
*This is an automated message which will be updated as changes are made to the [IA page](https://duck.co/ia/view/holiday)*

lib/DDG/Spice/Holiday.pm
```diff
@@ -68,6 +68,10 @@ handle query_lc => sub {
     $query =~ s/[^\w\s\-]//;
     $query =~ s/-/ /;
 
+    # Sanitize the query by replacing "day", eg:
+    #   "st patricks day" -> "st patricks"
+    $query =~ s/[[:space:]]day$//;
```
moollaza: I would suggest a minor simplification:
```suggestion
    $query =~ s/\sday$//;
```

share/spice/holiday/sanitize-holidays.sh
```diff
@@ -30,6 +30,9 @@ for country in `find ./countries -iname '*.txt' | xargs`; do
     sed -i 's/^[[:space:]]*//' $country   
     sed -i 's/[[:space:]]*$//' $country
 
+    # remove trailing suffix of "day", eg: "thanksgiving day"
+    sed -i 's/[[:space:]]day$//' $country
```
moollaza: I guess we can stick with `[[:space:]]` here as well for consistency, though `\s` should suffice.

moollaza (changes requested):
@machineloop this is looking pretty good.

Can you update the tests to show that the expected behaviour now works. i.e. the query "thanksgiving" with a location of Canada or USA, should return the appropriate result.

moollaza:
@machineloop apologies for the delay on reviewing. If it's not already there, can you also update the shell script to deduplicate the files?

Thanks!

PR #3634, opened 2021-04-16
Equaldex: Fixed API URL redirect to HTTPS

danleveille:

Currently, the Equaldex instant answer does not appear to be working. I believe this is because the API redirects to HTTPS. (Also added "lgbtq" as a synonym of "lgbt")

## People to notify

---

Instant Answer Page: https://duck.co/ia/view/{ID}

### DIFF
--- a/lib/DDG/Spice/Equaldex.pm
+++ b/lib/DDG/Spice/Equaldex.pm
@@ -7,9 +7,9 @@ use Locale::Country;
 
 spice is_cached => 1;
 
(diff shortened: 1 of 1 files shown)


moollaza:
@danleveille thanks for the PR. This Instant Answer is actually already offline, but you're correct that the HTTPS redirect would break it if it were live.

Given that DDH is under maintenance mode and we're only accepting critical changess, I'm going to close this PR.

